PR #12, opened 2017-02-01
feat(typography-links): add links component

yhuard:


yhuard:
I kept a few notes in the README about I think is a misconception.

kalinchernev:
Will have a look at this one

kalinchernev:
This one is as good as it could be.
@yhuard the external link logic is not something that we can understand, it's an internal know-how, so migration as-is is sufficient for the moment.

PR #13, opened 2017-02-01
feat(breadcrumb): include pattern

kalinchernev:


framework/molecules/breadcrumbs/breadcrumbs.html
```diff
@@ -0,0 +1,21 @@
+<nav class="breadcrumb breadcrumb-processed {{variant}}" role="navigation" aria-label="breadcrumbs">
```
yhuard: `breadcrumb-processed` is not used anywhere. `{{variant}}` is useless too

framework/molecules/breadcrumbs/breadcrumbs.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  title: 'Breadcrumbs',
+  label: 'Breadcrumbs',
+  status: 'ready',
+  variants: [{
```
yhuard: Maybe add the default variant here (even if it's not useful). It would be less magical, more predictive (but it's not mandatory)

yhuard:
By the way, I already migrated the JS one year ago! 🎉 https://github.com/ec-europa/ne-theme-framework-dev/blob/master/src/framework/components/breadcrumb/scripts/breadcrumb.js

From here, I'll resume the development of the breadcrumb so we can showcase it today

yhuard:
@kalinchernev: we've decided with @degliwe to NOT implement the scripts that rely on jQuery. They will need to be rewritten.

PR #21, opened 2017-02-02
feat(comments): include pattern

kalinchernev:
Notes that I've changed patterns' status to ready for all that do not require further work on the js part.


yhuard:
@degliwe what's our plan regarding the status of the components? When can we consider them as ready?

degliwe:
They are "ready" only if we don't foresee any changes (do they comply to our standard?)

yhuard:
Then none of them are...

framework/molecules/comments/.gitignore
```diff
@@ -0,0 +1 @@
+dist
```
yhuard: Needed?

framework/molecules/breadcrumbs/breadcrumbs.config.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
       _demo: {
         scripts: 'Europa.breadcrumbs();',
       },
+
```
yhuard: Not needed 😄

.gitignore
```diff
@@ -4,7 +4,6 @@
 # Ignore some folders
 node_modules
 /static/framework
-dist
```
yhuard: Wrong `.gitignore`, I was referring to `framework/molecules/comments/.gitignore` 😉

kalinchernev: Oh, sorry, I can push it back, but aren't all patterns supposed to have dist ignored on package-level? At least at the moment they all have it, so this change will be for all, no?

yhuard: The discussion is open. I'd rather have `dist` ignored in the root `.gitignore` than having a `.gitignore` per component. When we'll publish the components to npm, then we may reconsider it.

kalinchernev: So, for now I can remove this package-specific ignore rule for all, correct?

yhuard: For now, we should only have one `.gitignore` at the root of the project, with `dist` or `/dist`, I can accept both

kalinchernev: It's done

yhuard:
Good for me, you can merge 🎉

PR #26, opened 2017-02-03
feat(files): add files component

yhuard:


yhuard:
Forgot to migrate https://github.com/ec-europa/digital-transformation-dev/blob/develop/lib/themes/europa/js/components/file.js

framework/molecules/files/files.html
```diff
@@ -0,0 +1,12 @@
+<div class="file">
+  <span class="icon icon--file file__icon"></span>
```
kalinchernev: please add the icon as dep. as well

framework/molecules/files/files--video.html
```diff
@@ -0,0 +1,12 @@
+<div style="width: 50%;"> <!-- only required for the style guide -->
+  <div class="file file-video file-video--with-background">
+    <div class="file__container">
+      <iframe src="http://ec.europa.eu/avservices/play.cfm?ref=I101631" allowfullscreen="allowfullscreen" scrolling="no"></iframe>
+    </div>
+    <div class="file__caption">
+      <div class="field">
```
kalinchernev: the field seems to be a dep as well

yhuard:
Fixed

PR #27, opened 2017-02-03
feat(expandables): add expandables component

yhuard:


framework/molecules/expandables/expandables.html
```diff
@@ -0,0 +1,32 @@
+<div class="expandable {{modifier_class}}">
```
kalinchernev: let's please remove this class, it's useless i think

yhuard: Good catch, removed.

PR #28, opened 2017-02-03
feat(filters): add filters component

yhuard:


yhuard:
Original implementation heavily relies on JS: https://github.com/ec-europa/digital-transformation-dev/blob/develop/lib/themes/europa/js/components/filters.js

yhuard:
@kalinchernev in fact, those dependencies were needed because markup was rendered by the JS 😉  but it doesn't really matter anyway since we'll ask the JS to be rewritten

kalinchernev:
I noticed from the js, but since it's not there, the elements are not there, so in similar way as the ignore file, let's keep it clean of things we don't have at the moment ;)

PR #47, opened 2017-02-06
feat(top-images): init component

emeryro:


framework/atoms/top-images/top-images.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  title: 'Top images',
+  label: 'Top images',
+  status: 'wip',
+  collated: false,
```
yhuard: `collated` is useless in this case

yhuard:
Good job!

PR #57, opened 2017-02-07
chore(bem): check bem syntax

yhuard:
BEM Linting:

Add `@define nameofyourcomponent` in your scss file to turn on the linting of your file


yhuard:
@kalinchernev if you don't want that annoying npm error message, run `npm run lint --silent` instead. It should suit your needs

PR #63, opened 2017-02-17
chore(deps): remove dependency on babel for tests

yhuard:
Since the tests don't use async/await anymore, babel isn't needed.


degliwe:
:+1:

PR #64, opened 2017-02-19
chore(webpack): implement webpack - INNO-275

yhuard:
TODO:
- [x] Re-implement postcss processors (autoprefixer, cssnano)
- [x] Provide both dev AND prod configs
- [x] Fix broken components
- [x] Find a way to avoid assets duplication
- [ ] Make better use of lerna
- [ ] Fix production build of fractal


yhuard:
Latest commit is a big one. It provides configs for dev and prod builds and apply postcss processors (cssnano, autoprefixer)

yhuard:
All the commented lines on index.scss are broken components. Often they use @extend when they shouldn't. They need to be fixed one by one

yhuard:
Closing this.

PR #71, opened 2017-02-22
refactor(header): make better use of macros

yhuard:
TODO:

- [x] Make sure styles comply with Stylelint rules
- [x] Add visual regression tests


yhuard:
a11y check fails on site-headers:

``` 
[critical] Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds: 
- body > .site-header.site-header--homepage.language-en > .container-fluid.site-header__container > .top-bar > .top-bar__wrapper > .lang-select-site.clearfix > .lang-select-site__link > .lang-select-site__code > .lang-select-site__code-text
-- Element has insufficient color contrast of 1.00 (foreground color: #ffffff, background color: #ffffff, font size: 9.0pt, font weight: bold)
```

Either we add a background on `.lang-select-site__code-text` to make sure it has a sufficient contrast ratio even when the background icon is not there, or we ignore it. 

I suggest to fix it by adding a background.

yhuard:
Regarding the remaining Stylelint errors, we'd have to rewrite most of the framework to lower the specificity of numerous selectors...

degliwe:
about a11y check fails we should not take any action, just report them to the design team.

about visual regression, on travis:
ERROR: failed to upload: /home/travis/build/ec-europa/europa-component-library/test/errorShots/
  err: open /home/travis/build/ec-europa/europa-component-library/test/errorShots/: no such file or directory
ERROR: failed to upload: /home/travis/build/ec-europa/europa-component-library/test/screenshots/diff/
  err: open /home/travis/build/ec-europa/europa-component-library/test/screenshots/diff/: no such file or directory

degliwe:
:+1:

PR #72, opened 2017-02-24
chore(gitmagic): update contributing.json - has no issue

yhuard:
This PR updates gitmagic's rules.

Subjects must now look like: `feat(component): action - JIRA issue`. If there's no JIRA issue, then it should be `feat(component): action - has no issue`.  

Furthermore, you must now include a body to your PR, which should give some useful information (what it fixes, how to test).

Finally, there's no more rule on the commit messages. We don't need any since we use "squash and merge" on the PRs

### DIFF
--- a/contributing.json
+++ b/contributing.json
@@ -1,17 +1,15 @@
 {
-  "commit": {
-    "subject_cannot_be_empty": true,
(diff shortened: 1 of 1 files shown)


degliwe:
:+1:

PR #73, opened 2017-02-27
fix(gitmagic): fix rules - no issue

yhuard:
I hope it fixes the fact the gitmagic doesn't catch the JIRA issue in the title


degliwe:
👍

PR #77, opened 2017-03-02
fix(search): fix component search (on Github mainly) - INNO-284

yhuard:
You can already test the GitHub Pages that I've updated using this PR. The search should work fine.

Ultimately, this should be included in the theme and the JS should be transpiled... Maybe we could provide our own fractal theme.


kalinchernev:
:+1:

PR #79, opened 2017-03-03
refactor(structure): revamp framework’s architecture - INNO-291

yhuard:
Only the architecture of the folders has changed.

To make sure it still works, try to run the local server. Everything should work as before.


kalinchernev:
@yhuard how do you think if we re-order the items in the components widgets a bit.

components_title

I think it will be a bit more aesthetic to follow the inspiration:
Layout
Templates
Content
Components
Utilities

Templates might be right after layout as to inspire the usage of the items after that?
I'd also change the title of the widget to "Library" or "Patterns" as we have a repetition at the moment.

Otherwise, I don't see any issues :)

yhuard:
+1 for "Library"

Regarding the order of the menu entries, I don't really know. I would put "Content" first as it conveys the identity of the framework. "Layout" would come second, then component or utilities. Templates, on the other hand, should either come very first, or last. Or in a dedicated section. 

ping @degliwe @emeryro

degliwe:
👍

PR #80, opened 2017-03-03
fix(package.json): remove yarn from the engines list - no issue

yhuard:
Fixes https://github.com/yhuard/ecl-example/issues/3


kalinchernev:
:+1:

PR #81, opened 2017-03-04
feat(twig): use twig templates instead of nunjucks’ ones - INNO-298

yhuard:
Unfortunately, fractal doesn't handle the `.html.twig` extension very well. For now, we'll just use `.twig`. To test this PR, just make sure that everything works as before.


degliwe:
👍

PR #88, opened 2017-03-21
feat(base): provide reset, fluid type, type scale, etc. - INNO-336

yhuard:
TODO:

- [x] Add type scales
- [x] Fix stylelint reported errors


yhuard:
Updated with the font sizes 💪

framework/_global.scss
```diff
@@ -2,17 +2,9 @@
 
 body {
   -moz-osx-font-smoothing: grayscale;
-  font-family: $font-family-base;
-  font-weight: 400;
-  line-height: 1.667;
+  -webkit-font-smoothing: antialiased;
```
emeryro: We shouldn't force antialiasing and grayscale for default font (on Windows in doesn't change anything and it consumes resources).

framework/_global.scss
```diff
@@ -26,29 +18,44 @@ body {
...
-  border: none;
+  border-width: 0;
 }
 
 .navbar-collapse,
 .navbar-fixed-top,
 .navbar {
-  border: none;
+  border-width: 0;
 }
```
emeryro: navbar are related to Bootstrap and we don't use it, so we don't need to write specific rules

framework/_global.scss
```diff
@@ -26,29 +18,44 @@ body {
...
+.sr-only {
+  border-width: 0;
+  clip: rect(0, 0, 0, 0);
+  height: 1px;
+  margin: -1px;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  width: 1px;
+}
```
emeryro: sr-only class should be put in utilities

framework/base/_fluid-type.scss
```diff
@@ -0,0 +1,31 @@
+// Implement fluid type
+// More information: https://websemantics.uk/tools/responsive-font-calculator/
+
+// Default 1rem(16px) @ 30rem(480px) increasing to 1.5rem(24px) @ 120rem(1920px)
+$fluid-type-font-size-min: map-get($font-size, 'm') !default;
+$fluid-type-font-size-max: map-get($font-size, 'xl') !default;
```
emeryro: To be checked on a large screen if font isn't too big

framework/_global.scss
yhuard: Time to break everything and to remove bootstrap!

framework/base/_fluid-type.scss
yhuard: Indeed

PR #90, opened 2017-03-28
chore(docker): include image for windows users - INNO-350

kalinchernev:
- Follow the short section in the README bottom of page
- Try to develop [mounting a local volume](https://dev.to/derickbailey/10-myths-about-docker-that-stop-developers-cold)

Possibly useful commands:
- `docker ps` for listing
- `docker logs {container-id}` to see logs if errors


README.md
```diff
@@ -75,3 +75,18 @@ Then, you can type:
 ```
 yarn gh-pages
 ```
+
+## Docker
+- Install docker
+- `cd` to `europa-component-library` project folder
+- Build: `docker build -v $PWD:/usr/src/app -t {username}/europa-component-library .`
```
kalinchernev: Hi @voidtek 
Are you sure this is the correct build command?
It's not really working on linux:

```
 ✘  ~/projects/apps/europa-component-library   chore/docker  docker build -v $PWD:/usr/src/app -t kc/europa-component-library .
unknown shorthand flag: 'v' in -v
See 'docker build --help'.
```

and the build help menu indeed does not show the `-v` flag to be a valid option:
```
 ✘  ~/projects/apps/europa-component-library   chore/docker  docker build -h                                                  
Flag shorthand -h has been deprecated, please use --help

Usage:	docker build [OPTIONS] PATH | URL | -

Build an image from a Dockerfile

Options:
      --build-arg list             Set build-time variables (default [])
      --cache-from stringSlice     Images to consider as cache sources
      --cgroup-parent string       Optional parent cgroup for the container
      --compress                   Compress the build context using gzip
      --cpu-period int             Limit the CPU CFS (Completely Fair Scheduler) period
      --cpu-quota int              Limit the CPU CFS (Completely Fair Scheduler) quota
  -c, --cpu-shares int             CPU shares (relative weight)
      --cpuset-cpus string         CPUs in which to allow execution (0-3, 0,1)
      --cpuset-mems string         MEMs in which to allow execution (0-3, 0,1)
      --disable-content-trust      Skip image verification (default true)
  -f, --file string                Name of the Dockerfile (Default is 'PATH/Dockerfile')
      --force-rm                   Always remove intermediate containers
      --help                       Print usage
      --isolation string           Container isolation technology
      --label list                 Set metadata for an image (default [])
  -m, --memory string              Memory limit
      --memory-swap string         Swap limit equal to memory plus swap: '-1' to enable unlimited swap
      --network string             Set the networking mode for the RUN instructions during build (default "default")
      --no-cache                   Do not use cache when building the image
      --pull                       Always attempt to pull a newer version of the image
  -q, --quiet                      Suppress the build output and print image ID on success
      --rm                         Remove intermediate containers after a successful build (default true)
      --security-opt stringSlice   Security options
      --shm-size string            Size of /dev/shm, default value is 64MB
  -t, --tag list                   Name and optionally a tag in the 'name:tag' format (default [])
      --ulimit ulimit              Ulimit options (default [])
```

My docker:
```
 ✘  ~/projects/apps/europa-component-library   chore/docker  docker -v
Docker version 17.03.1-ce, build c6d412e
```

voidtek: I resolve this issue in the last commit and I aligned the logic of this docker with the Docker logic.

kalinchernev:
Hi @voidtek as the docker solution is not moving very well, I'll close this one.
A solution for Windows was also introduced in the meantime on another project repo -> check [this update](https://github.com/ec-europa/ffaua-styleguide/commit/af7ff15b0b368d876a46794251f97a9dbadc5461).
Please verify if this will solve the problem on Windows for you, if you are still on Windows, of course.

PR #91, opened 2017-03-28
refactor(statuses): update statuses and fix bug in IE - INNO-199

yhuard:
Update statuses


yhuard:
7 green statuses for now

PR #132, opened 2017-05-22
feat(footer): dtt footer, adapted - EUTHEME-265

planctus:
I did what i could trying to adapt this component to the new architecture, meaning that too many things have changed in the foundations to be able to just simply "migrate" the component as it was. 
I mainly moved everything to the new implementation, i introduced the is-external selector that i needed for the moment, i imagine that different approaches could be used also regarding this.
The overall result is way different from the released component here: https://ec-europa.github.io/ec-europa-theme/section-footer.html both in terms of source code and in "behavioral" terms. The requirement in terms of responsive behavior should still work, though.
I played a bit with the twig template, the idea would have been actually to have a sort of sub-component for the footer menu but i wasn't able to make it work with fractal.

### DIFF
--- a/framework/base/_ext-selectors.sass
+++ b/framework/base/_ext-selectors.sass
@@ -1,0 +1,3 @@
+// Import all ECL base selectors
+
+@import './ext-selectors/is-external';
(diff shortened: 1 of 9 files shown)


yhuard:
Closed by https://github.com/ec-europa/europa-component-library/pull/138 

Thanks @planctus for your help :)

PR #138, opened 2017-05-23
feat(footer): upgrading - INNO-513

kalinchernev:
Should be ok


yhuard:
I've added functional tests (please see Safari results, strange behavior) and turned on BEM-checking. Review coming soon

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,122 @@
...
-Markup: html/footer.html
-
-Style guide footer
-*/
-
-.footer {
-  background-color: $blue-medium;
+/**
+ * Footers component
+ * @define footers ; weak
```
yhuard: FYI, I had to make it "weak" because of `.ecl-*** a` selectors. Read more: https://github.com/postcss/postcss-bem-linter#conformance-tests

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,122 @@
...
+ */
+
+.ecl-footers__site-identity,
+.ecl-footers__site-corporate,
+.ecl-footers__ec {
+  @extend %hidden-print;
```
yhuard: Ideally, I'd use mixins instead of extends

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,122 @@
...
-.footer-bottom {
-  background-color: $brand-primary;
-  padding: 0.9em 0;
+.ecl-footers__social-icon {
+  &::before {
+    margin-right: 8px !important;
+  }
+
+  &::after {
+    content: none !important;
```
yhuard: Where does it come from?

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,122 @@
...
 }
 
-.footer,
-.footer-bottom {
-  @extend %hidden-print;
-  color: $white;
+.ecl-footers__title {
+  border-bottom: 2px solid map-get($colors, 'blue-50');
+  color: #fff;
+  margin-bottom: 1.1rem;
```
yhuard: `margin` and `padding` should always use the standards values we have defined

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,122 @@
...
+    // Make titles same height on mobile breakpoint.
+    min-height: 3.2rem;
   }
-}
 
-.footer__label {
-  display: block;
-  margin-right: 20px;
+  // Make titles height auto on large screens.
+  @media (min-width: 1291px) {
```
yhuard: No magic number please :)

framework/components/ecl-footers/ecl-footers.twig
```diff
@@ -1,86 +1,109 @@
...
+      <div class="section__body">
+        <div class="ecl-row">
+          <div class="ecl-col-sm ecl-footers__column">
+            <h4 class="ecl-h4"><a href="#">Digital single market</a></h4>
+          </div>
+          <div class="ecl-col-sm ecl-footers__column">
+            <p class="ecl-footers__label">Follow us:</p>
+            <ul class="ecl-footers__menu ecl-list-inline">
+              <li class="ecl-footers__menu-item ecl-list-inline__item">
+                <a href="#" class="ecl-icon ecl-icon--facebook ecl-footers__social-icon" target="_blank">Facebook</a>
```
yhuard: Question about the icons: why don't you print them in `span` elements?

framework/components/ecl-footers/_footers.scss
kalinchernev: I decided to keep this, as it's in the current codebase of europa-theme and I can't also understand it to know whether it's save to remove.

framework/components/ecl-footers/_footers.scss
kalinchernev: Another very specific rule in the codebase of europa-theme i'm afraid to touch to avoid broken hearts and big conflicts.

framework/components/ecl-footers/ecl-footers.twig
kalinchernev: No specific reason, will try to rework

framework/components/ecl-footers/_footers.scss
kalinchernev: Do you mean using a class like `ecl-mv-l` or using a var with standard value?

yhuard: Nope, I mean using `map-get($ecl-spacing, 'm')` or similar. Thus we use the same spacing everywhere (consistency matters)

framework/components/ecl-footers/_footers.scss
yhuard: Haha, I understand 😆 Anyway, I would use `@include media-breakpoint-up('xl') {` instead.

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,144 @@
...
+ */
+
+.ecl-footers__site-identity,
+.ecl-footers__site-corporate,
+.ecl-footers__ec {
+  @include ecl-hidden-print();
 
-.footer {
-  background-color: map-get($ecl-colors, 'blue-50');
+  a[#{$ecl-external-link-pattern}]:not(.is-internal):not(#{$ecl-ec-link-patterns})::after {
```
yhuard: Seems quite complicated... Could it be simplified? If not, let's go with it

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,144 @@
...
+  }
 }
 
-.footer,
-.footer-bottom {
-  @include ecl-hidden-print();
+.ecl-footers__title {
+  border-bottom: 2px solid map-get($ecl-colors, 'blue-50');
   color: #fff;
+  margin-bottom: 0.3rem;
```
yhuard: Please use standard spacing values :)

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,144 @@
...
-      &:hover,
-      &:active,
-      &:focus {
-        background: none;
-        text-decoration: none;
-      }
-    }
+  .ecl-footers__menu-item {
+    font-size: map-get($ecl-font-size, 'xs');
+    margin-bottom: 0.4rem;
```
yhuard: Standard value please :)

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,144 @@
...
   }
+}
 
-  .footer__menu-item-right {
-    @extend .footer__menu-item;
-    float: right;
+.ecl-footers__column {
+  padding-top: 1.6rem;
```
yhuard: Standard value please :)

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,144 @@
...
+    text-decoration: none;
   }
-}
 
-.footer__icon--facebook {
-  &::before {
-    font-size: 20px;
+  .ecl-footers__column {
+    padding-bottom: 0;
+    padding-top: 0.8em;
```
yhuard: Standard value please :)

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,144 @@
...
   }
-}
 
-.footer__icon--twitter {
-  &::before {
-    font-size: 18px;
+  .ecl-footers__column:last-child {
+    padding-bottom: 0.8em;
```
yhuard: Standard value please :)

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,144 @@
...
+  a {
+    color: #fff;
   }
 }
 
-.footer {
+.ecl-footers__ec {
+  background-color: $ecl-color-primary;
+  color: #fff;
+  padding: 0.9rem 0;
```
yhuard: Standard value please :)

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,144 @@
...
-      &:not(#{$ecl-ec-link-patterns}) {
-        @include _footer_external_link_mixin();
-      }
+    color: #fff;
+  }
+}
+
+@include ecl-media-breakpoint-up(sm) {
+  .ecl-footers__column {
+    padding-bottom: 2rem;
```
yhuard: Standard value please :)

framework/components/ecl-footers/_footers.scss
```diff
@@ -1,142 +1,144 @@
...
+  }
+
+  /* Specifics for this type of footer when non-mobile */
+  .ecl-footers__site-identity {
+    .ecl-footers__column {
+      padding-bottom: 1rem;
+    }
+
+    .ecl-footers__column:last-child {
+      padding-top: 1rem;
```
yhuard: Standard value please :)

framework/base/variables/_lists.scss
```diff
@@ -0,0 +1 @@
+$list-inline-margin: 15px !default;
```
yhuard: Please prefix with `$ecl-` ;)

framework/components/ecl-footers/_footers.scss
kalinchernev: I was thinking on this for some time. In direction, do we need to support this at this stage at all, as this is indeed very-specific things carried years for handling very special cases.
Since the scenarios for external, internal, etc. are dark magic to me, in order to avoid breaking some mystical ancient know-how I left it alone, I don't really know how to even test this correctly.

framework/base/variables/_spacing.scss
```diff
@@ -2,6 +2,7 @@
 
 $ecl-spacing: (
   none: 0,
+  xxxs: 0.33rem,
```
yhuard: More or less 6px then :) After shipping the first version, we should open the discussion with the design team about normalizing those values

PR #148, opened 2017-05-29
refactor(tables): update tables - INNO-517

emeryro:
Tables updated from DTT implementation (no DT specifications).
Responsive display is managed both by javascript and css (javascript converted from jquery to vanilla).

Code should probably be optimized in the future (once we have real specifications)


framework/content/ecl-tables/ecl-tables.js
```diff
@@ -0,0 +1,82 @@
+/**
+ * Tables related behaviors.
+ */
+
+// eslint-disable-next-line import/prefer-default-export
+export function eclTables() {
```
yhuard: What if you want to apply this transformation not on all tables but only on a subset? My two cents: you could add an optional parameter ` eclTables(elements = null)` and then reuse it `const tables = elements === null ? document.getElementsByClassName('ecl-table') : elements;`. Or you could pass the the query selector and then call `querySelectorAll` instead of `getElementsByClassName` :)

emeryro: It may be an improvement.
Anyway, the code will have to be updated, as now it isn't optimal (too many DOM browsing that should be put in variables).
I'd say, we merge the branch as is, and we create a ticket for this, with your comment. If we have time to deal with it before thursday, great, in not we have at least something working.

yhuard: Solution: we create a follow-up ticket and merge this PR.

PR #151, opened 2017-05-30
feat(lang-select-sites): refactoring the component - INNO-529

kalinchernev:
Pull is actually solving INNO-529, which is a site header, not a page header, but INNO-512 shows a site header in the name of a page header.


yhuard:
Does not seem aligned vertically?

kalinchernev:
Vertical alignment changed.
The tests seem to pass now, but still the build fails of a timeout...

yhuard:
Giving it a second look :)

yhuard:
I've merged it since the only failing test was due to a timeout (too many tests running at the same time on Sauce Labs)

PR #152, opened 2017-05-30
fix(tables): align content to the left -INNO-527

degliwe:
Fix a small defect compared to the reference


yhuard:
Let me add tests for the tables

yhuard:
As the screenshot show it, `text-align: initial` keeps the text centered in Safari and IE!

PR #164, opened 2017-06-06
feat(footers): correct a typo - INNO-569

kalinchernev:
Small correction taken by mistake from the old style guide.


yhuard:
Hehe, the refs need to be updated 😉

PR #172, opened 2017-06-07
refactor(buttons): make proper use of composition - INNO-590

yhuard:
Does the syntax of `ecl-buttons.twig` seem clear enough?


yhuard:
Hmmm, indeed it makes sense in some cases to use `embed` instead of `include`, and thus passing children elements instead of a simple label (only the `input` element can't accept children). Will work on it!

PR #184, opened 2017-06-12
feat(tabs): provide first version of navigation tabs - INNO-600

yhuard:
What should be the responsive behavior? Are there other variants (e.g. pills)?

TODO:
- [x] write tests
- [x] use standard variables


framework/components/ecl-navigation/README.md
```diff
@@ -0,0 +1,5 @@
+# Navigation
+
+## Default
+
+## Tabs
```
emeryro: Should be a little more descriptive

framework/components/ecl-navigation/_navigation.scss
```diff
@@ -0,0 +1,64 @@
...
+
+.ecl-navigation__item {
+  flex-shrink: 0;
+
+  .ecl-navigation--tabs & {
+    margin-bottom: -$ecl-navigation-border-width;
+  }
+}
+
+.ecl-navigation__item-link {
```
emeryro: we have to synchronize here: I named it __link in the horizontal menu, and not __item-link.
So which one do we keep?
I would go for __link as it is descriptive enough

framework/components/ecl-navigation/_navigation.scss
```diff
@@ -0,0 +1,64 @@
...
+  display: block;
+  font-weight: bold;
+  padding: map-get($ecl-spacing, 'xs') map-get($ecl-spacing, 's');
+  text-decoration: none;
+
+  &:hover {
+    text-decoration: underline;
+  }
+
+  /* stylelint-disable-next-line */
```
emeryro: Why? :)

framework/components/ecl-navigation/_navigation.scss
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  .ecl-navigation--tabs & {
+    border: $ecl-navigation-border-width solid map-get($ecl-colors, 'grey-50');
+    border-bottom-color: #fff;
+    text-decoration: none;
+
+    /* stylelint-disable-next-line */
+    &:focus {
+      background-color: $ecl-color-secondary;
```
emeryro: Not sure we need to re-declare background-color here

yhuard: Good catch 👍

framework/components/ecl-navigation/_navigation.scss
yhuard: Hhmmm, we need to sync. The horizontal menu is a navigation component

PR #185, opened 2017-06-12
feat(expandable): refactor basic collapse/expand functionality - INNO-580

yhuard:
Expandable (or collapsible, or toggle) is just a bit of JS and markup (no CSS). 

```
button
  aria-controls="example-expandable-0"
  aria-expanded="false"
  id="example-expandable-button-0"

div
  aria-hidden="true"
  aria-labelledby="example-expandable-button-0"
  id="example-expandable-0"
```


framework/components/ecl-expandables/_expandables.scss
```diff
@@ -1,45 +1,4 @@
-/*
-Expandable
-*/
-
-.expandable__content {
```
kalinchernev: Why removing?

framework/components/ecl-expandables/ecl-expandables.config.js
```diff
@@ -1,6 +1,16 @@
 module.exports = {
   title: 'Expandables',
   label: 'Expandables',
-  status: 'planned',
+  status: 'ready',
   tags: ['molecule'],
+  context: {
+    _demo: {
+      scripts: `
+        document.getElementById('example-expandable-button-0').addEventListener(
```
kalinchernev: Please add documentation for the JS part. Even when it's as simple as that, not every one has the time to read the code to understand the mechanics.

framework/components/ecl-expandables/expandables.js
```diff
@@ -0,0 +1,15 @@
...
+  const target = document.getElementById(
+    toggleElement.getAttribute('aria-controls')
+  );
+
+  // Get current status
+  const isExpanded = toggleElement.getAttribute('aria-expanded') === 'true';
+
+  // Set
+  toggleElement.setAttribute('aria-expanded', !isExpanded);
+  target.setAttribute('aria-hidden', isExpanded);
```
kalinchernev: A basic check whether this exists please.

framework/components/ecl-expandables/expandables.js
```diff
@@ -0,0 +1,15 @@
+export const toggleExandable = toggleElement => {
```
kalinchernev: I'd include some sort of default value for the arg.

framework/components/ecl-expandables/_expandables.scss
yhuard: Let me explain :)

What the DTT called [expandable](https://ec-europa.github.io/ec-europa-theme/section-expandable.html#kssref-expandable-1) is, in fact, nothing more than an accordion, right? The problem is that they also have [expandable links](https://ec-europa.github.io/ec-europa-theme/section-expandable.html#kssref-expandable-2) or [button collapse](https://ec-europa.github.io/ec-europa-theme/section-buttons.html#kssref-buttons-btn-collapse)... This is confusing. 

What I'm willing to achieve here is to simplify the definition of "expandable" ("expandable" and "collapsible" are the same thing, only the initial state is different). Put it simply, an expandable component is a container wrapping content which expands or collapses when you click on a lever (a button for example). That's all. "Button collapse", "expandable links", I don't see any difference. They are just special cases of the "expandable behavior". I will add variants for them, so there will be some CSS rules to be added, but the "ecl-expandable" CSS component won't exist because it is an abstract (or virtual) component (see it as an "interface" in OOP 😄)

framework/components/ecl-expandables/ecl-expandables.config.js
yhuard: Sure, I'll add some doc

framework/components/ecl-expandables/expandables.js
yhuard: Of course :)

framework/components/ecl-expandables/expandables.js
yhuard: Regarding this function: either the element to toggle exists, or it doesn't. I'll add a check but no default.

Besides that, we could maybe add another function which, on "DOMContentLoaded" event, will parse the DOM and automatically attach the click event listener to all the expandables links. It will only be a helper and will still have to be called by the implementer.

yhuard:
Thank you @kalinchernev for the precious review. 

In a nutshell, I'll add variants for "button collapse" and "expandable links", while the accordion component will be implemented later

framework/components/ecl-expandables/expandables.js
kalinchernev: :+1:

PR #190, opened 2017-06-14
feat(horizontal-menu): add component - INNO-595

emeryro:
Brand new component.
Specifications are in Jira ticket.

~~Note: currently, opening a group do not close the other ones. It will be added once the "accordion" component is available on ECL (in a follow-up ticket)~~

Won't do: 
- [ ] implement keyboard navigation (arrow keys)


kalinchernev:
In overall works well.
This is something a bit weird on emulated IE, could be different on real one:
https://drive.google.com/file/d/0B8knVVW86wNgVGZndFZKODMyVjg/view?usp=sharing

PR #197, opened 2017-06-15
refactor(messages): upgrade component - INNO-587

emeryro:
Message display + javascript behavior to dismiss them


framework/components/ecl-messages/_messages.scss
```diff
@@ -6,115 +6,91 @@ Markup: html/messages.html
 Style guide messages
```
yhuard: Missing `@define message` declaration

framework/components/ecl-messages/_messages.scss
```diff
@@ -6,115 +6,91 @@ Markup: html/messages.html
...
+  text-decoration: none;
+  text-shadow: none;
+  top: map-get($ecl-spacing, 'xxs');
+
+  &:hover {
+    background-image: url('../images/svg/close_hover.svg');
   }
 }
 
 .no-svg {
```
yhuard: Right now, we don't have `no svg` feature detection. Should we implement it as well as the `no js` detection?

framework/components/ecl-messages/ecl-messages.twig
```diff
@@ -1,5 +1,15 @@
-<div class="messages alert {{messages_type}}">
-  <a href="#" class="close" data-dismiss="alert">&times;</a>
-  <h3><span class="ecl-sr-only">{{messages_sr_only}}:</span>{{messages_title}}</h3>
-  <p>{{messages_data}}</p>
+<div class="ecl-message {% if modifier is defined %} ecl-message--{{ modifier }} {% endif %}">
```
yhuard: I'm wondering if it could be considered as a `dialog`... https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role

framework/components/ecl-messages/_messages.scss
kalinchernev: I'd say no, as @nelsongrin mentioned they plan to drop it, at least that was few months ago.

framework/components/ecl-messages/ecl-messages.twig
kalinchernev: Not to my knowledge. The only dialog I know of is the language switcher on inner pages. Galleries maybe also, but we have to double-check them when time comes.
These are the messages shown in Drupal, as far as I know.

yhuard: It should at least have an `alert` role: https://www.w3.org/TR/wai-aria/roles#alert

yhuard: And the close button should not have `data-dismiss="alert"` (that's bootstrap-specific), but it should have an `aria-label` :)

emeryro:
PR updated, ready for review

framework/components/ecl-messages/ecl-messages.twig
```diff
@@ -1,5 +1,15 @@
-<div class="messages alert {{messages_type}}">
-  <a href="#" class="close" data-dismiss="alert">&times;</a>
-  <h3><span class="ecl-sr-only">{{messages_sr_only}}:</span>{{messages_title}}</h3>
-  <p>{{messages_data}}</p>
+<div class="ecl-message {% if modifier is defined %} ecl-message--{{ modifier }} {% endif %}" role="alert">
+  <span class="ecl-sr-only">{{ srOnly }}</span>
+
+  {% if dismiss is defined and dismiss %}
+  <button type="button" class="ecl-message__dismiss" aria-label="Dismiss this alert">&times;</button>
```
yhuard: `&times;` should be in a `<span aria-hidden="true">`

framework/components/ecl-messages/_messages.scss
```diff
@@ -1,120 +1,63 @@
...
-      background-image: url('../images/png/close.png');
+.ecl-message__body {
+  margin: 0;
+}
 
-      &:hover {
-        background-image: url('../images/png/close_hover.png');
-      }
-    }
+.ecl-message__dismiss {
```
yhuard: Here I'd explicitly declare `width` and `height` and a background-size: contain`. No need for `line-height` or `font-size` then

emeryro:
Overlap has been fixed.
Modifing the dismiss icon causes display problems on IE (icon is truncated). I suggest not to waste more time on this and keep it as is for now.

yhuard:
Good enough for now. I've created a follow-up ticket with my latest comments

PR #200, opened 2017-06-19
feat(datepickers): add new way - INNO-592

kalinchernev:
The previous and next buttons are not ready, but the rest is hopefully ok.
I open this for first feedback as the component has been in progress for some time already.

Reference to [the lib](https://github.com/dbushell/Pikaday). A preview template is used to instantiate the demo since the integration is not really needed in the framework IMO.


framework/_partials/_preview-datepickers.twig
```diff
@@ -0,0 +1,34 @@
...
+  </head>
+  <body{% if _target.context and _target.context.global and _target.context.global.language %} class="language-{{ _target.context.global.language }}"{% endif %}>
+    {{ yield }}
+    <script src="{{ '/framework/scripts/europa.js' | path }}"></script>
+    {% if _target.context._demo and _target.context._demo %}
+    <script type="text/javascript">{{ _target.context._demo.scripts | raw }}</script>
+    {% endif %}
+
+    <!-- Specific scripts included in this preview template -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/pikaday/1.6.1/pikaday.min.js"></script>
```
yhuard: A bit messy but it works. Maybe by using proper template inheritance we could avoid rewriting the same code over and over again

framework/index.js
```diff
@@ -1,5 +1,6 @@
 // 1. Import polyfills (if needed)
```
yhuard: What happened here? 😄

kalinchernev: Mistake in merging ....

yhuard:
Please don't forget to add tests ;)

yhuard:
IE doesn't display the icons, probably because the parent has this `visibility: hidden`

kalinchernev:
I removed the visibility rules for the IE reason indeed.
@yhuard thank you for giving me the idea to use a set date, now the tests should be consistent.

yhuard:
Looks nice! :)

I've brought a few modifications to the CSS:
- added the stylelint `@define` declarations
- scoped pikaday's specific rules into ".ecl-pika-theme" namespace

And then I've changed the template in order to call the plugin with the right theme. This way, we ensure that our changes don't leak

PR #201, opened 2017-06-20
feat(tabs): add tabs component - INNO-617

yhuard:
TODO: 
- [x] add tests


framework/components/ecl-tabs/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@ec-europa/ecl-navigation-lists",
```
kalinchernev: `tabs`, small thing

framework/components/ecl-tabs/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "0.1.0",
+  "description": "ECL Tabs",
+  "main": "_ecl-tabs.scss",
+  "style": "_ecl-tabs.scss",
+  "peerDependencies": {
+    "@ec-europa/ecl-links": "^0.1.0",
```
kalinchernev: The version of is `0.5.0` at the moment, could be a bit higher

framework/components/ecl-tabs/tabs.js
```diff
@@ -0,0 +1,136 @@
+// Heavily inspired by the tab component from https://github.com/frend/frend.co
+
+// Query helper
+const q = (el, ctx = document) => [].slice.call(ctx.querySelectorAll(el));
```
kalinchernev: Maybe add it a bit more globally so all components can benefit?

yhuard: I've also thought about it. Making a small set of utilities provided with `ecl-base`. Will do it :)

framework/components/ecl-tabs/package.json
yhuard: How did I miss that? 😄 Thanks!

framework/components/ecl-tabs/package.json
yhuard: At some point, we will have to rethink the way we handle the dependencies between the packages. Right now, we only use `peerDependencies` which are more informational / not strict enough. I guess that we will need to switch to `dependencies`, but it will also slow the bootstrap phase since lerna will have to link all the dependencies between them correctly...

framework/components/ecl-tabs/tabs.js
yhuard: Done in `ecl-base/helpers/dom.js`

framework/components/ecl-tabs/ecl-tabs.twig
```diff
@@ -0,0 +1,20 @@
+<div class="ecl-tabs">
```
kalinchernev: A question here. When the D8 (potentially) uses the templates from our packages, is it that they can use it better if we provide the information in the templates via context?

framework/base/helpers/dom.js
```diff
@@ -0,0 +1,5 @@
+// Query helper
+export const queryAll = (selector, context = document) =>
```
kalinchernev: Thanks for this!

framework/components/ecl-tabs/ecl-tabs.twig
yhuard: Absolutely. The template has to be rewritten to introduce some logic into it. So does a couple of other components. We can discuss that tomorrow :)

PR #202, opened 2017-06-20
fix(reset): fix box-sizing and remove other rules - INNO-416

yhuard:
Some rules will be moved to the next "website" preset we will provide


yhuard:
Currently, the following components break:
- breadcrumbs
- footers --default and --custom
- navigation-menus

yhuard:
Breadcrumb is now better in fact. Checking the other components...

yhuard:
All look fine, I'll update the screenshots

PR #206, opened 2017-06-21
feat(legends): legends INNO-630

degliwe:
* upgrade component
* test
* add documentation


framework/components/ecl-forms/ecl-forms-legends/README.md
```diff
@@ -1 +1,4 @@
-# Legends
```
emeryro: In other components we kept the global title, so I assume we should keep it here too (and set the other title as second level title)

framework/components/ecl-forms/ecl-forms-legends/_legends.scss
```diff
@@ -1,13 +1,10 @@
...
-Legend
+/**
+* Legend component
+* @define form-legend
 */
 
 $form-item-legend-font-size: map-get($ecl-font-size, 'xs');
 
-.legend {
+.ecl-form-legend {
```
emeryro: not sure that we have to name class "ecl-form-legend", and not only "ecl-legend". For instance we name other component's classes "ecl-text-input" or "ecl-select"

framework/components/ecl-forms/ecl-forms-legends/ecl-forms-legends.twig
```diff
@@ -1 +1 @@
-<legend>Legend text comes here</legend>
\ No newline at end of file
+<legend class="ecl-form-legend">Legend text comes here</legend>
```
emeryro: We should put the content into a variable, so we could use composition.

framework/components/ecl-forms/ecl-forms-legends/_legends.scss
degliwe: Ex: we have 2 labels components (ecl-label and ecl-form-label) this legend is specific for form element so it seems more future proof and descriptive to stick to ecl-form-legend.

framework/components/ecl-forms/ecl-forms-legends/README.md
degliwe: Like https://ec-europa.github.io/europa-component-library/components/detail/ecl-buttons ?

emeryro: Well let's say like every other component, except button :)
We will have to synchronize here

framework/components/ecl-forms/ecl-forms-legends/_legends.scss
emeryro: To be discussed with the team, as we will have this issue again in the future

PR #207, opened 2017-06-21
feat(accordions): provide accordions component - INNO-615

yhuard:
Test that you can navigate with the keyboard (arrows up/down, space)


framework/components/ecl-accordions/_ecl-accordions.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.ecl-accordions__header:active,
+.ecl-accordions__header:focus,
+.ecl-accordions__header:hover,
+.ecl-accordions__header[aria-selected='true'] {
+  background-color: $ecl-color-secondary;
+}
+
+.ecl-accordions__panel {
+  background-color: #f5f5f5;
```
emeryro: should use ecl-colors palette (grey-5)

framework/components/ecl-accordions/_ecl-accordions.scss
```diff
@@ -0,0 +1,31 @@
+/**
+ * Accordions
+ * @define accordions
+ */
+
+.ecl-accordions__header {
```
emeryro: should use "accordion" instead of "accordions" for classes, as this is what we do for other components.

framework/components/ecl-accordions/accordions.js
```diff
@@ -0,0 +1,169 @@
...
+
+  // REVEAL API
+  return {
+    init,
+    destroy,
+  };
+};
+
+// module exports
+export default accordions;
```
emeryro: so in the end, accordions do not rely on expandables at all. That's quite a shame, but that's probably easier this way.

framework/components/ecl-accordions/ecl-accordions.twig
```diff
@@ -0,0 +1,18 @@
+<div class="ecl-accordions" role="tablist" aria-multiselectable="true">
+  <h2 id="accordion-header-1" class="ecl-accordions__header" role="tab" aria-controls="accordion-panel-1" tabindex="0" aria-selected="true" aria-expanded="true">
```
emeryro: I don't like using headings inside components, as we don't know if they will match page hierarchy correctly.
Moreover, as you don't use specific ecl classes for headings, it make no real sense.

framework/components/ecl-accordions/ecl-accordions.twig
```diff
@@ -0,0 +1,18 @@
...
+    <span class="ecl-accordions__header-icon ecl-icon ecl-icon--rounded ecl-u-bg-secondary ecl-icon--growth"></span> Jobs, Growth, Investment and Competitiveness with an additional quite long string
+  </h2>
+  <div id="accordion-panel-1" class="ecl-accordions__panel" role="tabpanel" aria-labelledby="accordion-header-1" tabindex="0" aria-hidden="false">
```
emeryro: What is the reason to set a tabindex here? It somehow mess with keyboard navigation (when using tabs)

framework/components/ecl-accordions/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@ec-europa/ecl-accordions",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "0.1.0",
+  "description": "ECL Accordions",
+  "main": "_ecl-accordions.scss",
```
emeryro: in other components we didn't put "ecl-" as prefix for sass file. I don't really know the reason, but we should keep it consistant.

emeryro:
Another small remark: on small screens icons and margin take a lots of space. Maybe we could reduce the margin here.

framework/components/ecl-accordions/_ecl-accordions.scss
yhuard: Could be. I'll rewrite it ;)

framework/components/ecl-accordions/_ecl-accordions.scss
yhuard: My bad, thanks for noticing

framework/components/ecl-accordions/accordions.js
yhuard: I'm thinking about making the "expandable" JS part of `ecl-base`. Kind of a helper which we would reuse in accordion, tabs, megamenu...

framework/components/ecl-accordions/package.json
yhuard: True

framework/components/ecl-accordions/ecl-accordions.twig
yhuard: Good remark, I'll remove it. http://www.oaa-accessibility.org/examplep/accordian1/ doesn't declare it either

framework/components/ecl-accordions/accordions.js
yhuard: Created INNO-640 for following-up

PR #212, opened 2017-06-23
fix(footers): use context data - INNO-644

kalinchernev:
Short summary:
- renames scss files as agreed recently
- adds zero margin for the same reason
- adds context data
- logic does not need variant any more, because of the type property
- updated tests

Notes:
- a11y tests fail with errors I can understand:
`expected { Object (stackTrace, message) } to be null`
- the logic for showing menu titles might seem complex, but I think it's not if one follows the custom menu which contains almost all scenarios. Might need more documentation to say that presence or not of certain properties yield various types of menus and their titles


yhuard:
@kalinchernev Your a11y tests failed because the selector you were targeting did not exist, that's it

PR #238, opened 2017-07-04
feat(profile-topbar): upgrade component - INNO-586

kalinchernev:
For `@ec-europa/ecl-context-navs` there is change only in the template to be reusable, but I haven't changed the CSS as it's not in scope of this ticket.


framework/components/ecl-profile-topbars/ecl-profile-topbars.scss
```diff
@@ -0,0 +1,31 @@
...
+}
+
+.ecl-profile-topbar__pane {
+  height: 100%;
+  min-height: 160px;
+  position: relative;
+
+  > .ecl-profile-topbar__pane-button {
+    bottom: 0;
+    margin-bottom: 0;
```
emeryro: This margin is useless as buttons already have margin:0 :)

framework/components/ecl-profile-topbars/ecl-profile-topbars.scss
```diff
@@ -0,0 +1,31 @@
...
+    position: absolute;
```
emeryro: Could have use a display flex instead of position absolute.
That works well here, so we can keep it, but if the component grow larger (with several context-navs), there will be overlaps.

framework/components/ecl-profile-topbars/package.json
```diff
@@ -1,15 +1,25 @@
 {
-  "private": true,
   "name": "@ec-europa/ecl-profile-topbars",
-  "version": "0.3.0",
+  "version": "0.4.0",
```
emeryro: No need to change version manually, it is handles automagically.

framework/index.js
```diff
@@ -20,4 +20,3 @@ export * from './components/ecl-navigation/ecl-navigation-menus/megamenu';
 export * from './components/ecl-tables/ecl-tables';
 export * from './components/ecl-tabs/tabs';
 // export * from './components/ecl-timelines/timelines';
-// export * from './components/ecl-profile-topbars/profile-topbars';
```
emeryro: I assume that the js was there to manage expandable block.

test/functional/profile-toolbars.js
```diff
@@ -0,0 +1,20 @@
+describe('profile-toolbars', () => {
+  before(() => {
+    // Set viewport size
+    browser.setViewportSize({
+      width: 1400,
+      height: 600,
```
emeryro: Screenshots could be smaller (but no need to change it now).

framework/components/ecl-profile-topbars/ecl-profile-topbars.twig
```diff
@@ -1,31 +1,23 @@
...
-
-    <div id="collapse" class="profile_topbar__expander">
-        <div class="container">
-            Expander content
-        </div>
+<div class="ecl-profile-topbar">
+  <div class="ecl-container">
+    <div class="ecl-row ecl-profile-topbar__wrapper">
+      <div class="ecl-col-md-2">
+        <img src="{{ profile.avatar }}"/>
```
emeryro: Missing alt for image

test/functional/profile-toolbars.js
```diff
@@ -0,0 +1,20 @@
...
+    browser.url('ecl-profile-topbars.html');
+  });
+
+  // Normal state
+  it('should match the reference screenshot', () => {
+    const screenshots = browser.checkDocument({
+      name: 'profile-toolbars',
+    });
+    expect(screenshots).to.matchReference();
+  });
```
emeryro: Please add accessibility tests too, as in other test files:

    it('should be accessible', () => {
      const a11yReport = browser.runAxeCore('ecl-profile-topbar').value;
      expect(a11yReport).to.be.accessible;
    });

emeryro (changes requested):
Looks good, but a few changes to be done, see below

framework/index.js
kalinchernev: Actually it contained a console statement to refactor the js. Since the expandable component can be simply instantiated via the local package js file, I did that directly there without this file.

framework/components/ecl-profile-topbars/package.json
kalinchernev: :+1:

framework/components/ecl-profile-topbars/ecl-profile-topbars.twig
kalinchernev: thanks, changed to `{% include '@ec-europa/ecl-images' with profile.image %}` :)

framework/components/ecl-profile-topbars/ecl-profile-topbars.twig
```diff
@@ -1,31 +1,23 @@
...
+      </div>
+      <div class="ecl-col-md-10 ecl-profile-topbar__pane">
+        {% include '@ec-europa/ecl-context-navs' with context_nav %}
+        {% include '@ec-europa/ecl-buttons' with profile.expandable.button %}
```
kalinchernev: Created INNO-682 as a follow-up.

framework/components/ecl-profile-topbars/ecl-profile-topbars.scss
```diff
@@ -0,0 +1,31 @@
...
+ * ECL Profile Topbar
+ * @define profile-topbar
+ */
+
+.ecl-profile-topbar {
+  margin: 0;
+}
+
+// This is esentially .section__group which can be used when refactored.
+.ecl-profile-topbar__wrapper {
```
kalinchernev: You're right, it's a perfect candidate for a re-usable class. It's `section__group` as of current state of the repo. So, i added them as a wrapper and added specifically the `section__group` name so that it's found in the code base and replaced when the latter is upgraded.

framework/components/ecl-profile-topbars/ecl-profile-topbars.scss
kalinchernev: Good point :+1:

framework/components/ecl-profile-topbars/ecl-profile-topbars.scss
kalinchernev: I think the approach of relative -> absolute positioning is simple enough for this case. Flex would be a bit overkill in my opinion.

kalinchernev:
Updates applied, ready for another review round.

framework/components/ecl-profile-topbars/ecl-profile-topbars.scss
emeryro: I'm not sure I understand what you say..
My remark was: instead of using external margins on a wrapper, we should probably use padding on the main .ecl-profile-topbar element.
That way, there won't be anything going outside the component.
And also if we do so, the wrapper itself becomes useless (no more css attributes on it)

emeryro (commented):
Almost good for me.
Just see my comment concerning the wrapper.

framework/components/ecl-profile-topbars/ecl-profile-topbars.scss
kalinchernev: I think I made it right this time, the result is not totally the same, but hopefully acceptable.

emeryro (approved):

degliwe (approved):

PR #242, opened 2017-07-06
feat(messages): replace <p> by <li> --INNO-678

degliwe:
see the title ;)


framework/components/ecl-messages/ecl-messages.twig
```diff
@@ -9,7 +9,11 @@
   <div class="ecl-message__title">{{ title }}</div>
   {% endif %}
 
-  {% if body is defined %}
-  <p class="ecl-message__body">{{ body }}</p>
+  {% if messages is defined %}
```
nelsongrin: You are opening accessibility issue here!
If we have only one message, you are creating a list for just one element.
The list should be created for more than one message.

PR #243, opened 2017-07-06
feat(tests): add tests for styles elements - INNO-649

emeryro:
Add visual regression tests.
Also put test files in folders, for easier maintenance.


test/functional/styles/colors.js
```diff
@@ -19,7 +19,7 @@ describe('colors', () => {
   context('with plain state', () => {
     it('should match the reference screenshot', () => {
       const screenshots = browser.checkDocument({
-        name: 'colors',
```
kalinchernev: Can we name the folder `content` to match the structure of the framework?

test/functional/styles/icons.js
```diff
@@ -0,0 +1,90 @@
+const variants = [
```
kalinchernev: let's `require` the `ecl-icons.config.js` directly and filter this so to avoid maintaining 2 definition sets?

test/functional/styles/icons.js
```diff
@@ -0,0 +1,90 @@
...
+  'presentation',
+  'info',
+  'share',
+  'faq',
+  'camera',
+  'user',
+  'logout',
+];
+
+variants.forEach(variant => {
```
kalinchernev: I think it'd be better to test `components/preview/ecl-icons` which has them all. That covers basically all modifiers and variants at the same time, 1 image per browser.

test/functional/styles/typography/headings.js
```diff
@@ -0,0 +1,33 @@
+const variants = ['h1', 'h2', 'h3', 'h4', 'h5', 'headline'];
```
kalinchernev: Could be useful to have it from `ecl-typography-headings.config.js` as well.

test/functional/styles/typography/lists.js
```diff
@@ -0,0 +1,33 @@
...
+variants.forEach(variant => {
+  describe(`typography-lists--${variant}`, () => {
+    before(() => {
+      // Set viewport size
+      browser.setViewportSize({
+        width: 800,
+        height: 600,
+      });
+      // Go to url
+      browser.url(`ecl-typography-lists--${variant}.html`);
```
kalinchernev: Why not have them in 1 page and screenshot as in the preview?

test/functional/styles/typography/paragraphs.js
```diff
@@ -0,0 +1,33 @@
+const variants = ['l', 'm', 's', 'xs', 'xxs'];
```
kalinchernev: Another potential candidate for a single test for all variants.

kalinchernev (changes requested):
CI gives error:
```
ERROR: Cannot find module '../../framework/components/ecl-buttons/variants.json'
```

Some tests can be combined and the variants can be fetched from the components definitions.

Thanks for adding tests! :)

test/functional/styles/icons.js
emeryro: Good point

emeryro:
Tests have been updated

yhuard (changes requested):
Now that we have adopted the new folder structure for the tests, this PR has to be reworked ;)

yhuard:
Replaced by PR #289

PR #246, opened 2017-07-06
feat(messages): make bullet visible --INNO-680

degliwe:
see title


kalinchernev:
Just an idea from previous comment:

```
  {% if messages is defined %}
  <ul class="ecl-message__body ecl-list">
    {% for message in messages %}
      <li>{{ message }}</li>
    {% endfor %}
  </ul>
  {% endif %}
```

could use a check of whether is [iterable](https://twig.sensiolabs.org/doc/2.x/tests/iterable.html) with else fallback to p if the list has 1 item ...

PR #247, opened 2017-07-06
feat(gallery): upgrade component - INNO-576

kalinchernev:
It's containing the basic gallery.
Overlay is in INNO-679


framework/components/ecl-galleries/ecl-galleries.scss
```diff
@@ -0,0 +1,97 @@
...
+  // Whereas the linter wants the rules to go a new line directly.
+  /* stylelint-disable-next-line */
+  padding: map-get($ecl-spacing, 'xxxs') map-get($ecl-spacing, 'xxs')
+    map-get($ecl-spacing, 'xxxs') map-get($ecl-spacing, 'xxxs');
+  position: absolute;
+  text-overflow: ellipsis;
+  width: 100%;
+}
+
+.ecl-gallery__icon {
```
yhuard: Could be `ecl-gallery__icon-wrapper` or `ecl-gallery__icon-container`

framework/components/ecl-galleries/ecl-galleries.scss
```diff
@@ -0,0 +1,97 @@
...
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  right: 0;
+  text-align: center;
+  width: 2.667em;
+}
+
+// The following specifics are valid only for the icons in gallery context.
+.ecl-gallery__icon .ecl-icon {
```
yhuard: If you wrap this into a `ecl-gallery__icon-wrapper`, you could then simply use `ecl-gallery__icon` to target the icon and lower down the specificity

yhuard (changes requested):
I have mixed feelings about this "component". It's  mixing atoms (images, captions, icons), molecules (image + caption w/ or w/o icons) but it mainly is a layout component. I would decouple it a bit: images is already there, but we could have the caption atom and the related molecules

kalinchernev:
I made the changes to the icon class (wrapper and icon, etc.).
For the general comment, I agree it's more than a molecule which was initially the tag, that's why I changed it to organism. I don't think it's a layout element, rather it's using layout classes to make the different sizes. The images are indeed included as a composition, though another ticket was also created to expand the image component to accept custom classes and attributes as other components - INNO-677. Maybe the image can be extended also to have the caption part as well, however so far there are no other use cases of such a thing elsewhere than this gallery component.

yhuard (approved):

yhuard:
Let's put it aside. We'll discuss it again if there comes a new component which reuses the captions or the logo. ;)

yhuard:
I don't know why the tests don't run, so I'm merging this one

PR #252, opened 2017-07-10
feat(expandables): update doc - INNO-657

emeryro:
Update doc


yhuard (commented):
Why did you remove all the whitespaces?

kalinchernev (approved):
The wiki does not have spacing as here.

emeryro:
Well, no particular reason, I just kept it coherant with what had been done on another readme file.
If that really bother you, we can put they back, but visually there is no difference.

PR #260, opened 2017-07-10
feat(lists): update doc - INNO-659

emeryro:
Update doc


kalinchernev (approved):
The text is not 1:1 as in the wiki page, but let's assume they will be ok with the fact the demo is showing the same information already.

PR #268, opened 2017-07-11
feat(tests): test only updated packages - INNO-691

yhuard:
This is a WIP POC of how we could improve tests.

Here I've modified the breadcrumb SCSS in order to trigger visual regression tests on breadcrumbs and other packages that depend on breadcrumbs.


yhuard:
Next steps:

- [ ] move all the test to their component folder
- [ ] update all the package.json to actually reflect the dependency tree

PR #323, opened 2017-07-25
feat(timelines): upgrade timelines - INNO-702

yhuard:
# PR description

Timeline component. Upgraded.

## QA Checklist

- [x] I have provided tests
- [x] I have filled the README.md file (at least a few lines)
- [x] My component is listed in the root README


framework/components/ecl-timelines/_timelines.scss
```diff
@@ -1,89 +1,59 @@
...
-    }
-
-    ul > li {
-      color: map-get($ecl-colors, 'grey-100') !important;
-    }
 
-    & > * {
-      margin-bottom: 0 !important;
-    }
+  &::before {
```
emeryro: Bullets are a little too high; there are not aligned with the title (due to its line-height)

framework/components/ecl-timelines/ecl-timelines.config.js
```diff
@@ -1,13 +1,72 @@
...
   preview: '@preview-center-transparent',
   tags: ['molecule'],
   context: {
+    limit: 3,
+    button_label: 'Show all timeline',
+    items: [
+      {
+        title: 'By late-2013',
+        body: `
+          <p class="ecl-paragraph">
```
emeryro: Shouldn't we use a block here instead?
As the content may be complex (with links and so), it would be cleaner.

framework/components/ecl-timelines/README.md
```diff
@@ -1,3 +1,14 @@
...
+# Timeline and Time schedule
 
-The JS behaviour of this component needs to be rewritten without jQuery.
+## Why and how to use this component
+
+A timeline is a way of displaying a list of events in chronological order.
+
+There are two variants of this component:
+- **timeline**
+- **time schedule**
```
emeryro: What about this variant? Isn't it part of the scope?

emeryro (changes requested):
Overall display is good, a few changes to be done

framework/components/ecl-timelines/ecl-timelines.config.js
yhuard: Sure, but how do you use a block in a `for` loop?

framework/components/ecl-timelines/README.md
yhuard: I guess that either you use dates (like `By late-2013`) or times (`2pm`)  and that's they would call "time schedule"

framework/components/ecl-timelines/timelines.js
```diff
@@ -2,7 +2,54 @@
...
+
+    if (button) {
+      button.addEventListener('click', () =>
+        expandTimeline(node, button, { classToRemove, hiddenElementsSelector })
+      );
+    }
+  });
+};
+
+export default timelines;
```
yhuard: Will check it, thanks for reporting

yhuard: Note: this has been fixed on slack (internally) :)

framework/components/ecl-timelines/ecl-timelines.config.js
emeryro: Hmm.. good question...

yhuard: Do we know a Twig expert?

kalinchernev: Just pasting an idea from @emeryro (from the chat) https://twig.symfony.com/doc/2.x/tags/extends.html#block-nesting-and-scope :)

yhuard:
It should be better now ;)

framework/components/ecl-timelines/README.md
kalinchernev: Indeed, it's about the content, not the design, so there's no need to add anything for it.

kalinchernev (approved):
Styles and js is good for me.
I think the contents of the item body is also ok for the moment.
Myself, I didn't find a better way with the block inheritance. I feel that for a sophisticated inheritance, it'd be better to consider block one for the whole item like in the example https://twig.symfony.com/doc/2.x/tags/extends.html#how-do-blocks-work (not sure it's worth digging this for the moment)

emeryro (approved):

PR #330, opened 2017-07-27
refactor(gallery): add the dialog/overlay display - INNO-723

kalinchernev:
# PR description
- Contains work on gallery (a split), dialog/overlay and carousel.

## QA Checklist

- [x] I have filled the README.md file (at least a few lines) - it's same as the base as they are tightly connected
- [x] I have provided tests
- [x] My component is listed in the root README - shall we have 1 package with 2 sub-packages, or 2 separate packages?


framework/components/ecl-carousels/ecl-carousels.twig
```diff
@@ -0,0 +1,91 @@
...
+    A list of `@ec-europa/ecl-images` components.
+
+#}
+
+
+{# Internal properties #}
+
+{% set _css_classes = 'ecl-carousel' %}
+{% set _extra_attributes = '' %}
+{% set _heading_attributes = '' %}
```
emeryro: We should provide default values for headings id and classes

framework/components/ecl-carousels/test/spec/carousel.js
```diff
@@ -0,0 +1,58 @@
+const componentUrl = 'ecl-carousels.html';
```
emeryro: There seems to be display issues with the screenshots (most of them don't display the pictures).

framework/components/ecl-galleries/ecl-gallery-dialog/.npmignore
```diff
@@ -0,0 +1 @@
+test
```
emeryro: test? :)

framework/components/ecl-galleries/ecl-gallery-dialog/ecl-gallery-dialog.config.js
```diff
@@ -0,0 +1,132 @@
...
+        link.appendChild(text);
+        link.title = "Click to test the modal";
+
+        // Add necessary demo attributes.
+        link.setAttribute('href', '#dialog');
+        link.setAttribute('class', 'ecl-link');
+        link.setAttribute('data-ecl-dialog', 'ecl-dialog');
+
+        // Show the link
+        document.body.append(link);
```
emeryro: On my IE the link isn't displayed.

framework/components/ecl-galleries/ecl-gallery-dialog/ecl-gallery-dialog.config.js
```diff
@@ -0,0 +1,132 @@
...
+        });
+      `,
+    },
+    dialog_settings: {
+      extra_attributes: [
+        { name: 'aria-labelledby', value: 'dialog-title' },
+        { name: 'aria-describedby', value: 'dialog-description' },
+        { name: 'id', value: 'ecl-dialog' },
+        { name: 'aria-hidden', value: 'true' },
+      ],
```
emeryro: Shouldn't we set the opening link as a part of the dialog component (instead of creating it manually with js)?

framework/components/ecl-galleries/ecl-gallery-base/ecl-gallery-base.scss
```diff
@@ -0,0 +1,96 @@
...
+}
+
+@include ecl-media-breakpoint-down(sm) {
+  .ecl-gallery__item {
+    float: left;
+    width: 50%;
+  }
+}
+
+@include ecl-media-breakpoint-between(sm, md) {
```
emeryro: Why is this? When resizing the window we switch from 4 column to 1, then to 2.
To manage responsive design with the grid you should better use dedicated classes (ecl-col-sm-*, ecl-col-md-*, etc.)

framework/components/ecl-carousels/ecl-carousels.scss
```diff
@@ -0,0 +1,120 @@
...
+  padding: map-get($ecl-spacing, 'l');
+}
+
+.ecl-carousel__meta {
+  margin-bottom: map-get($ecl-spacing, 'l');
+}
+
+// Every link has specific styling.
+.ecl-carousel__image-information a,
+.ecl-carousel__meta-link {
```
emeryro: You should use the "ecl-link--inverted" modifier

emeryro (changes requested):
Several questions and remarks, see below

framework/components/ecl-carousels/carousels.js
```diff
@@ -0,0 +1,175 @@
...
+    ).textContent = `${currentSlide + 1} / ${slides.length}`;
+  }
+
+  function showImageInformation() {
+    // Reset/Hide all.
+    const infoAreas = queryAll('[data-image]');
+    // If anything is visible.
+    if (infoAreas) {
+      // eslint-disable-next-line
+      infoAreas.forEach(area => (area.style.display = "none"));
```
kalinchernev: The logic at the moment is different for 2 reasons, but I can change if necessary:
- image hide/show is with opacity + translation since if there is no opacity 0 on hidden items, you will actually see the images next to each other like a series of frames. Another technique, of course, is to have items absolutely positioned in a relative container and display block only the featured. I personally decided to keep the series of frame like way so that if some day an animation is needed, it will definitely look much better
- also I decided to use js for part of the css transformation as they don't make sense in no-js mode

I am open for feedback in overall

framework/components/ecl-carousels/carousels.js
```diff
@@ -0,0 +1,175 @@
...
+    goToSlide(currentSlide + 1);
+    alignCenter();
+    setOffset();
+    announceCurrentSlide();
+    showImageInformation();
+  }
+
+  // Attach controls to a carousel.
+  function addCarouselControls() {
+    const navControls = document.createElement('ul');
```
kalinchernev: It's true it would have been easier just placing 2 buttons independently as many carousels do, though this [part of the best practices](https://www.w3.org/WAI/tutorials/carousels/functionality/#add-navigation-buttons) I implemented.

framework/components/ecl-carousels/carousels.js
```diff
@@ -0,0 +1,175 @@
...
+
+    addClass(navControls, 'ecl-carousel__controls ecl-list--unstyled');
+
+    navControls.innerHTML = `
+      <li>
+        <button type="button" class="ecl-icon ecl-icon--left ecl-carousel__btn ecl-carousel__btn--previous">
+          <span class="ecl-sr-only">Previous</span></button>
```
kalinchernev: Again, I agree. I was following the [accessibility suggestions](https://www.w3.org/WAI/tutorials/carousels/functionality/#add-navigation-buttons) to show these buttons only when js is enabled, as they don't have value otherwise having all images displayed.
Making those translatable in other systems will surely need another iteration to include a lib/framework specific globals for managing strings translation.

framework/components/ecl-carousels/carousels.js
```diff
@@ -0,0 +1,175 @@
...
+  function init() {
+    alignCenter();
+    addCarouselControls();
+    addLiveRegion();
+    goToSlide(0);
+    announceCurrentSlide();
+    showImageInformation();
+
+    // Re-align on resize.
+    window.addEventListener(
```
kalinchernev: Well noticed, I added it.

framework/components/ecl-carousels/ecl-carousels.config.js
```diff
@@ -0,0 +1,95 @@
...
+    _demo: {
+      scripts: `
+        document.addEventListener('DOMContentLoaded', function () {
+            ECL.carousels();
+        });
+      `,
+    },
+    extra_attributes: [
+      {
+        name: 'aria-labelledby',
```
kalinchernev: I left them out of template mainly because I imagine that some carousels will not have the headings. And the `aria-labelledby` will be wrongly used. So, having them both optionally avoids the possibility of wrong use because of a missing variable.

framework/components/ecl-carousels/ecl-carousels.config.js
```diff
@@ -0,0 +1,95 @@
...
+    ],
+    heading_title: 'Example gallery carousel',
+    heading_attributes: [
+      {
+        name: 'id',
+        value: 'ecl-carousel__heading',
+      },
+      {
+        name: 'class',
+        value: 'ecl-heading--3 ecl-sr-only',
```
kalinchernev: Thanks, I changed it. Since footer, every element should get our namespaces correctly, even if sometimes it's not really necessary, but it's good for consistency! :)

framework/components/ecl-carousels/ecl-carousels.config.js
```diff
@@ -0,0 +1,95 @@
...
+          label: 'Share',
+        },
+        description: `<p>Nulla consequat massa quis enim. Donec pede justo.</p>
+          <p>In enim justo, rhoncus ut, imperdiet a, venenatis vitae, justo. Integer tincidunt. <a href="#" title="Cras dapibus">Cras dapibus</a>. Vivamus elementum semper nisi.</p>`,
+        copyright: '© Copyright 2',
+      },
+      {
+        image: {
+          src: 'http://lorempixel.com/600/1200/business/3/',
+          alt: 'First item',
```
kalinchernev: Ooops, thanks! :)

framework/components/ecl-carousels/ecl-carousels.scss
```diff
@@ -0,0 +1,120 @@
...
+ * Carousel
+ * @define carousel ; weak
+ */
+
+.ecl-carousel {
+  align-items: stretch;
+  background-color: #000;
+  display: flex;
+  flex-direction: column;
+  margin: 0;
```
kalinchernev: Honestly, I tried, but the result was still different and not very likable - i.e. the content in the modal doesn't get fully displayed. I will try further.

framework/components/ecl-carousels/ecl-carousels.scss
kalinchernev: Thanks - added, additional styles out of the icons removed. The states are not 1:1 match, but it's really ok for me.

framework/components/ecl-carousels/test/spec/carousel.js
kalinchernev: True! I don't know how to solve this... I increased the timeout to 4 seconds and they still won't appear except 1 browser which I least expect to be working ... The only idea I have is to have the images in the static assets folder to that they are accessible offline, but it I think that means they will go in the build and dist. I'm open for ideas here.

framework/components/ecl-galleries/ecl-gallery-base/ecl-gallery-base.scss
kalinchernev: To be honest, I don't remember exactly, but I think it was indirect port of some of the breakpoints [in the css](https://github.com/ec-europa/ec-europa-theme/blob/master/sass/components/_gallery.scss#L138). Will double check this!

framework/components/ecl-galleries/ecl-gallery-dialog/ecl-gallery-dialog.config.js
kalinchernev: The link is created manually, as I think the trigger of a dialog/overlay should be independent from the js behavior. Meaning, the trigger could be a link, a button, or any other element with the attributes necessary for it to be a trigger. And because a trigger link is one example, I created it with JS.

framework/base/helpers/dom.js
```diff
@@ -3,3 +3,21 @@ export const queryAll = (selector, context = document) =>
   [].slice.call(context.querySelectorAll(selector));
 
 export default queryAll;
+
+export const hasClass = (element, cssClass) =>
```
kalinchernev: the `hasClass` is actually a helper for the other 2 functions I added hoping to make some common tasks easier. I really don't mind removing them?

emeryro: well the two other functions could be removed too, but so you will have to update your code

framework/components/ecl-carousels/ecl-carousels.twig
```diff
@@ -0,0 +1,91 @@
...
+    Conventional map of possible extra attributes for the wrapper.
+
+  - heading_attributes: (iterable)
+    Conventional map of possible extra attributes for the header.
+
+  - heading_title: (string)
+    Optional title of the heading. If skipped, heading will not be shown.
+
+  - carousel_images (iterable)
+    A list of `@ec-europa/ecl-images` components.
```
kalinchernev: Oh, actually the structure is not the component, well noticed! Thank you!

framework/components/ecl-carousels/ecl-carousels.twig
kalinchernev: Let's discuss on slack if the heading would be an obligatory element and it's defaults.

framework/components/ecl-galleries/ecl-gallery-dialog/.npmignore
kalinchernev: Will try to make something useful.

framework/components/ecl-galleries/ecl-gallery-dialog/ecl-gallery-dialog.config.js
kalinchernev: Will check, thanks

framework/base/helpers/dom.js
kalinchernev: Sure, removed.

framework/components/ecl-galleries/ecl-gallery-base/ecl-gallery-base.scss
yhuard: This point is still unclear. I would use the grid if needed and remove this custom code (`float: left; width: 50%;` ...)

yhuard (commented):
I don't understand the utility of the `dark` dialog 🤔

framework/components/ecl-galleries/ecl-gallery-base/ecl-gallery-base.scss
kalinchernev: I didn't manage find a good reason to keep these. Will push update in a bit.

framework/components/ecl-carousels/ecl-carousels.scss
kalinchernev: Approach changed a bit in the meanwhile, but I used the instructions in latest push.

framework/components/ecl-galleries/ecl-gallery-dialog/.npmignore
kalinchernev: Tests added.

yhuard (approved):
Thanks for the hard work!

emeryro (approved):

PR #374, opened 2017-08-10
feat(docker): dockerize the environment - no issue

yhuard:
# PR description

Check the README to see how it works.


yhuard:
@kalinchernev honestly I haven't put much time in it, it could be interesting to see how we could easily share a volume between the container and the host ;)

PR #383, opened 2017-08-16
docs(banners): update documentation - INNO-766

yhuard:
# PR description

Just a documentation update


kalinchernev:
Forced merge as the tests were failing because of missing connection to the sauce labs, not because of regressions.

PR #396, opened 2017-08-18
chore(packages): upgrades - noissue

kalinchernev:
# PR description

Updates in presets and wdio frameworks, no breaking changes described in the latter.


yhuard (approved):
Quick tip: when updating all the packages, make sure to delete both `yarn.lock` AND `node_modules`, otherwise yarn will first look into `node_modules` before searching on the internet for updated sub-dependencies

PR #408, opened 2017-08-21
fix(template): missing variable in if-logic - no issue

3wh:
Missing variable in the "is not" logic causes the footer twig template to break (at least in twig-2.4.3/PHP)

# PR description

Bugfix for template

## QA Checklist

Commit done from my smartphone (GH down), no particular QA has been done

- [ ] `package.json` is up-to-date and `@ec-europa/ecl-base` is part of the dependencies
- [ ] I have provided tests


yhuard:
I was wondering why Travis builds keep failing but it makes sense: since the PR comes from an external repository (`3wh`), the build fails. Otherwise, it would be possible to steal our env variables. So it's just a security mechanism :) 

We will by-pass Travis for this one but we need to think about how we can accept external PRs.

Thanks @3wh for your contribution! Much appreciated 👍

PR #468, opened 2017-10-12
feat(list-items): add highlight - INNO-828

emeryro:
# PR description

Add highlight modifier and fix responsive display issue

## QA Checklist

- [x] I have provided tests
- [x] I have filled the README.md file (at least a few lines)
- [x] My component is listed in the root README


HAL-Patch-INNO:
Deploy preview ready!

Built with commit a1c775c58f18c153fabbd60c2538b65aad051a59

https://deploy-preview-468--europa-component-library.netlify.com

framework/components/ecl-listings/ecl-listings.scss
```diff
@@ -57,11 +57,3 @@
     }
   }
 }
-
-@include ecl-media-breakpoint-up(lg) {
```
kalinchernev: Well noticed

framework/templates/ecl-templates-listings/ecl-templates-listings.twig
```diff
@@ -290,6 +290,38 @@
...
       ]
     } %}
+
+    <h3 class="ecl-heading ecl-heading--h3">Highlight listing (three columns)</h3>
+
+    {% include '@ec-europa/ecl-listings' with {
+      variant: 'three-columns',
+      items: [
+        {
+          variant: 'highlight',
```
kalinchernev: A `screen_reader` would be useful containing the title rather than the default.

framework/components/ecl-list-items/ecl-list-items.scss
```diff
@@ -45,9 +45,41 @@
...
+  flex-shrink: 0;
   height: auto;
   margin-right: map-get($ecl-spacing, 'm');
   max-height: 5.55rem;
   max-width: 5.55rem;
   width: auto;
 }
+
+// Highlited item
+.ecl-list-item--highlight .ecl-list-item__link {
```
kalinchernev: Would it be better to update the twig logic adding additional checks and simplifying the markup rather than going more specific and weak on the rules here?

Quickly checking the markup, there are more elements which are not used rather than used:

```twig
<li class="ecl-list-item ecl-list-item--highlight">
  <a href="" class="ecl-link ecl-list-item__link">
    <div class="ecl-u-sr-only">List item</div>
    <div class="ecl-list-item__primary">
      <img alt="State of the European Union 2017" src="../../assets/demo_images/nature-demo-1.jpg">
    </div>
    <div class="ecl-list-item__body">
      <!-- unused -->
      <div class="ecl-list-item__meta">
        <div class="ecl-meta ecl-meta--highlight">
        </div>
      </div>
      <h3 class="ecl-list-item__title ecl-heading ecl-heading--h3">State of the European Union 2017</h3>
      <!-- unused -->
      <p class="ecl-list-item__detail ecl-paragraph">
      </p>
      <!-- unused -->
      <div class="ecl-list-item__action">
      </div>
    </div>
  </a>
</li>
```

kalinchernev (changes requested):
It's using the correct components.
It could be better to update the template rather than the specificity of the css though

framework/components/ecl-list-items/ecl-list-items.scss
emeryro: Well, it is the same for other variants. At first I used a specific markup for each variant, but that ended in duplicated code. So I decided to use the same markup for all variants, and decide what to display in the context.
What could be improved here is to add tests in markup to not display empty html tags. Or we can do a specific markup for each variant, as you proposed.

To be discussed...

framework/templates/ecl-templates-listings/ecl-templates-listings.twig
emeryro: indeed

framework/components/ecl-list-items/ecl-list-items.scss
kalinchernev: Which-ever is more convenient for you.
Making some additional checks to avoid empty tags will be safer also in terms that it won't have breaking changes, so i'm ok with that and having the specificity if it's necessary.

degliwe (approved):

PR #508, opened 2017-11-28
fix(tables): refactor tables headers - INNO-852

raduchiriac:
# PR description

We found out that in ecl-tables.js there's an assumption that tables always have leading row. This is not always the case.

## QA Checklist

- [ ] `package.json` is up-to-date and `@ec-europa/ecl-base` is part of the dependencies
- [ ] I have provided tests


HAL-Patch-INNO:
Deploy preview ready!

Built with commit 0d15b62d2d36d66182d2fc912596d2870659be79

https://deploy-preview-508--europa-component-library.netlify.com

kalinchernev:
Some notes on how to ensure that existing behaviors (what are they) would be useful.

drupol:
Oooh yes!

kalinchernev:
Ticket INNO-852 contains links to internal specifications and discussions around the table. In short, this pull request changes the selector for the JavaScript behaviors so that they are triggered only in specific cases, when content editors know what they are doing. The JS part should literally be an enhancement,  and should not be run at all cases. (confirmed with Davide after Design Team suggested me to do so)

Existing responsive styles are kept, they already solve the responsive per-se, no need for changes.

Tests are also updated to demonstrate the specific behaviors of data in specific responsive tables.

framework/components/ecl-tables/ecl-tables.js
```diff
@@ -6,7 +6,9 @@
 
 export function eclTables(elements = null) {
   const tables =
-    elements === null ? document.getElementsByClassName('ecl-table') : elements;
+    elements === null
```
yhuard: Why not `elements || document.getElementsByClassName('ecl-table--responsive')`?

kalinchernev: Good point, the default value for the argument is there indeed.

kalinchernev:
Updates pushed, ready for another review.
- documentation updated, Filipa pinged, maybe we can wait 1-2 days if feedback
- added collator as the other components
- had to add new preview template for tables, as I didn't find another way to access the demo scripts from which contact in the default preview one
- reference screenshots updated to also demonstrate the difference between default and enhanced variants

PR #546, opened 2018-01-18
fix(comments): operation links as a list - INNO-886

raduchiriac:
# PR description

Instead of a simple DIV with <a> we wrap the comments actions into UL > LI

## QA Checklist

* [x] `package.json` is up-to-date and `@ec-europa/ecl-base` is part of the dependencies
* [x] I have checked the dependencies
* [x] I have given the fractal status “ready” to my component
* [x] I have declared `@define mycomponent` in the SCSS file
* [x] I have specified `margin: 0;` on the CSS component
* [x] I have provided tests
* [x] I follow the naming guidelines
* [x] the component supports composition
* [x] there are no hardcoded strings (all content come from the context)
* [x] I have filled the README.md file (at least a few lines)
* [x] My component is listed in the root README
* [x] My PR has the right label(s)


HAL-Patch-INNO:

Built with commit 4971a4df502cab43d1814073491c0b76dd8d0ad0

https://deploy-preview-546--europa-component-library.netlify.com

framework/components/ecl-comments/ecl-comments.twig
```diff
@@ -36,7 +36,7 @@
 <div class="{{ _css_class }}" {{ _extra_attributes }}>
   {% if image is defined %}
   <div class="ecl-comment__image-wrapper">
-    <img class="ecl-comment__image" src="{{ image.src }}" alt="{{ image.alt }}"">
+    <img class="ecl-comment__image" src="{{ image.src }}" alt="{{ image.alt }}">
```
yhuard: Good catch!

framework/components/ecl-comments/ecl-comments.twig
```diff
@@ -49,17 +49,23 @@
       <p class="ecl-paragraph">{{ body }}</p>
     </div>
     {% if operations is defined %}
-    <div class="ecl-comment__operations">
+    <ul class="ecl-listing ecl-comment__operations">
```
yhuard: The actual class is `ecl-list ecl-list--inline`, not `ecl-listing`. ;)

"Listings" is more a layout component (https://ec-europa.github.io/europa-component-library/components/detail/ecl-listings), I know it can be confusing...

framework/components/ecl-comments/ecl-comments.twig
```diff
@@ -49,17 +49,23 @@
...
       {% if operations.reply is defined and operations.reply.visible %}
-        <a class="ecl-link ecl-comment__reply" href="{{ operations.reply.href }}">{{ operations.reply.title }}</a>
+        <li class="ecl-list-item">
```
yhuard: https://ec-europa.github.io/europa-component-library/components/detail/ecl-list-items: "List items are intended to be used only in listings component."

Since we don't use the listing component, we don't need this class. `<li>` is enough

framework/components/ecl-comments/ecl-comments.twig
```diff
@@ -49,17 +49,23 @@
...
+          <a class="ecl-link ecl-list-item__link ecl-comment__reply" href="{{ operations.reply.href }}">{{ operations.reply.title }}</a>
```
yhuard: And here, we don't need `ecl-list-item__link` either (same for the lines below)

yhuard:
It's technically possible (using `:first-letter`) to enforce the case but I think it's a decision that needs to be taken by the DT. Either it's free and the user chooses the case (nothing imposed by the component), or it's enforced by the component. What you suggest here is kind of a strict rule and will impact all the users of the component. It's not my call  😉

(aside) Interesting article on the topic:
https://medium.com/@jsaito/making-a-case-for-letter-case-19d09f653c98

PR #553, opened 2018-01-26
chore(sass-compilation): allow compilation with webpack - INNO-904

yhuard:
# PR description

This PR aims at making sure we can easily compile the ECL (JS + SCSS) with webpack.

For this purpose, we provide a really basic webpack setup demo (in `demo/webpack`) with a single entry point importing the main files from `@ec-europa/ecl-preset-full`. When running `yarn build` from this folder, we should expect to get a `dist` folder containing both the compiled CSS and JS, as well as the assets (images, fonts).

When you load the JS + CSS in the fractal theme (manual step here, please update the partial accordingly), you should get the exact same result as when you load the "original" JS + CSS build with the `ecl-builder`.

Note: `package.json` and SCSS files have been updated to make sure the compilation can actually happen. By default, when you import a module, webpack will try to resolve the `module` and then the `main` fields from `package.json`. If a component contains both a JS and a SCSS file, the `main` or `module` will resolve to the JS file -> in the SCSS, we need to make sure the file we import is the SCSS one and not the JS one.


HAL-Patch-INNO:

Built with commit 701dcb5c7238e7ca409def3b4b6443e4d68ee607

https://deploy-preview-553--europa-component-library.netlify.com

demo/webpack/webpack.config.js
```diff
@@ -0,0 +1,86 @@
...
+const postcssNormalize = require('postcss-normalize');
+
+module.exports = {
+  entry: ['./index.js'],
+  output: {
+    filename: '[name].js',
+    path: path.resolve(__dirname, 'dist'),
+    library: 'ECL',
+    libraryTarget: 'umd',
+  },
```
raduchiriac: Add a `stats: 'errors-only',` (_opinions?_)

raduchiriac (dismissed):

yhuard:
I've just run ALL the tests, it went fine. Only 4 screenshots had to be removed (unused).

```
Number of specs: 356
1086 passing (6274.40s)
```

PR #584, opened 2018-03-05
refactor(naming): rename packages - INNO-927

yhuard:
TODO: rename twig templates and fractal config files as well


kalinchernev:
@yhuard thank you for the the suggestion - it's ok for me.
If previous discussions around generic holds, tests should be removed from there, this could be done in the main pull request together with other follow-ups.
@ec-europa/inno do you all agree on naming here?

degliwe:
Good for me

PR #601, opened 2018-03-13
refactor(sass): import sass files correctly - INNO-973

yhuard:
~~Needs https://github.com/ec-europa/ecl-toolkit/pull/63 to be merged first~~

Now ready. 

This PR makes use of the new ECL toolkit 🎉 Please do not use ~ as a prefix when importing files from the node_modules, it won't work. Instead, configure the `includePaths` in the `ecl-builder.config.js` to map to your `node_modules` folder location.

Also, there's no more black magic when resolving the Sass imports. It's 100% pure node-sass


HAL-Patch-INNO:

Built with commit be13bdf7fb323c8a38ead2161f6c9725806b9cc1

https://deploy-preview-601--europa-component-library.netlify.com

demo/webpack/webpack.config.js
```diff
@@ -62,7 +65,13 @@ module.exports = {
                 keepQuery: true,
               },
             },
-            'sass-loader?sourceMap',
+            {
+              loader: 'sass-loader',
+              options: {
+                sourceMap: true,
```
kalinchernev: Just wondering whether the demo is to be treated as a production-friendly example where the regular env variable `NODE_ENV === 'production` is to be respected to toggle the source maps?

packages/ecl-fractal-theme/assets/scss/fractal.scss
```diff
@@ -1,17 +1,17 @@
 // Mixins
-@import '~sass-mq/mq';
+@import 'sass-mq/mq';
 
 // Import normalize.css
 @import-normalize;
```
kalinchernev: Please change this one as well

kalinchernev (commented):
So far tests run green and preview looks ok, I don't see any blockers, few comments above

packages/ecl-fractal-theme/assets/scss/fractal.scss
yhuard: good catch, thanks!

demo/webpack/webpack.config.js
yhuard: In a real project, you usually have `webpack.dev.js` and `webpack.prod.js`, where this kind of choice (to include the sourcemap or not) is made. I didn't want to overcomplicate the simple example here, that's all 👼

Issue #593, opened 2018-03-08
Inpage navigation not working correctly in mobile layout


jcastro-malet:
Hello
I've found this problem with my code, but I checked that it also happens in the demo. The test was done with Google Chrome 62

### Expected behavior
When you access an inpage navigation using a mobile device, a sticky select-item should appear, letting you navigate through the different headers. I understand this is the expected behaviour, because as I explain later, I've been able to see this behaviour at some points.

### Actual behavior
Right now if you access the demo from a mobile device, the select-item doesn't appear.
The demo only shows this select-item if you access the page from a normal browser, and then change the page's size, for example using the development tools of Google Chrome

### Steps to reproduce the issue
**Case 1:**
- Access the inpage navigation demo from a mobile device: https://ec-europa.github.io/europa-component-library/components/preview/ecl-navigation-inpages
- Go up and down with the scroll, checking that the select-item doesn't appear.

**Case 2:**
- Access the inpage navigation demo using Google Chrome: https://ec-europa.github.io/europa-component-library/components/preview/ecl-navigation-inpages
- Open the development tools (F12). Press Ctrl-Shift-M to change to mobile layout
- Go up and down with the scroll, checking that the select-item used for navigation appears in the upper part of the page.
- Now refresh the page (F5)
- The select-item disappears, so you can't navigate through the headers.

Thank you very much

yhuard:
Thank you @jcastro-malet for reporting this issue! I can confirm I can reproduce it.
We have added this issue to our backlog and we will keep you posted on any future developments.

jcastro-malet:
@yhuard thank you very much

jcastro-malet:
I tested the solution, but I've seen that although with this change the mobile version starts working, I'm finding some problems with IE11. When the left side of the inpage navigation goes sticky, then the right side moves to the left, taking the same space as the sticky block.
The only solution I've found is going back to the original code, that didn't worked with the mobile layout.

yhuard:
You're right... On IE (no `position: sticky` support), the left column gets `position: absolute` when sticky, that's why the right column takes its place.

I'll try to find a better solution. Thanks for reporting it (again)

yhuard:
I've come up with 2 "solutions":
- the first one: we acknowledge that IE doesn't support `position: sticky` and we consider this behavior as an optional enhancement (-> progressive enhancement)
[...]

PR #613, opened 2018-03-22
fix(inpage-nav): fix mobile issue - INNO-943

yhuard:
Fix the inpage navigation issue reported in https://github.com/ec-europa/europa-component-library/issues/593. It was basically a structural issue. 

I also "updated" the generic component and its systems-related counterparts.


HAL-Patch-INNO:

Built with commit b5337ee9b7e8109f5bbb8870b6b292275d05c86c

https://deploy-preview-613--europa-component-library.netlify.com

src/systems/ec/ec-component/ec-component-navigation/ec-component-inpage-navigation/ec-component-inpage-navigation.twig
```diff
@@ -7,41 +7,4 @@
...
-          'extra_classes': 'ecl-inpage-navigation__link',
-          'variant': 'standalone',
-          }|merge(link|default({}))
-        %}
-      </li>
-      {% endfor %}
-    </ul>
-  </div>
-</nav>
+{% include '@ecl/generic-component-inpage-navigation' with { system: 'ec' } %}
```
emeryro: I understand the need to pass the system to parent template, but this is somehow redundant (we are using an EC component, so we should use system: ec in any cases).
Maybe we could find a way to pass this parameter automatically (could be done in another task, not a blocker)

emeryro (commented):
changes are working fine.
One remark: inpage navigation is only working when displaying the component on full screen (not in the demo). But I guess this is acceptable. Maybe we should just add a line about it in the readme to prevent confusion.

Also see comment below

yhuard:
Interesting, I didn't test it in the "demo" iframe. That's weird

src/systems/ec/ec-component/ec-component-navigation/ec-component-inpage-navigation/ec-component-inpage-navigation.twig
yhuard: This was the simplest solution I came up to. If you have a better idea, I'll welcome it :)

emeryro: not yet, I'll open a ticket to think about it

yhuard:
It seems the class "ecl-inpage-navigation--sticky" is not applied in the demo

emeryro (approved):
review ok

PR #615, opened 2018-03-23
refactor(reset): remove margin: 0 from the reset - INNO-947

yhuard:
This is the first PR of the "let's get rid of the reset" series.

It also gets rid of the useless "ECL editor" presets. I'm wondering if we should remove the ecl-preset-base as well or not. I don't think anybody uses it anyway


HAL-Patch-INNO:

Built with commit a2c12df4aaf5b30b1203cc2195bda4437bc8911c

https://deploy-preview-615--europa-component-library.netlify.com

emeryro (approved):
review ok

yhuard:
CI ok (timeouts only)

PR #620, opened 2018-03-26
fix(reset-img): remove img from reset - INNO-945

yhuard:
Question: should we keep the following code in the preset "full"?

```css
img {
  height: auto;
  max-width: 100%;
}
```

Or only use `.ecl-image`?


HAL-Patch-INNO:

Built with commit 51da21a5b613307c6ba78c2ebf798fa72d477a92

https://deploy-preview-620--europa-component-library.netlify.com

degliwe (approved):

PR #629, opened 2018-04-04
fix(reset): remove typography rules from reset - INNO-949

yhuard:
It's the final PR regarding the reset 🕐🎉🍾

Two options:
1. apply a class on the most top-level acceptable DOM element (easy way: body), like I did here with `.ecl-typography`
2. refactor almost all the components to explicitly apply the font rules, _à la_ https://github.com/carbon-design-system/carbon-components/search?l=SCSS&q=font-family&type=

I went with option 1 because it was faster but option 2 is more robust. It could be part of the normalization process :D cc @emeryro

### DIFF
--- a/src/systems/ec/ec-style/ec-style-typography/ec-style-typography/README.md
+++ b/src/systems/ec/ec-style/ec-style-typography/ec-style-typography/README.md
@@ -1,3 +1,7 @@
 # Typography
 
 Be careful when applying the `.ecl-typography` class.
+
+If you embed the ECL in an existing website, don't apply this class on the `body` or it will affect the existing styles. Apply it on the top-level ECL element.
+
+If the ECL is your main framework, then you can add the `.ecl-typography` class to the `body` element.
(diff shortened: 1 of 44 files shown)


HAL-Patch-INNO:

Built with commit 9b66ae0a7d50a2b4b62451a9f13b43d71afba4f1

https://deploy-preview-629--europa-component-library.netlify.com

yhuard:
TBD: merging `typography-heading` and `typography-paragraph` into typography` while moving `typography-list` into a component

src/systems/ec/ec-style/ec-style-typography/ec-style-typography/README.md
```diff
@@ -0,0 +1,3 @@
+# Typography
+
+Be careful when applying the `.ecl-typography` class.
```
kalinchernev: Could you please elaborate?

kalinchernev (commented):
The approach seems ok to me. Applying a class selectively on particular element to scope the styles looks better than applying the rules on individual components. It'll also be easier to maintain the system in this way.

Only 2 unclear items:
- is it worth keeping empty twig templates in generic components
- readme file warns about usage but doesn't give details

kalinchernev (approved):
Thanks!

PR #637, opened 2018-04-12
feat(focus): add focus styles on links - INNO-1017

yhuard:
These changes only add the focus styles on `.ecl-link`. In the future, if links don't have that class, we will have to decide whether we should add the missing class or if we should embed the styles in the component itself.


HAL-Patch-INNO:

Built with commit a3395ca9817292c9f1838d657bb458d7ca69ae58

https://deploy-preview-637--europa-component-library.netlify.com

kalinchernev (approved):

PR #643, opened 2018-04-16
fix(twig): normalize twig templates - INNO-1019

yhuard:
Update components from accordion to feedback-message. 
Skip dropdown and expandable which are not reusable and will need a full upgrade.
Also, apply generic updates (`extra_class` -> `extra_classes`)


emeryro (changes requested):
Reviewed components from accordion to feedback-message

emeryro (dismissed):

emeryro (approved):

PR #680, opened 2018-04-27
fix(inpage-nav): fix inpage-nav - INNO-943

yhuard:
The goal of this PR is to finally have the `inpage navigation` component working correctly.

IMHO, the sticky behavior should be considered as a progressive enhancement. If the browser supports `position: sticky`, then the nav will be sticky. Otherwise, it will just go away with the rest of the page.


HAL-Patch-INNO:

Built with commit 9e5b224580a29d9775f52f9d94ed860d89d8e539

https://deploy-preview-680--europa-component-library.netlify.com

PR #697, opened 2018-05-31
chore(build): copy generic templates on change - INNO-994

yhuard:
To test:
- checkout the branch
- `yarn`
- `yarn start:ec` or `yarn start:eu`
- update a generic component -> the template is copied into the `{ec|eu}/_imports` folder, but the view is not rebuilt
- open the related system component's config.js, add or remove a comment -> this will trigger a fractal build of the component and update the view in the browser

### DIFF
--- a/tools/ec-styleguide/package.json
+++ b/tools/ec-styleguide/package.json
@@ -8,23 +8,27 @@
   "scripts": {
     "build:preset":
       "cd ../../src/systems/ec/ec-preset/ec-preset-website && npm run build",
-    "clean": "rimraf ./static/ec-preset-website",
+    "clean": "rimraf ./static/ec-preset-website ../../src/systems/ec/_imports",
     "copy:preset":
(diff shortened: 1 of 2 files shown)


HAL-Patch-INNO:

Built with commit f5cc165401ec7490a6ef70062cf70121a6426415

https://deploy-preview-697--europa-component-library.netlify.com

kalinchernev (approved):

kalinchernev:
Merged, it's a good step forward.

In my case, I had to run `echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p` to avoid issues with watchers' limitations.

I also tried to import the already created instance of fractal and work with the [server](https://fractal.build/guide/web/server) programatically having the idea of using the [reload](https://browsersync.io/docs/api#api-reload) available in the background, but it doesn't work because requiring the fractal configuration re-runs the file and there are issues with symlinks.

So for now, it's good addition and we'll simply keep a file with component's config or other asset open to edit to trigger the reload.

PR #750, opened 2018-08-02
feat(themes): separate themes from design tokens - INNO-1148

yhuard:
Themes now live in `src/systems/{system}/themes/{theme}` and they depend on their respective design tokens. Useless presets have been removed as well. Only `full` and `website` presets remain.

We still have to figure out how to associate themes and presets and how themes affect components. In a nutshell, we should have a "build" which is a theme applied to a preset and then run visual regression tests on this "build" if that makes sense.


HAL-Patch-INNO:

Built with commit d420b16b880163711040953ff4e915038857f923

https://deploy-preview-750--europa-component-library.netlify.com

yhuard:
Maybe we can take inspiration from https://github.com/opentable/design-tokens

PR #758, opened 2018-08-08
feat(colors): add more colours - no issue

thanasisrig:
Added
---------
 COLOR_NOTIFICATIONS_BLUE: "#006fb4"
 COLOR_NOTIFICATIONS_GREEN: "#467a39"
 COLOR_NOTIFICATIONS_ORANGE: "#f29527"
 COLOR_NOTIFICATIONS_RED: "#da2131"
 COLOR_BLUE_5: "#f2f5f9"


degliwe:
The naming is too specific. token should be values without too much context.
ex:  with `COLOR_ORANGE_100: "#f29527"` it will be mapped to a variable called `COLOR_NOTIFICATIONS_WARNING: COLOR_ORANGE_100` hence giving context to this value.

The proposed name `COLOR_NOTIFICATIONS_ORANGE` is actually mixing nature of the information (this token is a color, orange) and the meaning (this token is used for warning notifications)

For instance, `COLOR_NOTIFICATIONS_BLUE` should be together with the other blue like `COLOR_BLUE_5`.

thanasisrig:
@degliwe Feel free to use whatever name works best for you!

yhuard (approved):

PR #781, opened 2018-08-23
chore(website): switch CSS depending on the system - INNO-1173

yhuard:
- Depending on the system the user browses, the right stylesheet will be loaded -> the React component will be correctly themed
- Make blockquote React component generic

Current solution: inject the styles dynamically as soon as you enter `/ec` or `/eu`. This is not ideal but I need to figure out how to dynamically load and **unload** styles with https://github.com/webpack-contrib/mini-css-extract-plugin.


HAL-Patch-INNO:

Built with commit cf7bbfd304e3e5f8c5d037d0b891da098dc8ee6c

https://deploy-preview-781--europa-component-library.netlify.com

degliwe (approved):

PR #794, opened 2018-09-03
fix(language-list): put spanish in second position - no-issue

degliwe:
Simply move spanish in second position as per:
http://publications.europa.eu/code/pdf/370000fr.htm#langues


HAL-Patch-INNO:

Built with commit 67b8555a7b26ef046da3ca9352818ba6cdd6c4f9

https://deploy-preview-794--europa-component-library.netlify.com

PR #818, opened 2018-09-21
refactor(website): use 12-columns grid - INNO-1231

yhuard:
Replace the previous with a 12-columns grid. The containers keep the same width. The breakpoints are the same. It's not exactly the ECL grid.


HAL-Patch-INNO:

Built with commit d17f68db26bb42ff900a7cba905478f179ea7636

https://deploy-preview-818--europa-component-library.netlify.com

PR #826, opened 2018-09-28
feat(iframe): showcase iframes - no issue

yhuard:
Display showcase in iframe, reusing storybook's iframe.

Can be seen here: https://deploy-preview-826--europa-component-library.netlify.com/ec/components/hero-banner/showcase/

Pros:
- really easy to implement
- add "Fullscreen" link

Cons:
- doesn't work without JS
- height is not automatically adjusted as it is when it's an actual React component

If we replace all our showcases by iframes, we can get avoid injecting/ejecting the EC/EU CSS -> smaller website, faster to load.

Possible enhancement: display the iframe in a device look https://www.w3schools.com/howto/howto_css_devices.asp

Notes:
-----

- This branch is actually based on `feat/hero-INNO-1233` but in order to run netlify, I had to open a PR based on next-v2. If we merge it, we should merge it into `feat/hero-INNO-1233` first.

- [react-snap](https://github.com/stereobooster/react-snap) messes up with storybook... If we could prevent it from rendering some specific routes, or if storybook used the history API, then we'd be good

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,7 +2,6 @@
   command = "yarn run dist"
   publish = "dist/website"
 
-
 [build.environment]
   NODE_VERSION = "8"
   YARN_VERSION = "1.9.4"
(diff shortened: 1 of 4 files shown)


HAL-Patch-INNO:

Built with commit 880fbe3a10e6a67e1b931118b2bb3c52cf15513f

https://deploy-preview-826--europa-component-library.netlify.com

degliwe (approved):

PR #837, opened 2018-10-03
docs(Footer-EC): provide first version

kibos11:
This is the first commit of this content (just a wiki mirror)


HAL-Patch-INNO:

Built with commit f3030653c16dcf4d2b11d9013f642095b08a2294

https://deploy-preview-837--europa-component-library.netlify.com

src/systems/ec/specs/components/footer/docs/usage.md
```diff
@@ -1 +1,31 @@
-Do's and don'ts
+The footer appears at the bottom of all  ec.europa.eu pages.
```
yhuard: There are 2 spaces between `all` and `ec.europa.eu`, typo?

src/systems/ec/specs/components/footer/docs/usage.md
```diff
@@ -1 +1,31 @@
...
+
+It is divided into 2 functional areas:
+* the main content area
+* the secondary content area
+
+![Image](http://inno-ecl.s3.amazonaws.com/media/images/EC/Footer%20EC/Footer_Corporate.png)
```
yhuard: `http` -> `https`

src/systems/ec/specs/components/footer/docs/usage.md
kibos11: About branch name convention I apologize, I couldn't find any doc for that (and it was late night so I couldn't ask you). I will follow the pattern that you mention. Also, by mistake I create some branches that you can delete (George-Blockquote-Usage, kibos11-patch-1, kibos11-patch-2).

src/systems/ec/specs/components/footer/docs/usage.md
kibos11: Ok, thanks. I will make all the new link https from now on. I have another question. Can we use SVG in some case instead of .png or .jpg? Could be convenient in some cases.

yhuard: Yes, you can use SVG. It's often an even better option actually 👍

src/systems/ec/specs/components/footer/docs/usage.md
yhuard: You don't have to apologize for that, there is no documentation nor real convention, but rather commonly used patterns. As long as you don't use parentheses in the branch name, it should be just fine 😉

yhuard: I'll delete the branches you've mentioned

src/systems/ec/specs/components/footer/docs/usage.md
kibos11: That's great. Makes things much easier for me!

src/systems/ec/specs/components/footer/docs/usage.md
kibos11: It was my experiment with SVGs. Since it's ok to use them then you can merge it I guess.

PR #843, opened 2018-10-05
fix(cards): remove min-width and fix react components - INNO-1187

yhuard:
As discussed on slack:
- we remove the min-width
- we use a fluid width (100%)
- we keep the max-width


HAL-Patch-INNO:

Built with commit d9d81f7f03f6de37ec0d30b3631094a9818f897e

https://deploy-preview-843--europa-component-library.netlify.com

emeryro (approved):

PR #852, opened 2018-10-08
docs(links): add version01

kibos11:
There are many things that need a fix but the wiki page is still under work I think.

### DIFF
--- a/src/systems/ec/specs/components/link/docs/usage.md
+++ b/src/systems/ec/specs/components/link/docs/usage.md
@@ -1,1 +1,53 @@
-Do's and don'ts
+## When to use
+
+When you need to direct users to a different page or to provide access to further information.
+
+## Link types
+* in-line text link
+  * internal links
+  * external links
+* stand-alone link (have no underline)
+  * more link
+  * expandable link
+  * social media link
+
+## When to use this component
+
+### in-line link
+
+When links are inserted within paragraph text blocks
+Style blue, underlined (displayed in white colour if the background colour is blue)
+
+#### internal link
+
+when links point to URLs inside the ec.europa.eu domain
+style 
+underlined link
+
+#### external link
+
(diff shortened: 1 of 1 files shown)


src/systems/ec/specs/components/link/docs/usage.md
```diff
@@ -1 +1,53 @@
...
+  * more link
+  * expandable link
+  * social media link
+
+## When to use this component
+
+### in-line link
+
+When links are inserted within paragraph text blocks
+Style blue, underlined (displayed in white colour if the background colour is blue)
```
yhuard: Style specs should go into the "style" tab, not "usage"

src/systems/ec/specs/components/link/docs/usage.md
```diff
@@ -1 +1,53 @@
...
+
+Used on the Expandable component
+
+#### social media link
+
+Used on the Social media - Follow. and Social media - Share. components
+
+## When not to use 
+
+For actions or important links to other pages -  use buttons instead
```
yhuard: Please do not confuse links with buttons. Links are used for navigation. Buttons are used for actions (e.g. submitting/resetting a form).

- http://www.webaxe.org/proper-use-buttons-links/
- https://medium.com/simple-human/but-sometimes-links-look-like-buttons-and-buttons-look-like-links-9b371c57b3d2

yhuard (changes requested):

src/systems/ec/specs/components/link/docs/usage.md
degliwe: https://a11y-101.com/design/button-vs-link

kibos11: I will make sure the relevant person to check it and updated. I just copy paste the content that I have in the wiki. Consider it as a draft.

PR #871, opened 2018-10-10
docs(cards): add version01

kibos11:
The images are pngs in this pages since there was a lot of photos


HAL-Patch-INNO:

Built with commit 6c1c980fa338ad4e9247b77638e85a51d5484288

https://deploy-preview-871--europa-component-library.netlify.com

degliwe (changes requested):
1. do not use tables for layout, this break the responsivness of the page (look at the colours page for reference)
2. Landscape cards does not exists (yet) as we don't have specs for it.
3. this is a showcase page, not a usage page.

PR #884, opened 2018-10-15
fix(image): add example image - INNO-1257

emeryro:
# PR description

The image we were using as example has been removed. A new image has been set; not it doesn't rely on external site.


src/systems/ec/specs/components/card/demo/data--card.js
```diff
@@ -1,9 +1,10 @@
+const demoImage = require('@ecl/ec-resources-example/example-image.jpg');
```
yhuard: Specs are meant to be generic. You're importing a jpg in a JS file. It implies people will have to use webpack or another module bundle in order to load this file...

Can you move the "require logic" outside of this file? Only the path to the image should remain here. :)

HAL-Patch-INNO:

Built with commit 944a60eb6d3c600500e061bec74d63a8651510fd

https://deploy-preview-884--europa-component-library.netlify.com

yhuard (changes requested):

src/systems/ec/resources/example/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "private": true,
+  "name": "@ecl/ec-resources-example",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "2.0.0-alpha.1",
```
yhuard: please update version number :)

edit: everywhere

yhuard (changes requested):

emeryro:
Ready for another round

yhuard (approved):

PR #899, opened 2018-10-17
chore(logo): remove standalone logo - INNO-1260

yhuard:


HAL-Patch-INNO:

Built with commit a8bee12bf9555917941b87972f9a4ee503e3713e

https://deploy-preview-899--europa-component-library.netlify.com

src/systems/ec/implementations/react/packages/site-header/SiteHeader.jsx
```diff
@@ -70,7 +87,6 @@ SiteHeader.defaultProps = {
   logo: {
     title: '',
     alt: '',
-    language: '',
```
emeryro, line 73: We should already add the possibility to have different languages (as you did for EU).
For now we only have the EN version of the logo, but it will be translated for sure.

emeryro (changes requested):

src/systems/ec/implementations/react/packages/site-header/SiteHeader.jsx
yhuard: We'd have a new variable, not used anywhere, eslint complaining about that... Are you sure you really want this? 

If the designers provide translated logos, then we could update the component. It won't be a BC since we would just add a new property.

emeryro (approved):

PR #944, opened 2018-10-29
feat(message): apply new specs - INNO-1276

emeryro:
# PR description

Apply [new specifications](https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Messages+-+Design+Specifications) for message component


HAL-Patch-INNO:

Built with commit b963eeb2731c6cc3e0a4416fb493295c6458c08d

https://deploy-preview-944--europa-component-library.netlify.com

yhuard (changes requested):
- the generated code outputs a "submit" type for the close button, it should be `type="button"`
- the "close" behaviour is missing. I think there should be a basic generic implementation of this behaviour using the `data-ecl-*` attribute that simply removes the element from the DOM on click. TBD if needed

src/systems/ec/implementations/vanilla/packages/ec-component-message/ec-component-message.js
```diff
@@ -0,0 +1,56 @@
...
+
+  destroy() {
+    if (this.attachClickListener && this.close) {
+      this.close.removeEventListener('click', this.handleClickOnClose);
+    }
+  }
+
+  handleClickOnClose() {
+    // IE way to remove a node...
+    if (this.element.parentNode) {
```
yhuard, line 48: The action on the DOM should be handled by React (as it's done in the breadcrumb) -> you could pass a new parameter to the constructor's options which is a callback that is called (if defined) in `handleClickOnClose`

yhuard (dismissed):

degliwe (approved):

PR #950, opened 2018-11-07
feat(social-media-share): add component - INNO-1264

planctus:
# PR description

This pull request is introducing the social-media-share component as an independent one.
It has only been placed in the ec system, although i believe this is a common one, likely to be shared with the EU system as well.
Being this a component "served" by webtools i had many doubts regarding the way we would, should provide it, for now and only for demo purposes i added the needed styles for the buttons to appear as they should, this would require the definition of the list of supported social networks to be able to define the style (background, basically) for all of them.
I added the vanilla, the react and the specs packages, it works apparently fine in the storybook but when i move into the website i see that the "playground" and "full screen" links are not working.

Furthermore i noticed that because of the outline we have on focused elements we needed (only for demoing) purposes, to add an element to create a bit of space on the left, this was "easy" in storybook, could not find a proper way to do the same with the website.
Same thing for an "additional" css to be used only for demoing purposes, i was able to set it for storybook in a "clean" way, didn't find a way to easily do it in the website.

There is an issue with the google icon, the "default" svg for that comes with a "fill" #4B4B4C and i don't think this is the expected behavior.

## QA Checklist

* [ ] I have provided tests


HAL-Patch-INNO:

Built with commit 4642c6050e5485bd1388b56e5c10fccb47103881

https://deploy-preview-950--europa-component-library.netlify.com

src/systems/ec/implementations/react/packages/social-media-share/stories/Index.jsx
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs } from '@storybook/addon-knobs';
+
+import demoContent from '@ecl/ec-specs-social-media-share/demo/data';
+
+import SocialMediaShare from '../SocialMediaShare';
+
+import '@ecl/ec-specs-social-media-share/demo/demo.css';
```
degliwe: Can't we get rid of it?

src/systems/ec/implementations/react/packages/social-media-share/SocialMediaShare.jsx
```diff
@@ -0,0 +1,52 @@
...
+import socialSprite from '@ecl/ec-resources-social-icons/dist/sprites/icons-social.svg';
+
+import Link from '@ecl/ec-react-component-link/Link';
+
+const SocialMediaShare = ({ links, className, ...props }) => {
+  const classNames = classnames(className, 'ecl-social-media-share');
+
+  return (
+    <div {...props} className={classNames}>
+      <ul className="ecl-social-media-share__list">
```
degliwe, line 14: Does not provide the label "Share this page". Without it it's impossible to know what those buttons are for.

src/systems/ec/implementations/react/packages/social-media-share/stories/Index.jsx
planctus: mmh..yes, sure. is that i wanted to fix that little problem regarding the outline in storybook and then i started wondering about how to to do this without affecting the component, that's the reason why of this demo.css. Plus with that i was fixing a bit the demo for the google icon which is "broken" but surely this needs to be fixed properly by replacing the svg, i'll push without this "helper" css, these two issues will pop up, though.

src/systems/ec/implementations/react/packages/social-media-share/SocialMediaShare.jsx
planctus: Yeah, but that was on purpose, because in the specifications there is no "share this page". I'll add it, anyhow.

planctus:
@degliwe, i pushed the requested changes, hopefully.. ;)

src/systems/ec/implementations/react/packages/social-media-share/SocialMediaShare.jsx
degliwe: You are right. the specs are incomplete. I've reported it (see my comment on the specs page)
Let's wait for new specs.

emeryro:
Drone reports a few issues to be fixed: https://drone.fpfis.eu/ec-europa/europa-component-library/2951/5 (don't mind the warnings)

planctus:
Thanks @emeryro for noticing that, first time i access the "drone" builds. Should be fixed now, but i see that when i go "fullscreen" in the social media share component, i lose the description...

planctus:
The g+ icons have been replaced with new ones, the ones specified here: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Social+Media+-+Share+%7C+Design+Specifications
Also the ones in the "social icons" here:
https://webgate.ec.europa.eu/CITnet/confluence/pages/viewpage.action?pageId=837977554
are going to be updated, the new ones completely replace the old ones, basically.

emeryro (dismissed):
A few margin tricks should be added to fix following cases:

- horizontal display, lots of items: vertical spacing missing. This is not in the specs, so either we ask for a specific value, or we use the one from vertical display.
image

- vertical display: due to bottom margin on items, the components takes extra space below (red area). This could be a little tricky to solve... maybe negative margin could work...
image

emeryro:
This is almost good, just a few last things:

- put the text `(work in progress)` in src/website/src/pages/ec/components/social-media-share/docs/usage.md (similar to what we have on other components)
- set the status to `ready` in src/website/src/pages/ec/components/social-media-share/index.md
- add the EU version (copy/paste of everything)

src/systems/ec/implementations/react/packages/social-media-share/package.json
```diff
@@ -0,0 +1,38 @@
+{
```
yhuard, line 1: Please make the package private ;-) (check the other React packages, they're all private now)

src/systems/ec/implementations/vanilla/packages/ec-component-social-media-share/ec-component-social-media-share.scss
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  .ecl-social-media-share__link {
+    display: flex;
+    font: $ecl-font-s;
+  }
+
+  // Icon
+  /* stylelint-disable-next-line order/order */
+  @include ecl-icon-rollover(ecl-social-media-share);
```
yhuard: why not including it first? (and thus avoiding the stylelint error)

src/systems/ec/implementations/vanilla/packages/ec-component-social-media-share/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "name": "@ecl/ec-component-social-media-share",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "2.0.0",
+  "description": "ECL EC Social Media Share",
+  "style": "ec-component-social-media-share.scss",
+  "sass": "ec-component-social-media-share.scss",
+  "dependencies": {
+    "@ecl/ec-base": "^2.0.0",
+    "@ecl/ec-component-icon": "^2.0.0"
```
yhuard: it depends on `@ecl/ec-component-link` too, doesn't it?

src/systems/ec/specs/components/social-media-share/demo/data.js
```diff
@@ -0,0 +1,96 @@
...
+      ],
+    },
+    {
+      href: '/example',
+      label: 'Facebook',
+      className: 'ecl-social-media-share__link--facebook',
+      variant: 'standalone',
+      iconPosition: 'before',
+      icon: [
+        {
```
yhuard, line 31: the problem with this structure is that you have to remember that the first item is the default icon and the second icon is the icon on hover... but what happens if you add a third icon?

yhuard (changes requested):
A few remarks here and there (also valid for EU) but overall it looks good :)

planctus:
@yhuard, thanks for the review, i should have made the requested changes, for the icons we are not relying on the array index anymore, see if it's better like this.

yhuard (approved):

PR #951, opened 2018-11-07
feat(social-media-follow): add modifier for webtools - INNO-1264

planctus:
# PR description

In this pull request the social-media-share component is implemented as a modifier of the existing social-media-follow component.
This is more or less the way it is currently defined in the DT styleguide, the component is just the same, its style changes a bit and there's no vertical variant, but yeah, this would be theoretically the way i would define this, as a variant and not an independent component.
Being this served by an "external service", though, this could be confusing, furthermore the main component would need to be redifined, if the "action" is the modifier the component should be social-media and then we would have --follow --follow-vertical and --share, something like that.
But again, webtools is only serving the third one, doesn't proabably make much sense to define a "bigger" component because we need to provide this to a system where only one of the modifier "styles" would be needed.
That's why you have also https://github.com/ec-europa/europa-component-library/pull/950 where the component is defined as an independent one.

## QA Checklist

* [ ] I have provided tests


HAL-Patch-INNO:

Built with commit 19abdeb60558aa1ae48daa956f043de0a69933e7

https://deploy-preview-951--europa-component-library.netlify.com

degliwe:
I would not go for a modifier even the components seems similar at first for the following reasons:

- The component is called social-media-follow hence having a variant dedicated to share (the opposite to follow) is misleading so it would imply renaming the whole component.
- The variant name "webtools" isn't appropriate, ECL is supposed to be agnostic. Also "webtools" doesn't convey any meaning.
- The 2 components look similar today but they serve 2 purpose with different guidelines, do's and don't, they might also evolve differently.

planctus:
@degliwe, i perfectly agree, just the read the description of this pr where i highlighted the fact that this was not a final "proposal", what i mean is that going this way we would have to change the component definition from the basis, as i wrote there.

PR #952, opened 2018-11-08
feat(gallery): apply new specs - INNO-1162

emeryro:
# PR description

add component gallery using provided [specifications](https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Gallery+Design+Specifications) (and also updates mades in the comments, but not visible yet on the mockups)

A few remarks concerning the display of images:
To achieve required design (always full width on each line), there is a css property `object-fit` which works just fine. But of course, it is not supported by IE. 
So display is conform to specifications on modern browsers, and partially degraded on IE.

**Important to notice:** 
this PR add the possibility to have icons before the label in Buttons (needed for the overlay).


HAL-Patch-INNO:

Built with commit 6576ba413dd79543759ed398209ef8698579d242

https://deploy-preview-952--europa-component-library.netlify.com

degliwe (changes requested):
I would suggest to remove the img all together and not use object-fit.
Adapt the div "ecl-gallery__image-container" to be like
`<div class="ecl-gallery__image-container" role="img" aria-label="Description of the image" title="Tooltip for users not using assistive technologies" />`

https://www.w3.org/WAI/GL/wiki/ARIATechnique_usingImgRole_with_aria-label_forCSS-backgroundImage

emeryro:
The problem is that the `` tag is used to set the dimension of the image container (keeping its ratio).
If we remove it, display will be altered (unless we manually do calculations in javascript)
For instance, here is the current display:
image
And here it is without `` (but with a fixed height):
image

In the overlay we would also have to make a request to get the image, just to get its dimension

degliwe:
- update the markup to use https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption instead of multiple divs
- provide a yellow outline for keyboard navigation
- to me it seems acceptable to use object-fit, the fallback for IE can simply be to display black bands horizontal or vertical.

src/systems/ec/implementations/react/packages/gallery/src/GalleryItem.jsx
```diff
@@ -0,0 +1,69 @@
...
+              size="l"
+              className={classnames(
+                item.icon.className,
+                'ecl-gallery__description-icon'
+              )}
+            />
+          )}
+        </figcaption>
+
+        <div className="ecl-gallery__meta" data-ecl-gallery-meta hidden>
```
degliwe: I would suggest to get rid of this div and rely on data-attributes
https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

degliwe (changes requested):
- looks good on IE11

degliwe (changes requested):
Almost there
- add focus trapping https://css-tricks.com/a-css-approach-to-trap-focus-inside-of-an-element/
- please use images for the demo to make it ore obvious how this is altering the content
- ESC key should close the modal

src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.scss
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  .ecl-gallery__image-container {
+    background-color: $ecl-color-black-100;
+    flex-grow: 1;
+    margin: 0;
+  }
+
+  .ecl-gallery__item-link:focus {
+    outline: none;
```
degliwe, line 42: Why not keeping the outline on the link instead of the image container?

degliwe (changes requested):
- focus lost on modal close look how bootstrap deal with it https://github.com/twbs/bootstrap/issues/12364
- user expect keyboard space key to open the modal (as well as return key)

degliwe (changes requested):
- there's now 2 gallery in the menu of the website (Gallery and Galleries)

src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.scss
emeryro: Because it just doesn't work in this case. I guess the image is displayed on top of the link's outline.

yhuard (changes requested):
There's an issue on the mobile layout:

image

src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
```diff
@@ -223,10 +226,21 @@ class Gallery {
     } else {
       this.overlay.removeAttribute('open');
     }
+
+    // Focus item
+    this.selectedItem.focus();
```
degliwe, line 260: what about users closing the modal with esc key?

degliwe (changes requested):

src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
emeryro: it's already there

degliwe (dismissed):

src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
```diff
@@ -0,0 +1,299 @@
...
+    this.overlayNext = queryOne(this.overlayNextSelector, this.overlay);
+
+    // Polyfill to support <dialog>
+    this.isDialogSupported = true;
+    if (!window.HTMLDialogElement) {
+      this.isDialogSupported = false;
+    }
+
+    // Hack to detect when focus gets out of the modal
+    this.overlay.addEventListener('transitionend', () => {
```
yhuard: This hack doesn't work well. If you click anywhere on the modal, the focus will always come back to the close button. 

Have you considered other options like https://github.com/davidtheclark/focus-trap or https://hiddedevries.nl/en/blog/2017-01-29-using-javascript-to-trap-focus-in-an-element or https://allyjs.io/tutorials/accessible-dialog.html?

yhuard (changes requested):

degliwe (dismissed):

yhuard (changes requested):
The download link is missing the `download` attribute and `href`s are empty

yhuard (changes requested):
The download `href` and probably the share `href` as well should be tied to the object item, not to the gallery

yhuard (approved):

PR #958, opened 2018-11-19
docs(grid): adding new content

kibos11:
Please check the changes. If you need to correct something, please do. If we can change the text format of the dynamic content of the breakpoints and grid data it will be great. We can remove all caps and make it appear like normal text maybe?


HAL-Patch-INNO:

Built with commit 0917e5be7f8413c9cd5be6c20da0af944d81884f

https://deploy-preview-958--europa-component-library.netlify.com

emeryro (approved):

PR #983, opened 2018-12-19
feat(back-to-top): apply specs - INNO-1244

emeryro:
# PR description

Update footer to add [back to top link](https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Button+-+Back+to+top+-+Design+specifications)

### DIFF
--- a/src/systems/ec/implementations/react/packages/footer/Footer.jsx
+++ b/src/systems/ec/implementations/react/packages/footer/Footer.jsx
@@ -4,8 +4,28 @@ import classnames from 'classnames';
 
 import Link from '@ecl/ec-react-component-link/Link';
 
-const Footer = ({ identity, sections, common, className, ...props }) => (
-  <footer {...props} className={classnames(className, 'ecl-footer')}>
+const Footer = ({
+  backToTop,
+  identity,
+  sections,
+  common,
+  className,
+  ...props
+}) => (
+  <footer
+    {...props}
+    className={classnames(className, 'ecl-footer', {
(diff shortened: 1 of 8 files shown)


HAL-Patch-INNO:

Built with commit 6fa16e3f6396ed27b5f21f1ce91cc7de445ded7f

https://deploy-preview-983--europa-component-library.netlify.com

src/systems/ec/implementations/react/packages/footer/Footer.jsx
```diff
@@ -4,8 +4,28 @@ import classnames from 'classnames';
...
+  ...props
+}) => (
+  <footer
+    {...props}
+    className={classnames(className, 'ecl-footer', {
+      'ecl-footer--custom': identity && identity.title,
+    })}
+  >
+    {/* Back to top */}
+    {backToTop && (
```
yhuard: The condition is not good enough. If the user provides `backToTop: {}`, then the condition is truthy (test in your console: `if ({}) console.log('the test passed')`, you'll see). 

A better check would look like `backToTop && backToTop.label && backToTop.label.length > 0`

yhuard (changes requested):

src/systems/ec/implementations/react/packages/footer/Footer.jsx
emeryro: if think that `backToTop && backToTop.label` should be enough

yhuard (approved):

PR #984, opened 2018-12-19
feat(checkbox): apply new specifications - INNO-1307

planctus:
Design specs: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Checkboxes+-+Design+specifications

- EC version: https://deploy-preview-984--europa-component-library.netlify.com/ec/components/forms/checkbox/code/
- EU version: https://deploy-preview-984--europa-component-library.netlify.com/eu/components/forms/checkbox/code/

I'm honestly a bit confused, i followed this: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/03.+Checkbox
but i had implemented specs for the help text, error text, that i don't find anymore, apparently...
Anyway, we miss the svg for the presentation in the website and i "blindly" applied the concept of "grouping" the same way as it was done for the radios, but there is no "checkbox group" thing in aria, for example.


HAL-Patch-INNO:

Built with commit e15161e507b51e400183f375a3f85d4d7718190b

https://deploy-preview-984--europa-component-library.netlify.com

HAL-Patch-INNO:

Built with commit 455031e49f737c00275001cc0ac29745020ca455

https://deploy-preview-984--europa-component-library.netlify.com

src/systems/ec/implementations/react/packages/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,96 @@
...
+        <div className="ecl-feedback-message">{invalidText}</div>
+      )}
+      {helperText && <div className="ecl-help-block">{helperText}</div>}
+    </div>
+  );
+};
+
+Checkbox.propTypes = {
+  id: PropTypes.string.isRequired,
+  checked: PropTypes.bool,
```
yhuard: It's good that you follow the [uncontrolled components](https://reactjs.org/docs/uncontrolled-components.html) practice in the context of ECL. Since we don't want to impose React, it makes sense.
But `checked` controls the `defaultChecked` property of the checkbox, thus I'd rename it `defaultChecked` as well. `checked` is misleading, as a dev I expected it to control the `checked` property... Can you please rename it?

src/systems/ec/implementations/react/packages/checkbox/stories/Index.jsx
```diff
@@ -0,0 +1,61 @@
...
+storiesOf('Forms/Checkbox', module)
+  .addDecorator(withKnobs)
+  .add('default', () => {
+    const { label } = demo.unchecked;
+
+    return (
+      <Checkbox
+        id="checkbox-id"
+        label={label}
+        checked={boolean('checked', false)}
```
yhuard: Since `checked` is actually `defaultChecked`, changing it in the knobs does not affect the component. For this reason, I suggest removing the knob for `checked`.

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,91 @@
...
+  $fake-checkbox-size: 0.875rem
+) {
+  .ecl-checkbox {
+    align-items: center;
+    display: flex;
+    position: relative;
+  }
+
+  .ecl-checkbox__input {
+    cursor: pointer;
```
yhuard: why is there a `cursor: pointer` here? maybe it's inherited but we should not use a pointer cursor on checkboxes (the native checkbox html element uses the default cursor) :)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,91 @@
...
+    margin-left: 3px;
+    opacity: 0;
+    padding: 0;
+    position: absolute;
+    z-index: 1;
+  }
+
+  .ecl-checkbox__label {
+    color: $ecl-color-grey-100;
+    font-size: $ecl-font-m;
```
yhuard: `$ecl-font-m` = `normal normal 400 1rem/1.25rem Arial,sans-serif;`
it's not what you want to use here ;)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,91 @@
...
+    font-weight: normal;
+    padding: 2px 0 0 $ecl-spacing-xl;
```
yhuard: where do these `2px` come from?

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  .ecl-checkbox__icon {
+    border: $thickness-base solid $border-color;
+    cursor: pointer;
+    display: inline-flex;
+    left: 0;
```
yhuard: can you explain why `left: 0;`? I don't get it 👼

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,91 @@
...
+  .ecl-checkbox__label {
+    color: $ecl-color-grey-100;
+    font-size: $ecl-font-m;
+    font-weight: normal;
+    padding: 2px 0 0 $ecl-spacing-xl;
+  }
+
+  .ecl-checkbox__icon {
+    border: $thickness-base solid $border-color;
+    cursor: pointer;
```
yhuard: no `cursor: pointer` please :)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
yhuard: btw, opacity is set to 0, there's no need to use a custom cursor

yhuard: in the end, `.ecl-checkbox__input` should be the equivalent of a "visually hidden" component but you'll need to tweak the markup a bit: the `<label>` should encapsulate the `<input>`, the `<span>`s for the icon and the text label. https://www.w3schools.com/howto/howto_css_custom_checkbox.asp

src/systems/ec/implementations/react/packages/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,96 @@
...
+          name={name || undefined}
+          className="ecl-checkbox__input"
+          disabled={disabled}
+        />
+        <span className="ecl-checkbox__icon" />
+        <span className="ecl-checkbox__icon ecl-checkbox__icon--checked">
+          <Icon shape="ui--check" size="xs" />
+        </span>
+
+        {label && (
```
yhuard: `label` is mandatory (for accessibility reasons)
the label could be hidden or displayed (default: displayed)

src/systems/ec/implementations/react/packages/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,96 @@
...
+        <input
+          {...props}
+          type="checkbox"
+          id={id}
+          defaultChecked={checked}
+          name={name || undefined}
+          className="ecl-checkbox__input"
+          disabled={disabled}
+        />
+        <span className="ecl-checkbox__icon" />
```
yhuard: not sure why this span is needed... couldn't it all be done within 1 span? by adding a custom class to the following <Icon>, you could decide to show or hide the icon depending on the state of the checkbox ;)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
yhuard: or ignore it, since it might go away with the refactoring of the markup

yhuard (changes requested):

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
planctus: @yhuard, i honestly don't like the markup proposed by the w3c in this example, ok that is the w3c but still the reason why a label would contain an input looks cumbersome to me.
I mean that the solution always existed to circumvent the need for attributes, id for the "input" and "for" for the label, but we are saying that we consider the label a mandatory element and therefore the "for" attribute for it, as well.
Why do you think we would have to go this way, instead?

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
planctus: Can i ask you then what is the correct interpretation of a requirement defined as such? Font-size: M  :)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
planctus: It's just unneeded, so in a way or another it will go away :)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
yhuard: Actually, whether you put the `<input>` inside the `<label>` or not doesn't really matter (the `<input>` is visually hidden anyway). If the `<label>` encapsulates the `<input>`, the markup is lighter: no need for a `id` on the `<input>` and `for` on `<label>`. That's all. There's no strong need to go for one solution or the other, both are valid ;) 

But the label should at least contain the icon and the text (whether the text is visually hidden or not). Thus when you click on the icon, which is inside the label, you trigger the right checkbox automatically

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,88 @@
...
+  $thickness-max: 3px,
+  $border-color: $ecl-color-grey-25,
+  $border-color-focus: $ecl-color-yellow-100,
+  $border-color-checked: $ecl-color-blue-100,
+  $fake-checkbox-size: 0.875rem
+) {
+  .ecl-checkbox {
+    align-items: center;
+    display: flex;
+    position: relative;
```
emeryro, line 33: We should have a `margin:0` here (there is one for each default component css classes)

src/systems/ec/implementations/react/packages/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,87 @@
...
+        <input
+          {...props}
+          type="checkbox"
+          id={id}
+          defaultChecked={defaultChecked}
+          name={name || undefined}
+          className="ecl-checkbox__input"
+          disabled={disabled}
+        />
+        <span className="ecl-checkbox__icon">
```
emeryro: The class names are a little confusing; `ecl-checkbox__icon` is if fact the box, and `ecl-checkbox__checked` is the icon.
That's not a major issue, but we could have used for instance `ecl-checkbox__box` and `ecl-checkbox__icon`

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  .ecl-checkbox__input {
```
emeryro, line 36: the real (hidden) checkbox is in fact smaller than the box displayed. So if you don't click in the middle it won't work.
we could add a `transform: scale(1.5)` on it to fix that (or anything else if you find a cleaner solution)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,88 @@
...
+    opacity: 0;
+    padding: 0;
+    position: absolute;
+    z-index: 1;
+  }
+
+  .ecl-checkbox__label {
+    color: $ecl-color-grey-100;
+    font: $ecl-font-m;
+    padding-left: $ecl-spacing-xl;
```
emeryro: to be more accurate we should have something like `padding-left: calc({box_width} + {box_padding} + {box_border} + {spacing_s}`. That way we ensure that the spacing before the label will always be exactly as on the specs.
(by chance it will give the same value, 2rem, but this is a coincidence)

src/systems/ec/implementations/react/packages/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,87 @@
...
+          <Icon className="ecl-checkbox__checked" shape="ui--check" size="xs" />
+        </span>
+
+        <label
+          className={classnames(
+            'ecl-form-label ecl-form-label--inline ecl-checkbox__label',
+            {
+              'ecl-form-label--invalid': invalid,
+              'ecl-form-label--hidden': hideLabel,
```
emeryro: hidding the label breaks the display of the box:
image

src/systems/ec/implementations/react/packages/checkbox/stories/Index.jsx
```diff
@@ -0,0 +1,12 @@
...
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import demo from '@ecl/ec-specs-checkbox/demo/data';
+import Checkbox from '../Checkbox';
+
+storiesOf('Forms/Checkbox', module)
+  .add('default', () => <Checkbox {...demo.default} />)
+  .add('disabled', () => <Checkbox {...demo.disabled} />)
+  .add('invalid', () => (
+    <Checkbox {...demo.invalid} onClick={e => e.preventDefault()} />
```
emeryro: Why is there a onClick here?
Invalid checkbox should be usable. If the user can't check it, he can't solve the error

emeryro (changes requested):
Hi, a few more points have to be fixed (see below).

Also display is broken on IE11 (unfortunately we have to maintain compatibility...)
image

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
yhuard: there's a cleaner solution: the one I've suggested earlier
i.e. putting the icon within the label and visually hiding the input

planctus: ok, i went finally including the span with the icon into the label, but this introduced an issue with the "hidden-label" use case, i had to adapt then the behaviour of this hidden label element.

src/systems/ec/implementations/react/packages/checkbox/Checkbox.jsx
planctus: Yeah, now that we have the icon inside the label the problem shifted, but it was still there, meaning that with an hidden label we would have had also the fake checkbox hidden. Fixed by overriding the behaviour of the hidden label in the checkbox component.

src/systems/ec/implementations/react/packages/checkbox/stories/Index.jsx
planctus: Mmhh..but we don't provide this behaviour. We only define the styles and the markup associated, the preventDefault there was just to avoid the inconsistent result of a checked checkbox with errors still in place, if you started playing with it, it was not demoing the behavior of a form element with errors. 
So maybe this should be a generic onChange callback that removes the error classes and elements from all the components if they contain any.
This checkbox in react has been defined as an uncontrolled component, i am now handling this behavior with an onChange callback for this component only, don't know whether this is the right approach.

src/systems/ec/implementations/react/packages/checkbox/Checkbox.jsx
planctus: Agree, i switched to the classes you suggested.

emeryro:
It is already better.
A few things left:
- concerning the checkboxes with error, what I meant that it should not be disabled (sorry if that was not clear). I see that you implemented a function to remove the invalid state; we probably didn't have to go that far. Now that it is done, we can keep it.
- when there is no label, there should also be no right margin on the box
- still to be done:
  - https://github.com/ec-europa/europa-component-library/pull/984/files/70110d1ade2563859ff11fd00f5989bd40134a50#r248272713
  - https://github.com/ec-europa/europa-component-library/pull/984/files/70110d1ade2563859ff11fd00f5989bd40134a50#r248262459

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
planctus: @emeryro, the padding is not anymore in the label, therefore no need to make complex calculations, i think, we are now setting the requested distance between the "checkbox" and the label  via a margin on the checkbox (the one you want to be removed when the label is hidden.. ;) )

src/systems/ec/implementations/react/packages/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,102 @@
...
+  defaultChecked: false,
+  disabled: false,
+  helperText: '',
+  hideLabel: false,
+  invalid: false,
+  invalidText: '',
+  label: '',
+  name: '',
+  removeError: () => {
+    const inPage = queryOne('.ecl-checkbox--invalid');
```
yhuard: what's the rationale behind this?
to me, it's a business decision, not a UI one but maybe I'm wrong :)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
yhuard: https://a11yproject.com/posts/how-to-hide-content/ 
can you still visually hide the input? thanks :)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,103 @@
...
+
+    + label .ecl-checkbox__icon {
+      visibility: visible;
+    }
+  }
+
+  .ecl-checkbox__input:focus,
+  .ecl-checkbox__input:active {
+    + label > .ecl-checkbox__box {
+      border: $thickness-max solid $border-color-focus;
```
yhuard: for the focus, we should use `outline` instead of `border`

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,103 @@
...
+      padding: 0;
+    }
+  }
+
+  .ecl-checkbox__input:disabled {
+    cursor: not-allowed;
```
yhuard: why do you change the `cursor`? we should stick to the browser's defaults unless the designers specifically ask for this (and have good reasons, supported by research)

yhuard (changes requested):
hey @planctus 
I have a couple of remarks here and there but I think we're near the finish line 🍻

src/systems/ec/implementations/react/packages/checkbox/Checkbox.jsx
planctus: Eh, it is just a matter of being able to display in the demo the style specifications for a checkbox with errors.
Before i was preventing the input from being checked (to avoid the inconsistent styling that it would have been creating), after the romain's review i decided to apply this simple js to make the component usable in the website and in storybook.
I absolutely agree with you, this is here just to properly demo this version of the component.

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
planctus: Ok, it has to be said that all the other form elements defined in the ECL v2 are actually using a border, though.
About the input i now used the technique you suggested, but this is now preventing the checkbox from working in case the label is missing the "for" attribute, while the previous solution was a little bit more flexible, from this point of view.

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
planctus: Removed.

yhuard: thanks 👍

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
yhuard: You're completely right about the other components, we should fix them as well!
Thank you for reporting it, really appreciated 👍 

The `for` attribute is mandatory since the `input` is outside the `label`, and this is a good thing. The fact that it was working without it just because there's was some obscure hidden input above the icon was a dangerous side effect (am I exaggerating? 😄). Now, the `for` attribute also semantically links the label with the input. For screen readers, this is an absolute necessity (unless you have an `aria-label` but this is another topic 😆)

yhuard (changes requested):
Hi @planctus 

We'll try to move forward with this PR. If designers come back to us with better specs, then we'll make another PR to update our styles. ;)

Meanwhile, when the label takes multiple lines (as shown below), the display is broken:

image

Can you fix it? You might want to ask the designers if the checkbox should be vertically centered or kept on top, I don't know...

Thanks!

src/systems/ec/implementations/react/packages/checkbox/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "2.0.0",
+  "description": "ECL EC React Checkbox",
+  "dependencies": {
+    "@ecl/ec-base": "^2.0.0",
+    "@ecl/ec-react-component-icon": "^2.0.0",
+    "@ecl/ec-component-form-feedback-message": "^2.0.0",
+    "@ecl/ec-component-form-form-group": "^2.0.0",
+    "@ecl/ec-component-form-help-block": "^2.0.0",
+    "@ecl/ec-component-form-label": "^2.0.0",
+    "@ecl/ec-component-form-checkbox": "^2.0.0"
```
yhuard: `"@ecl/ec-component-form-checkbox": "^2.0.0"` is actually a `devDependencies` and `"@ecl/ec-react-component-icon": "^2.0.0",` is the only `dependencies`. All the other dependencies can be removed. Please see https://github.com/ec-europa/europa-component-library/blob/v2/src/systems/ec/implementations/react/packages/button/package.json 

PS: this is not a blocker since the React component is not published, but it's cleaner to make it right

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,102 @@
+/**
+ * Checkbox component
+ * @define checkbox
+ */
+
+// Import base
+@import '@ecl/ec-base/ec-base';
+
```
yhuard, line 8: since `.ecl-checkbox__icon` is applied to the `<Icon>` component, you must make sure that the SCSS code for the icon has already been included with `@include check-imports(('ec-component-icon'));` (as https://github.com/ec-europa/europa-component-library/blob/v2/src/systems/ec/implementations/vanilla/packages/ec-component-button/ec-component-button.scss#L9)

src/systems/ec/implementations/react/packages/checkbox/package.json
yhuard: Actually, if we want to be consistent, you should follow this example: https://github.com/ec-europa/europa-component-library/blob/v2/src/systems/ec/implementations/react/packages/text-input/package.json

Except that in your case, you have `"@ecl/ec-react-component-icon": "^2.0.0"` in your  `dependencies`. All the `@ecl/ec-component-*` dependencies are `devDependencies`. You don't need to depend on `@ecl/ec-base`

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@ecl/ec-component-form-checkbox",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "2.0.0",
+  "description": "ECL EC Form checkbox",
+  "style": "ec-component-form-checkbox.scss",
+  "sass": "ec-component-form-checkbox.scss",
+  "dependencies": {
+    "@ecl/ec-base": "^2.0.0"
```
yhuard: `@ecl/ec-component-icon`is also a dependency of this component ;-)

yhuard (changes requested):
Just a few dependency-related stuff to fix and we're good :)

yhuard:
I forgot to mention it, but it'd be better if the examples used the `name` attribute on the `<input>`

emeryro:
Component seems good to me.
Just one question: why is there knobs in storybook only for default checkboxes?

emeryro:
Wait... On IE there is still a display issue when label is long.
image

You just have to add a `flex-shrink: 0` I guess

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,104 @@
...
+    margin: 0;
+    position: static;
+    text-indent: -9999px;
+    width: auto;
+  }
+
+  .ecl-checkbox__box {
+    border: $thickness-base solid $border-color;
+    display: inline-flex;
+    height: $fake-checkbox-size;
```
yhuard: Ok, I have to ask; why didn't you use the width/height from the specs? (`1.25rem`)
It would have been much easier... Just use `1.25rem` with a `box-sizing: border-box`, remove the padding (why is there a padding anyway?), and leverage the `inline-flex` with `justify-content: center; align-items: center;`. This way, the checkbox will be use the values from the specs and the icon will be automatically centered.

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,104 @@
...
+    width: $fake-checkbox-size;
+
+    .ecl-checkbox__label--hidden & {
+      margin: 0;
+    }
+  }
+
+  .ecl-checkbox__icon {
+    fill: $ecl-color-white-100;
+    height: $fake-checkbox-size;
```
yhuard: Why do you override the icon's height?

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,104 @@
...
+    visibility: hidden;
+  }
+
+  .ecl-checkbox--invalid .ecl-checkbox__box {
+    border-color: $ecl-color-error;
+  }
+
+  .ecl-checkbox__input:checked {
+    + label .ecl-checkbox__box {
```
yhuard: Prefer `.ecl-checkbox__label` over `label`;)

src/systems/eu/implementations/react/packages/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,88 @@
...
+            {
+              'ecl-form-label--invalid': invalid,
+              'ecl-checkbox__label--hidden': hideLabel,
+              'ecl-form-label--hidden': hideLabel,
+            }
+          )}
+          htmlFor={id}
+        >
+          <span className="ecl-checkbox__box">
+            <Icon className="ecl-checkbox__icon" shape="ui--check" size="xs" />
```
yhuard: The specs say "S", why use "xs" then?

yhuard (dismissed):

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
planctus: well, in a default content-box implementation the padding was compensating the different border-size and computing the requested size, but i agree, it's simply better using border-box.

src/systems/eu/implementations/react/packages/checkbox/Checkbox.jsx
planctus: Yeah, i know, i now applied the size defined in the specs, the reason why i was using the smaller one is that when the checkbox is checked and focused it looks bad with the other one, also in the "mocks" it seems to be the smaller.

emeryro: The icons seems indeed a little too big now...
I don't see any information about its size in the specs (maybe I missed it...).

Edit: ok,indeed it is "S" in the specs. So let's keep it that way

emeryro: Fun fact: display of the focus is different on IE
image

But honestly I don't consider it as an issue.

emeryro (dismissed):
Almost good.
The last thing I see is for knobs: 
- there is no real value in changing `id` and `name`, so no need to put them in knobs
- order of knobs in not the same for `default` story; order is better in the other stories (I know, I may be a little perfectionist).

yhuard:
Before I forget, designers will update the design specs to provide examples where multiple checkboxes are used, with a group label, etc. As they are supposed to do with the radio group. How much could this impact your work?

planctus:
I'm honestly a bit confused, i followed this: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/03.+Checkbox
but i had implemented specs for the help text, error text, that i don't find anymore, apparently...
Anyway, we miss the svg for the presentation in the website and i "blindly" applied the concept of "grouping" the same way as it was done for the radios, but there is no "checkbox group" thing in aria, for example.

yhuard:
Ciao @planctus! We don't have the SVG for the checkbox, I'll report it to Jean ;)

Regarding checkbox groups: please make sure to use `<fieldset>` and `<legend>` (as @emeryro did in https://github.com/ec-europa/europa-component-library/pull/1222/files#diff-dad26ca421f114b31a8b82d21831b7e8R33)

I hope will we be able to unblock the form inputs in general today or tomorrow, I'll get back to you when we have news

planctus:
I made these changes, @yhuard but it doesn't really look like anything we could deliver or even finalise with the situation we have regarding generic style associated to the form elements. Meaning that the styles for the checkboxes can't be defined without conflicting with current generic styles, at the moment, plus i see much confusion regarding the classes we are using and the attempt to keep a BEM based approach, furthermore i see in the page with the "updated" specs just a screenshot with no info on it, impossible to use this as a reference for specs. So in the end i'm putting back on hold this pull request, i think it was not even time to restart working on it.

degliwe (changes requested):
Behavior should be closer to
https://deploy-preview-984--europa-component-library.netlify.com/ec/components/forms/radio/code/

- labels should not get selected on quick clicks on the checkbox see https://developer.mozilla.org/en-US/docs/Web/CSS/user-select
- disabled checkbox should not have hover state
- label should be inline not block, at the moment user can change the state of the checkbox by mistake by clicking/touching in any position horizontal to the checkbox and label (especially risky for mobile user who simply want to scroll). This problem also exist on the radio.

planctus:
Thanks @degliwe for the review, hope i fixed what needed to, i did this in the last commit:

1. Making the label flex-inline, so the input doesn't get selected on "proximity" clicks
2. Disabled checkboxes will not have style for :hover and :focus
3. I set the user-select to none for the label when the input is focused so the label doesn't get selected when double clicking
4. Re-organized the sass file and removed some preexisting code

Thanks

degliwe:
Few points

1. value attribute is leaking on the wrapping div
2. `user-select:none` don't need to be on the label with focus and active, but just on `ecl-checkbox__box`
3. I don't see the value of the optInOut in the code. it's just one checkbox why would add extra class for it?

planctus:
@degliwe , thank you again for reviewing, i removed this opt-in-out thingy, moved the user-select and handled the value attribute. Hope it's better.. ;)

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,146 @@
...
+  .ecl-checkbox--disabled {
+    .ecl-checkbox__label {
+      color: $label-color-disabled;
+      cursor: not-allowed;
+    }
+
+    .ecl-checkbox__box {
+      background-color: $box-background-disabled;
+      border-color: $box-border-color-disabled;
+      user-select: none;
```
degliwe:
```suggestion

```

src/systems/eu/implementations/vanilla/packages/eu-component-form-checkbox/eu-component-form-checkbox.scss
```diff
@@ -0,0 +1,146 @@
...
+  .ecl-checkbox--disabled {
+    .ecl-checkbox__label {
+      color: $label-color-disabled;
+      cursor: not-allowed;
+    }
+
+    .ecl-checkbox__box {
+      background-color: $box-background-disabled;
+      border-color: $box-border-color-disabled;
+      user-select: none;
```
degliwe:
```suggestion

```

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,146 @@
...
+
+  .ecl-checkbox__box {
+    align-items: center;
+    border: $border-width solid $border-color;
+    box-sizing: border-box;
+    display: inline-flex;
+    flex-shrink: 0;
+    height: $box-size;
+    justify-content: center;
+    margin-right: $ecl-spacing-s;
```
degliwe, line 59:
```suggestion
    margin-right: $ecl-spacing-s;
    user-select: none;
```

src/systems/eu/implementations/vanilla/packages/eu-component-form-checkbox/eu-component-form-checkbox.scss
```diff
@@ -0,0 +1,146 @@
...
+
+  .ecl-checkbox__box {
+    align-items: center;
+    border: $border-width solid $border-color;
+    box-sizing: border-box;
+    display: inline-flex;
+    flex-shrink: 0;
+    height: $box-size;
+    justify-content: center;
+    margin-right: $ecl-spacing-s;
```
degliwe, line 59:
```suggestion
    margin-right: $ecl-spacing-s;
    user-select: none;
```

degliwe (changes requested):
I don't find any reference of hidden label in the specs, where is this coming from?

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
```diff
@@ -0,0 +1,133 @@
...
+  $box-border-color-hover: $ecl-color-blue-100,
+  $invalid-border-color: $ecl-color-error,
+  $invalid-border-width: 2px,
+  $label-color: $ecl-color-grey-100,
+  $label-color-disabled: $ecl-color-grey-50,
+  $outline-color: $ecl-color-yellow-100,
+  $outline-width: 2px
+) {
+  .ecl-checkbox {
+    margin: 0;
```
degliwe, line 32:
```suggestion
    align-items: flex-start;
    display: flex;
    flex-direction: column;
    margin: 0;
```

degliwe (changes requested):
See my suggestion

src/systems/ec/implementations/vanilla/packages/ec-component-form-checkbox/ec-component-form-checkbox.scss
planctus: i agree, i see that we handle better the internal margin set by the helper text, this way.

degliwe (approved):
Thank you for your patience!
Good job!

degliwe (approved):

PR #988, opened 2019-01-07
chore(dependencies): update dependencies - no issue

yhuard:
It includes an update of `webpack-dev-server` which raised a security alert (low severity) and fixes many small ESLint errors (missing property checks).

Storybook has not been updated because 4.1.x breaks a11y (see https://github.com/storybooks/storybook/issues/4889). This bug has been fixed in 4.2.0-alpha.10 but the DLLs break react-snap (see https://github.com/storybooks/storybook/issues/5055). Once the issue with the DLLs is fixed, we can update to storybook 4.2.x.


HAL-Patch-INNO:

Built with commit 659b3a169fcb072a322d96b4f157f4d0cb1c4709

https://deploy-preview-988--europa-component-library.netlify.com

planctus (approved):

PR #990, opened 2019-01-09
fix(eu-footer): update layout - INNO-1311

yhuard:
This PR updates the markup of the EU footer but doesn't include any breaking change. If people don't update their markup, they will still have the same ol' footer, they won't be affected by this PR.


HAL-Patch-INNO:

Built with commit 19300654ebf19afa4ec1ea32c3fc7711f9f3db5b

https://deploy-preview-990--europa-component-library.netlify.com

emeryro (commented):
Overall it looks fine.
I'm just not a big fan of this `ecl-footer__columns--new-layout` modifier. If we change layout again will it be `ecl-footer__column--new-new-layout`?

Maybe we could have added the modifier on the `<footer>` tag instead of a inner tag, and use `--compact`, or `--small`, as it seems to be the goal of this new layout.

emeryro (approved):

PR #968, opened 2018-11-30
refactor(website): apply specs

emeryro:
# PR description

Apply specs for website found [on the wiki](https://webgate.ec.europa.eu/CITnet/confluence/pages/viewpage.action?pageId=858098090)

Mostly:
- Switch tabs to dropdown on mobile

PR #995, opened 2019-01-25
chore: update dependencies - no issue

yhuard:
Also:
- fix a bug introduced in https://github.com/ec-europa/europa-component-library/pull/968 which was making the website dev server crash (bad PropTypes)
- use React lazy and Suspense instead of react-loadable. You will get a warning in dev mode but this one will be fixed by react-router in the next release (nothing to worry about)
> Warning: Failed prop type: Invalid prop `component` of type `object` supplied to `Route`, expected `function`.
- unleash the bots! (stale + semantic-pull-requests)

Edit: unfortunately, prettier is pretty heavy on the client side

image

Ideally, it should be run at build time (check webpack, babel macros). Meanwhile, I'll revert the change of beautifier


HAL-Patch-INNO:

Built with commit 83f6640995a1a99c50b11f0bbe3896454a1035dc

https://deploy-preview-995--europa-component-library.netlify.com

degliwe (approved):

PR #1000, opened 2019-01-29
fix: use outline for form elements - INNO-1318

planctus:
# PR description

## QA Checklist

* [ ] I have provided tests


HAL-Patch-INNO:

Built with commit 2e97985d9e875d0b9d0f49aa1e4cd2d98857fd13

https://deploy-preview-1000--europa-component-library.netlify.com

emeryro (changes requested):
Using a macro here seems to be a good idea.
But the result isn't as expected; the values passed to the browser are strings so it isn't taken into account.
image

yhuard:
> Using a macro here seems to be a good idea.
> But the result isn't as expected; the values passed to the browser are strings so it isn't taken into account.
> image

we should have deleted this mixin long ago... it's not used anywhere and it makes the code less readable :(

planctus:
Well, i saw we had a mixin and i used that, it's true that the advantage of using it is questionable but since it is a repeated chunk of code the one we would use, a mixin is still the way i would go for.
Anyway, since this was already "deprecated",  i'll remove the usage of the mixin and the mixin file at this point ;)
Thanks for the reviews

emeryro (approved):

yhuard (approved):

Issue #1003, opened 2019-02-04
[RFC] Language list - INNO-1321


yhuard:
## Current implementation

In our current implementation, links look like this:

```html 
<a href="/example#language_ga" class="ecl-language-list__link ecl-link ecl-link--standalone">Gaeilge</a>
```

## What can we improve?

Obviously, 2 attributes are missing:
- `lang`: because "Gaeilge" is not in English (the language of the document)
- `hreflang`: because the link points to a page which is not in English

One third attribute can be discussed:
- `rel="alternate"`: 

> When used in combination with the `hreflang` attribute (with a value other than that of the document itself), rel="alternate" means a link to **a version of the current document**, but in the language indicated by the hreflang attribute.

— http://microformats.org/wiki/rel-alternate

To me, we should use this attribute in the "overlay language list", where links point to the same page in another language.
It should not be used in the "splash page language list", where links point to a completely different document than the current one. It seems to me that https://ec.europa.eu/info/ misuses the `rel="alternate"` attribute.

- Are we all on the same page here? 
- Are there any other attributes we could or should use?

emeryro:
Most SEO examples I see use `hreflang` in a `<link>` tag (in the `<head>`), but indeed it is valid in a `<a>` tag too (even if not widely used). I wonder if we shouldn't enforce the use of `<link>` instead...

`rel="alternate"` seems indeed fine for overlay (but again, it is commonly used in `<link>`)
 
But the `lang` parameter doesn't seems to be valid here. It isn't listed in allowed parameters
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a
https://www.w3schools.com/tags/tag_a.asp

yhuard:
> even if not widely used

what does make you think that?

> I wonder if we shouldn't enforce the use of `<link>` instead...

instead of `<a>`? 🤔 

> But the `lang` parameter doesn't seems to be valid here.

> Global Attributes
> The `<a>` tag also supports the Global Attributes in HTML.

— https://www.w3schools.com/tags/tag_a.asp 

-> https://www.w3schools.com/tags/ref_standardattributes.asp 😃

emeryro:
> what does make you think that?

Just the fact that almost all articles concerning `hreflang` I find uses `<link>` and I really have to specifically search for usage in `<a>` to find some.

Ok for the `lang`

emeryro:
[...]

PR #1007, opened 2019-02-07
refactor(language-list): improve markup - INNO-1321

emeryro:
# PR description

Apply improvements discussed in [RFC](https://github.com/ec-europa/europa-component-library/issues/1003)


HAL-Patch-INNO:

Built with commit 15481e1adedc163438c1df9b7817debf1fe36503

https://deploy-preview-1007--europa-component-library.netlify.com

src/systems/ec/implementations/react/packages/language-list/LanguageListItem.jsx
```diff
@@ -15,19 +15,26 @@ const LanguageListItem = ({ label, href, isActive }) => (
         size: 'xs',
       },
     }}
+    hrefLang={lang}
```
yhuard: `hrefLang` and `lang` attributes should be conditional, only appear when the `lang` parameter is a non-empty string

and if `isActive === false`? this is an open question. Is there any added value to adding these attributes if the link points to the current language?

yhuard (changes requested):

src/systems/ec/implementations/react/packages/language-list/LanguageListItem.jsx
emeryro: I think it doesn't hurt to have it for active item too.
It is less usefull than for other items, but information is still relevant.

yhuard (changes requested):
```js
    {...lang && {
      lang,
      hrefLang: lang,
      ...(isOverlay && { rel: 'alternate' }),
    }}
```
instead of 
```js
    {...lang && { hrefLang: lang }}
    {...lang && { lang }}
    {...isOverlay && { rel: 'alternate' }}
```
and then we're good :)

yhuard (changes requested):
> Object.keys: argument is not an Object

Error in IE

yhuard (approved):

PR #1013, opened 2019-02-13
fix(ecl-builder): update dependencies, fix scripts build - no issue

yhuard:
This PR contains some urgent fixes for v2.1. That's why it's not branched from v2-dev but from v2 instead.

Default exports don't make it to the final JS build. We have to use named exports.

```js
class Accordion {
```

becomes

```js
export class Accordion {
```

Also:
- import the ECL builder within this repository
- update dependencies, especially those for the ECL builder


HAL-Patch-INNO:

Built with commit f9657227544675c20497a050414a531cd1ffb609

https://deploy-preview-1013--europa-component-library.netlify.com

degliwe (approved):

PR #1034, opened 2019-03-13
feat(template): add language list - INNO-1389

emeryro:
# PR description

Add overlay behavior for language switcher (on templates only)

## QA Checklist

* [ ] I have provided tests


HAL-Patch-INNO:

Built with commit fceafa6a4a46ee9901320d91d84d14a165c5f144

https://deploy-preview-1034--europa-component-library.netlify.com

src/systems/ec/implementations/react/packages/campaign-page/src/CampaignPage.jsx
```diff
@@ -13,13 +14,15 @@ import SiteHeader from '@ecl/ec-react-component-site-header';
...
   heroBanner,
+  languageList,
   mediaContainer,
   footer,
   card,
   blockquote,
 }) => (
   <Fragment>
     <SiteHeader {...siteHeader} />
+    <LanguageListOverlay {...languageList} hidden="true" />
```
yhuard, line 25: (open question) if the content is there, why would we hide it from screen readers?

yhuard (changes requested):
thanks for this PR!
overall it works well, I have one question about the implementation and the use of the `hidden` attribute + we might want to add a focus trap once the overlay is open 😉

src/systems/ec/implementations/react/packages/campaign-page/src/CampaignPage.jsx
emeryro: it is hidden as long as the language switcher button is not "pressed".
If we don't hide it by defaut, the screen reader will read it on every page before the content

yhuard: "the screen reader will read it on every page before the content" -> that's why there's a skip link... oh my... why don't have the "skip link" component yet! 😱

Anyway, I don't know which solution is the best, whether to hide the language list from screen reader or not. On one hand, if it's not hidden, you can navigate through the list of languages without having to explicitly click on the button (better discoverability). On the other hand, it can "pollute" the user experience by reading a bunch of links that the user might not be interest in. Shouldn't we ask the UX team what's the best solution?

emeryro: We can ask the UX team.
We just don't have to wait for the answer to merge this PR ;)

yhuard: Wouldn't it be considered as a breaking change?

emeryro: It's a template, not a component.
I don't think we could have breaking changes for templates (they are just here to help the implementation)

yhuard: This PR also touches the vanilla component, it's not only about the template... If tomorrow we decide to not use the `hidden` attribute but to use a `visually hidden` class instead, it might not be a BC indeed... As long as the `hidden` attribute still works... Do we have utilities for `hidden` and `visually hidden`?  That would mostly solve our issue here :)

yhuard (approved):

PR #1040, opened 2019-03-19
fix(user-menu): fix mobile display - INNO-1400

yhuard:
Note for implementers: make sure to add the new `.ecl-site-header__user-menu` class to the user menu if you have one.

Note for testers: you can only check this PR locally. For example, you can uncomment the user menu in `src/systems/ec/ec-component/ec-component-site-header/ec-component-site-header.config.js` and then test the site header ;)

### DIFF
--- a/src/generic/generic-component-site-header/generic-component-site-header.scss
+++ b/src/generic/generic-component-site-header/generic-component-site-header.scss
@@ -42,4 +42,18 @@
   .ecl-site-header__search {
     flex-grow: 1;
   }
+
+  .ecl-site-header__user-menu {
+    display: flex;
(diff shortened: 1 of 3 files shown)


HAL-Patch-INNO:

Built with commit 5b632a51dad25bd327eba5bb5f6f3af172095d27

https://deploy-preview-1040--europa-component-library.netlify.com

emeryro (commented):
Display isn't really pretty, but as a fix it does its job.
Maybe we should switch to full width for site switcher one breakpoint ahead. With the user menu it looks somehow misaligned.
image

emeryro (approved):

PR #1063, opened 2019-04-02
fix(gallery): prevent body scroll - INNO-1428

emeryro:
# PR description

Prevent page body from scrolling when gallery overlay is displayed

To test: 
- reduce window width to have a vertical scrolling on the gallery
- click on a gallery item to open the overlay
- no scrollbar displayed, scrolling with the mouse does not affect the display
- close the overlay; body hasn't scrolled (last selected gallery item is focused)

### DIFF
--- a/src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
+++ b/src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
@@ -258,6 +258,10 @@ export class Gallery {
 
     // Focus item
     this.selectedItem.focus();
+
+    // Enable scroll on body
+    document.body.classList.remove('ecl-u-disablescroll');
+    document.body.scroll = 'yes';
   }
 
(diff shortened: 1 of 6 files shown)


HAL-Patch-INNO:

Built with commit 84a100d8ab47dffe695158689fd2d5ec62bcac86

https://deploy-preview-1063--europa-component-library.netlify.com

yhuard (changes requested):
- why `document.body.scroll`?
- can we prevent the scroll jump? (adding/removing the disablescroll class at the right time)

emeryro:
- `document.body.scroll` isn't mandatory.. I can remove it
- what scroll jump? I don't have any...

yhuard (changes requested):
one more thing: if the modal is longer than the viewport height, we can't scroll in it 

maybe a `overflow: auto;` on `.ecl-gallery__overlay[open]` could solve it

yhuard (approved):

PR #1068, opened 2019-04-04
feat(print): add print styles - INNO-1432

emeryro:
# PR description

PoC for print CSS.
It contains:
- print css for accordion and sub components
- print css for standard page and sub components

TODO
- [x] Fix display of website (css not loaded correctly)
- [ ] Update utilities for print
- [ ] Handle media container component (thumbnail?)

PR #1071, opened 2019-04-08
docs(print): how to handle print - INNO-1432

emeryro:
# PR description

RFC for print

PoC: https://github.com/ec-europa/europa-component-library/pull/1068


HAL-Patch-INNO:

Built with commit dddc3b1ede9dcf05da6fbcbd149688f141e4533c

https://deploy-preview-1071--europa-component-library.netlify.com

degliwe:
1. Do we want 1 stylesheet mixing both screen and print styling or two separated stylesheet? (I suggest two)
2. I suggest that pages should print well mostly based on markup (no css), then we will create css rules to enhance and optimize printing, but page without any style should print well. (ex: accordion should print as a description list)
3. print css should be minimal and mostly generic rules (ex: font size) we could have rules set specifically for some components but they should remain the exception.

emeryro:
1. Question added to the document
2. For some components we can rely on the markup, but we can't garantee that it is the case for all components (accordion isn't a description list anymore for instance)
3. To be discussed, but for me print css should be more intrusive in order to serve its purpose (removing some background or whole components, displaying certain part of components of not, ...)

yhuard:
1. I would use 2 separate stylesheets: 1 for `screen`, 1 for `print`. What about `speech`? If we have styles that are common to different media types, then it would make sense to mix them (`screen,print` or `all`). In terms of performance, the bigger the print styles become, the better it is to separate them from the screen styles.

2. It would be nice to check the document without CSS at all indeed, but we already know there are components we don't want to display in print mode (search, skip to content, etc.). "Components should print well without CSS" is a nice to have, but it's not essential.

3. As I said in 1., I'd go for the basic approach, there's no CSS by default except the minimalist one provided by the print stylesheet.

emeryro:
Just for fun, here is the accordion without CSS
image

kalinchernev:
My preference would go into the approach which would make it easy to maintain the code.

- File structure > Option 2: Two separated css
- CSS structure > Option 3: Separate all styles

I haven't checked current organization and haven't tried the options with code to verify if it's actually going to worth the separation, though from previous experience with other similar topics, as long as rules and files are separate, one can focus more easily and remove unnecessary code more easily. (again coming to maintenance)

Of course, in terms of resulting assets, it'd be best to keep things in common files in order to optimize the network requests.

yhuard:
> Of course, in terms of resulting assets, it'd be best to keep things in common files in order to optimize the network requests.

I would care about network requests if we were to split the stylesheets in many small pieces (let's say more than 10 stylesheets). Here, we are talking about splitting it into 2 or 3. If those 2 or 3 stylesheets are served through a CDN, I don't really care about the network requests. Furthermore, since the stylesheets are downloaded independently and since only the ones fulfilling the media query are considered high priority (render blocking), it really doesn't matter.

Nonetheless, we could define a threshold starting from which we decide to split the stylesheets. If the print stylesheet is under a certain limit (4 kB? 8 kB?), it's usually best to serve 1 stylesheet (screen + print) if we consider that `screen` is the default media type. Above this threshold, the downside of making 2 requests is compensated by the fact that one of the stylesheets (let's say the "print" one) is given a lower priority compared to the other, and thus made non-blocking.

emeryro:
**File structure:**
It seems that _Option 2: Two separated css_ is the way to go for most of us. So I'd say that this point is clear enough

**CSS structure:**
I added an _Option 4: Start without CSS_ after your feedback

yhuard:
In `CSS structure`, what's the difference between options 1 and 4? Is it just that option 4 has a global reset for print, while 1 hasn't?

planctus:
For me the main difficulty regarding the definition of the rules for printing is whether they are part of the component definition or outside of it. 
Most of times the issue is that you would have something like an hidden-print selector to extend, directly in the component, but generic rules for printing (removing backgrounds, changing font sizes, spacing)  that are not specific to it, i'd like to avoid this kind of mixed situation.

We might have two "main" css, one for screen and one for print, but also the second would be the result of the composition of a "base" css, with the specific ones coming from the component, like for the default one.
So, the component might come with:
component-screen.scss
component-print.scss

We might then compile all the print styles together (i would imagine the print css to have its own mixins, in case we need them) in the print.css keeping the "default" style untouched.

emeryro:
@yhuard indeed, the difference is mainly the reset css. And so the fact that styling is done on default html markup instead of component classes.
In option 1 print rules are always set at component level, in option 4 it is the case only if global reset rules are not enough.

yhuard:
I'm in favor of:
- separating the CSS definitions of print and screen (2 stylesheets)
- setting the styles on a per-component basis (no reset)

emeryro:
I'm in favor of:
- File structure: two separated css (Option 2)
- CSS structure: no global reset, base print css rules on component classes (Option 1)

degliwe:
We all agree on this.

```html
<head>
--
<link rel="stylesheet" type="text/css" href="ecl.css"  media="screen">
<link rel="stylesheet" type="text/css" href="ecl-print.css" media="print">
</head>
```

- We create print styling for the smallest components (links, paragraph...)
- We create only extra print styling for complex components when needed (the goal being to keep the amount of CSS needed as small as possible)

yhuard:
Also agreed: creation of specific design tokens for print (spacing, typography mostly)

docs/conventions/print.md
```diff
@@ -1,8 +1,42 @@
 # How to handle print css rules in ECL
```
yhuard: can you use this template please? https://github.com/ec-europa/europa-component-library/pull/1075/files#diff-6390ccae540cbffdb4a9337ef2f2b8cd :)

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
+# How to handle print css rules in ECL
+
+## Adopted solution
+
+- Print CSS and screen CSS are kept into separated files
+    ```
```
yhuard:
```suggestion
    ```html
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+      ...
+      <link rel="stylesheet" type="text/css" href="ecl-<preset>.css"  media="screen">
+      <link rel="stylesheet" type="text/css" href="ecl-<preset>-print.css" media="print">
+    </head>
+    ```
+
+- No global reset CSS for print
+
+- Print CSS is applied per component, using component's classes. Only needed rules are reported, to keep print CSS as small as possible.
+    ```
```
yhuard:
```suggestion
    ```css
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+    (screen css)
+    .my-component {
+      background-color: <yellow>;
+      color: <grey>;
+      display: flex;
+      font: <font-m>;
+      width: 100%;
+    }
+    ```
+    ```
```
yhuard:
```suggestion
    ```css
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+Below are the discussed options, for the record
+
+
+## File structure
+
+Mainly anwser this question: should we group all CSS in one file or split it in several files?
+
+### Option 1: One CSS for everything
+
+It means that screen and print CSS rules would be in a single SCSS file.
```
yhuard: not necessarily, we could have 2 SCSS files and still get one CSS file in the end :°

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+Print CSS rules would be put in a `@media print` section.
+All SCSS files are merged in a `ecl-<preset>.css` file.
+
+Example:
+
+```
```
yhuard:
```suggestion
```css
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+
+```
+<link rel="stylesheet" type="text/css" href="ecl-<preset>.css">
+```
+
+### Option 2: Two separated css
+
+It means adding an extra CSS file for print only
+Each component would have 2 SCSS files (one for print one for screen)
```
yhuard: not necessarily, we could have 1 SCSS file with 2 mixins... :°

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+Screen files are merged into a `ecl-<preset>.css` file; print files are merged into `ecl-<preset>-print.css`.
+
+Example:
+
+```
```
yhuard:
```suggestion
```css
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+(my-component.scss)
+.my-component {
+  <component-css-rules>
+}
+```
+
+```
```
yhuard:
```suggestion
```css
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+(my-component-print.scss)
+.my-component {
+  <component-print-css-rules>
+}
+```
+
+```
```
yhuard:
```suggestion
```html
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+}
+
+@media print {
+  .my-component {
+    <component-print-css-rules>
+  }
+}
+```
+
+```
```
yhuard:
```suggestion
```html
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+
+Independently to file structure, the goal here is to discuss how we write CSS rules for print.
+
+### Option 1: Separate screen and print
+
+It means having all css rules related to screen display grouped together, and all print rules grouped together somewhere else.
+
+Example:
+
+```
```
yhuard:
```suggestion
```css
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+
+- Have to specify all css rules in print (duplication)
+
+### Option 2: Define common rules and override for print
+
+It means having css rules without specific media target (the way it is now), and override what we need for print.
+
+Example:
+
+```
```
yhuard:
```suggestion
```css
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+
+- Have to override css rules
+
+### Option 3: Separate all styles
+
+It means having first common rules for screen and print, then specific rules.
+
+Example:
+
+```
```
yhuard:
```suggestion
```css
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+- Need to update existing css for all components
+
+### Option 4: Start without CSS
+
+It means that component's CSS will be loaded for screen only. Print would be based on default dipslay (without CSS), with some custom rules if needed.
+It also implies to have global CSS rules for print, based on html markup.
+
+Example:
+
+```
```
yhuard:
```suggestion
```css
```

docs/conventions/print.md
```diff
@@ -0,0 +1,232 @@
...
+}
+```
+
+Pros:
+
+- Clear separation of styles
+
+Cons:
+
+- Have to specify all css rules in print (duplication)
```
yhuard: it's only duplicated if we want the same result... most of the time, I guess we won't want the same output :)

yhuard (changes requested):

yhuard:
Now I have a question/suggestion that might need some testing: what if we have both

```html
<link rel="stylesheet" type="text/css" href="ecl-<preset>.css"  media="screen">
<link rel="stylesheet" type="text/css" href="ecl-<preset>-print.css" media="print">
```

and:

```css
/* ecl-<preset>.css */
@media screen {
  ...
}
```

```css
/* ecl-<preset>-print.css */
@media print {
  ...
}
```
?

i.e. double security. 

It's possible in CSS 3 to nest media queries, cf https://www.w3.org/TR/css3-conditional/#processing but it seems (surprise!) that IE [doesn't support it](https://caniuse.com/#feat=css-mediaqueries), which mean we'd need to flatten the queries...

emeryro:
Document has been updated.
Concerning the use of `@media` inside css already called for a specific media, I don't really see the value...

yhuard:
> Concerning the use of @media inside css already called for a specific media, I don't really see the value...

it's just about making sure the styles are not applied to the wrong context (e.g. if someone forgets to add the `media` attribute on the `link`)

emeryro:
It could be an improvement, but this is mainly for the screen css (that's where the media could be forgotten). For the print css it is quite obvious.
So I'd say we could create a ticket for that, but it won't block the print css tasks

yhuard (approved):

PR #1075, opened 2019-04-09
docs: define which browsers we support - INNO-1438

yhuard:
The goal of this RFC is to document:
- which browsers we "officially" support
- the browserslist config we are going to use from now on


HAL-Patch-INNO:

Built with commit 9a73eebe801a07afa1ac0b70a0926180b7c1ea24

https://deploy-preview-1075--europa-component-library.netlify.com

yhuard:
https://twitter.com/csswizardry/status/1093133185146388480 :)

yhuard:
> The EU and EC web presences must offer an optimal browsing experience on every browser which has been used by at least 2% of its users in the last 12 months.
> For 2019, the minimum browser versions that must be supported are:
> 
> - Chrome 55.0
> - Chrome Mobile 55.0
> - Firefox 45.0
> - Internet Explorer 11.0
> - Microsoft Edge 14
> - Safari 10.0
> - Opera 52.0
> - Samsung Browser 4.0
>
> The content must also be accessible to those visitors not using mainstream browsers. This means applying a progressive enhancement approach, where features not supported by the latest browsers technologies can still function without adversely affecting the user experience.
> Assistive technology (screen readers or any other alternative web browsing systems) and open source operating systems (e.g. Linux, FreeBSD) should be tested even if they account for less than 2% of users.

https://wikis.ec.europa.eu/display/IPG/05.+Browser+support

This gives us a good starting point :)

emeryro:
I guess the most flexible option is 2.2 (dynamic based on custom stats), but I have no idea about the amount of work it would require.

The easiest one is to take the list given by IPG and apply is as is. So no question, we just use the "official" list.

planctus:
I also see that if we have a list coming from IPG we could simply stick to that, even though there is this generic remark regarding "assistive technologies and open source operating systems" that makes it undetermined enough not be a list anymore.

yhuard:
So far, we all agree that the bare minimum should be the list provided by the IPG. Translated into a browserslist config:

```
Chrome >= 55
ChromeAndroid >= 55
Firefox >= 45
IE >= 11
Edge >= 14
Safari >= 10
Opera >= 52
Samsung >= 4
```

Missing: `iOS` (Safari iOS), `FirefoxAndroid`, `Android` (Android WebView), `OperaMini`, `OperaMobile`...

To do: check the stats to find out which browsers we should also support

degliwe:
I would support Alternative 2.1: based on caniuse stats
I agree that 2.2 is the "tailor made" option, but I don't necessarily think this is worth the effort or the way to go.
European citizens and people living in Europe are the one ECL is supposed to support, not only the people currently  consuming EC sites (civil servant, ngo's...)
We want to broaden the audience and the EC web presence to be inline with the reality "on the ground" and not the "bureaucratic reality"

yhuard:
The [latest version of the IPG](https://wikis.ec.europa.eu/display/WEBGUIDE/04.+Browser+support) has a new set of minimum browser versions that must be supported:

>Chrome 62.0
>Chrome Mobile 55.0
>Firefox 50.0
>Internet Explorer 11.0
>Microsoft Edge 16
>Safari 10.0
>Opera 52.0 

The first rule is specific to every website:

> All European Commission websites must offer an optimal browsing experience (i. e. all functional and visual bugs must be fixed) on every browser which has been used by at least 2% of its total users in the last 12 months.

We can't offer a browserslist config that covers the needs of every website. Instead, if we target `> 2%` (global) or `> 2% in alt-eu` (Europe only), it should cover most of the cases. We could even lower the threshold (say `1%` or `O.5%` ).

Note that browserslist's `defaults` is: `> 0.5%, last 2 versions, Firefox ESR, not dead`. We could ignore `last 2 versions` which adds a lot of noise (IE Mobile, Baidu browser, KaiOS Browser...), keep the rest, localise it (`in alt-eu`), and add the minimum browser versions list from the IPG:

browserslist (or `.browserslistrc`)
```
> 0.5% in alt-eu
Chrome >= 62
ChromeAndroid >= 55
Firefox >= 50
Firefox ESR
IE >= 11
Edge >= 16
Safari >= 10
Opera >= 52
not dead
```

You can see the results here: https://browserl.ist/?q=%3E+0.5%25+in+alt-eu%2C+Chrome+%3E%3D+52%2C+ChromeAndroid+%3E%3D+55%2C+Firefox+%3E%3D+50%2C+Firefox+ESR%2C+IE+%3E%3D+11%2C+Edge+%3E%3D+16%2C+Safari+%3E%3D+10%2C+Opera+%3E%3D+52%2C+not+dead

Compared to our current config: https://browserl.ist/?q=%3E+1%25%2C+last+2+versions%2C+not+ie+10%2C+not+ie_mob+10%2C+not+safari+%3C+10

emeryro:
Latest proposal looks good to me

degliwe (approved):

PR #814, opened 2018-09-18
feat(storyshots): integrate storyshots - INNO-1226

yhuard:
# PR description

Known issues: sauce labs' safari can't access localhost, even when SC is running.

PR #1080, opened 2019-04-11
feat(storyshots): run some tests

yhuard:
Will eventually be merged in https://github.com/ec-europa/europa-component-library/pull/814 

New features:
- local server: automatic port detection
- local tests: automatically launch Sauce Connect
- serve images

We still can't parallelise tests locally because it would launch multiple instances of Sauce Connect, which is not possible. Ideally, for all the browsers, all the systems, we should only start 1 instance of Sauce Connect. Then, for each system, we should start 1 server. The ideal parallel tests setup should be 1 Sauce Connect, 2 servers (EC, EU), nothing more. To be done later


HAL-Patch-INNO:

Built with commit 3df1d7d3ead9704dd2dc4c7ba4a6aa4a68a1a3cf

https://deploy-preview-1080--europa-component-library.netlify.com

PR #1083, opened 2019-04-15
fix(buttons): remove "button" appearance - INNO-1280

yhuard:
This PR can be manually tested. If you change the "file download" button to a link with the same classes and attributes, it should display well now.

PR #1085, opened 2019-04-15
fix(buttons): remove appearance - INNO-1280

yhuard:
Same as https://github.com/ec-europa/europa-component-library/pull/1083, but for ECL 2


HAL-Patch-INNO:

Built with commit c38d10e91b6892dfea235cd2ee5e1d5d5ecada09

https://deploy-preview-1085--europa-component-library.netlify.com

emeryro (changes requested):
If I do the same test as on V1 (change markup from `<button>` to `<a>`), display is broken.
image

Without `appearance: none` it is not correct, but at least it's not broken...

yhuard:
@emeryro good catch, it should be fixed now ;)

emeryro (approved):

PR #1102, opened 2019-04-30
feat(message): add print css - INNO-1467

emeryro:
# PR description

Add print css for Message

### DIFF
--- a/src/systems/ec/implementations/vanilla/packages/ec-component-message/ec-component-message-print.scss
+++ b/src/systems/ec/implementations/vanilla/packages/ec-component-message/ec-component-message-print.scss
@@ -6,7 +6,7 @@
 // Import base
 @import '@ecl/ec-base/ec-base';
 
-@mixin ecl-message-print() {
+@mixin ecl-message-print($boder-width: 1.5pt) {
   .ecl-message {
     margin: 0;
     padding: $ecl-spacing-print-m;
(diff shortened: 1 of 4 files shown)


HAL-Patch-INNO:

Built with commit 0070837cec9a8cf62086c5bdd09e3e7467c72940

https://deploy-preview-1102--europa-component-library.netlify.com

yhuard (commented):
my only concern is about the use of `px` in print... shouldn't we use `pt`?

What are the best practices regarding units (relative vs absolute) in print?

Rephrased: do we want to give to the user the possibility to scale the font/spacing (using relative units) in print or do we enforce the scale?

https://gist.github.com/basham/2175a16ab7c60ce8e001#css-units

src/systems/ec/implementations/vanilla/packages/ec-component-message/ec-component-message-print.scss
```diff
@@ -0,0 +1,55 @@
+/*
+ * Message
+ * @define message
+ */
+
+// Import base
+@import '@ecl/ec-base/ec-base';
+
+@mixin ecl-message-print($boder-width: 1.5pt) {
```
yhuard: I'd rather use `mm` or `cm` here

- typography: `pt`
- spacing and other measures: `mm`, `cm`

wouldn't it be better this way?

yhuard (changes requested):

yhuard (approved):

PR #1108, opened 2019-05-02
chore: update dependencies - INNO-1474

yhuard:
This PR also: 
- fixes a design change in MDX where even HTML elements now pass in the custom components transform (MDXProvider)
- Storybook: add "CSS Resources" panel with the possibility to inject a fake "reset" (global rules) -> our components must not change when the reset is activated - INNO-1445


HAL-Patch-INNO:

Built with commit e8cdb9bcbdb5452b3d82541e32bf6382d482cc3c

https://deploy-preview-1108--europa-component-library.netlify.com

emeryro (dismissed):
- When we open storybook for the first time on a browser, the display of the sidebar is broken (too small)
image
After resizing it, it is fine (I guess it is saved somewhere), but that's not very obvious that it can be resized

- CSS is not applied on IE
image

degliwe (approved):

PR #1113, opened 2019-05-08
feat(list): add standard list - INNO-1214

emeryro:
# PR description

Add [list component](https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Lists+-+Design+specifications).

It includes:
- ordered list
- unordered list
- description list

Print version is also provided.


HAL-Patch-INNO:

Built with commit 50075eefde633ba9d33a5a16d8d4aa208dc50d75

https://deploy-preview-1113--europa-component-library.netlify.com

degliwe (changes requested):
general:
- we have 3 types of list, each should be a different component.
- for the sake of consistency, the 3 stories in storybook should be "ordered, unordered, description"

ordered list:
- doesn't exists in blank
- doesn't exists with dividers
- one example showcasing a list with some nesting is enough (don't need more stories for this one)
- knobs are unclear (bullet is a boolean, ordered is another, user will uncheck bullet if they want ordered, not clear to use (solved if they are 2 components))

unordered list:
- the divider is missing on the last element (on the specs it's there)
- unordered without bullet example should be illustrated only with links (an unordered list without bullet with text looks very weird) this example should be named "link list" this is the only moment the usage guidelines will suggest the use of unordered with bullets.

src/systems/ec/implementations/react/packages/list/src/OrderedListWithData.jsx
```diff
@@ -0,0 +1,38 @@
...
+import OrderedList from './OrderedList';
+import OrderedListItem from './OrderedListItem';
+
+const OrderedListWithData = ({ items, className, ...props }) => {
+  return (
+    <OrderedList {...props} className={className}>
+      {items.map(item => {
+        if (item.nested) {
+          return (
+            <Fragment key={item.label}>
```
yhuard: the Fragment is not needed here I guess

src/systems/ec/implementations/react/packages/list/src/UnorderedListWithData.jsx
```diff
@@ -0,0 +1,42 @@
...
+import UnorderedList from './UnorderedList';
+import UnorderedListItem from './UnorderedListItem';
+
+const UnorderedListWithData = ({ variant, items, className, ...props }) => {
+  return (
+    <UnorderedList {...props} className={className}>
+      {items.map(item => {
+        if (item.nested) {
+          return (
+            <Fragment key={item.label}>
```
yhuard: same remark, `Fragment` is probably not needed, `key` can be applied to `UnorderedListItem` directly

src/systems/ec/implementations/vanilla/packages/ec-component-list/ec-component-list.scss
```diff
@@ -0,0 +1,127 @@
...
+@mixin ecl-ordered-list(
+  $text-color: $ecl-color-grey-100,
+  $divider-color: $ecl-color-grey-15
+) {
+  .ecl-ordered-list {
+    color: $text-color;
+    font: $ecl-font-prolonged-m;
+    margin: 0;
+    // Approximately centered bullet
+    padding-left: calc((#{$ecl-spacing-s} * 2) + 0.3em);
```
yhuard, line 20: you expected me to leave a comment here... didn't you? "centered" compared to what?
I see that the examples from the specs are too simple, what should the list look like when the counter reaches 100 for example?

src/systems/ec/implementations/vanilla/packages/ec-component-list/ec-component-list.scss
```diff
@@ -0,0 +1,127 @@
...
+  .ecl-unordered-list {
+    color: $text-color;
+    font: $ecl-font-prolonged-m;
+    margin: 0;
+    // Approximately centered bullet
+    padding-left: calc((#{$ecl-spacing-s} * 2) + 0.3em);
+  }
+
+  .ecl-unordered-list .ecl-unordered-list {
+    margin-top: $ecl-spacing-xs;
```
yhuard, line 52: wondering... if the same top spacing is used between list items AND within an item, between the label and the nested list, how do you know the nested list is not another list item (not related to the label)?

src/systems/ec/implementations/react/packages/list/src/OrderedListWithData.jsx
yhuard: I think it could be compacted like:
```
<OrderedListItem key={item.label}>
  {item.label}
  {item.nested && <OrderedListWithData items={item.nested} />}
</OrderedListItem>
```

end result:
```
const OrderedListWithData = ({ items, className, ...props }) => {
  return (
    <OrderedList {...props} className={className}>
      {items.map(item => (
        <OrderedListItem key={item.label}>
          {item.label}
          {item.nested && <OrderedListWithData items={item.nested} />}
        </OrderedListItem>
      ))}
    </OrderedList>
  );
};
```
would it work too?

yhuard (changes requested):
a few improvements possible here and there but overall, it looks good ;)

src/systems/ec/implementations/vanilla/packages/ec-component-list/ec-component-list.scss
emeryro: "centered" as in "the same space before and after the bullet" (spacing-s).
The approximate comes from the `0.3em`, which is the size of the bullet (more or less).
But I agree that it is relevent mostly for unordered list.

src/systems/ec/implementations/vanilla/packages/ec-component-list/ec-component-list.scss
emeryro: well, with the indentation

yhuard (approved):

PR #1129, opened 2019-05-16
feat(stack): add stacks - INNO-1492

yhuard:


HAL-Patch-INNO:

Built with commit 0f65c897d24245cec198d2a20ea84cea7746b4cf

https://deploy-preview-1129--europa-component-library.netlify.com

src/systems/ec/implementations/react/utilities/stacks/stories/Index.jsx
```diff
@@ -0,0 +1,189 @@
...
+      'Container'
+    );
+
+    const alignItems = select(
+      'Align items',
+      {
+        Start: 'start',
+        End: 'end',
+        Center: 'center',
+        Baseline: 'baseline',
```
emeryro, line 64: `align-items: baseline` displays the items centered on the demo, where it should be at baseline.

emeryro (changes requested):
I wonder who is the target of this documentation...
This is mostly an "how to use CSS flexbox" description. Is it the role of ECL to teach such things?
Can't we have just a few generic stack examples (the most used ones), and maybe a link to this full documentation if needed?
I fear that it would be too long and complex to read.

src/systems/ec/implementations/react/utilities/stacks/stories/Index.jsx
yhuard: where's the baseline in the demo? at the center :)

one easy way to verify it is to remove `align-items: center;` from the box "11"

image
 
now the difference is obvious 😄

yhuard:
> This is mostly an "how to use CSS flexbox" description.

In the context of the ECL, i.e. with the utilities provided by the ECL, not everything that flexbox allows to do is possible. On the other hand, most (if not all) ECL utilities are described here. You know what you can do with the ECL utilities (and you learn about flexbox :) ).

> the most used ones

what are the most used stacks? most used by whom?

when we'll start making "compositions" (under "Templates"), we'll probably create some complex layouts, we'll need the full potential of these utilities

emeryro (approved):

PR #1134, opened 2019-05-23
feat(navigation-lists): create navigation lists composition - INNO-1490

yhuard:


HAL-Patch-INNO:

Built with commit a9b828cee0876d72998dfb242fcc79d5f2a4a212

https://deploy-preview-1134--europa-component-library.netlify.com

src/systems/ec/implementations/react/templates/compositions/navigation-lists/examples/Default.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import Link from '@ecl/ec-react-component-link';
+import { UnorderedList, UnorderedListItem } from '@ecl/ec-react-component-list';
+
+export default () => (
+  <div>
```
emeryro, line 6: I'm not a big fan of this "naked" div here... Though I don't see any other semanic tag that may fit.

src/systems/ec/implementations/react/templates/compositions/navigation-lists/examples/Default.jsx
```diff
@@ -0,0 +1,43 @@
...
+    <h2 className="ecl-u-ma-none ecl-u-type-prolonged-xl ecl-u-type-bold">
+      <Link
+        href="#example"
```
emeryro: This kind of target triggers accessibility error in storybook

src/systems/ec/implementations/react/templates/compositions/navigation-lists/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "description": "ECL EC React Navigation Lists",
+  "main": "index.js",
+  "module": "index.js",
+  "dependencies": {
+    "@ecl/ec-react-component-link": "^2.4.0",
+    "@ecl/ec-react-component-list": "^2.4.0"
+  },
+  "devDependencies": {
+    "@ecl/ec-utility-spacing": "^2.4.0",
+    "@ecl/ec-utility-typography": "^2.4.0"
```
emeryro, line 17: it also uses color utilities

emeryro (changes requested):
Overall, it looks good. 

A few questions:
- should we add an example with a separator?
- should we add an example with multiple navigation lists (wrapped in a `<nav>`)?

src/systems/ec/implementations/react/templates/compositions/navigation-lists/examples/Default.jsx
yhuard: Indeed, it's ugly. It will make more sense when used in a composition of composition, when you need to apply classes on the container (the `div` here). For now, I don't see any better solution...

src/systems/ec/implementations/react/templates/compositions/navigation-lists/examples/Default.jsx
yhuard: True, I will fix it 👍

src/systems/ec/implementations/react/templates/compositions/navigation-lists/package.json
yhuard: Good catch 👍

yhuard:
> should we add an example with a separator?

Nope. After discussion with the designers, "Secondary links" are not part of the composition anymore.

> should we add an example with multiple navigation lists (wrapped in a `<nav>`)?

Not here. Unless the designers specifically require it. But the last time we spoke, the page was good.

emeryro (approved):

PR #1176, opened 2019-06-26
feat(links): add CTA variant - INNO-1580

yhuard:
This PR adds a new variant to the link component: "CTA" (call-to-action). It looks like the CTA button, except that the link is underlined. Both hero and page banners use this new variant.

### DIFF
--- a/src/systems/ec/implementations/vanilla/packages/ec-component-link/ec-component-link.scss
+++ b/src/systems/ec/implementations/vanilla/packages/ec-component-link/ec-component-link.scss
@@ -69,6 +69,34 @@
     }
     /* stylelint-enable */
   }
+
+  .ecl-link--cta {
+    background-color: $ecl-color-yellow-100;
+    box-sizing: border-box;
+    color: $ecl-color-black-100;
+    display: inline-block;
+    font: $ecl-font-m;
+    font-weight: $ecl-font-weight-bold;
+    padding: $ecl-spacing-s $ecl-spacing-m;
+    text-decoration: underline;
+
+    /* stylelint-disable */
+    &.ecl-link--icon {
+      display: inline-flex;
+      text-decoration: none;
+    }
+
+    &:hover,
+    &:active {
+      border: 2px solid $ecl-color-black-100;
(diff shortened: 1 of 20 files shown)


HAL-Patch-INNO:

Built with commit 8ce2d4b006ca272aafe8843aedc60e016b369582

https://deploy-preview-1176--europa-component-library.netlify.com

src/systems/ec/implementations/vanilla/packages/ec-component-link/ec-component-link.scss
```diff
@@ -69,6 +69,34 @@
     }
     /* stylelint-enable */
   }
+
+  .ecl-link--cta {
```
emeryro, line 73: the button uses class `ecl-button--call`. We should keep some consistency (`ecl-link--call`)

src/systems/eu/implementations/vanilla/packages/eu-component-link/eu-component-link.scss
```diff
@@ -69,6 +69,34 @@
...
+
+  .ecl-link--cta {
+    background-color: $ecl-color-yellow-100;
+    box-sizing: border-box;
+    color: $ecl-color-black-100;
+    display: inline-block;
+    font: $ecl-font-m;
+    font-weight: $ecl-font-weight-bold;
+    padding: $ecl-spacing-s $ecl-spacing-m;
+    text-decoration: underline;
```
emeryro: this is already defined in ecl-link, no need to duplicate (same for hover and outline in focus)

emeryro (changes requested):
almost good!

src/systems/ec/implementations/vanilla/packages/ec-component-link/ec-component-link.scss
yhuard: I disagree :) 
1. a link is not a button, there's no reason to maintain consistency between them
2. `cta` is much clearer than `call`

emeryro: ok, as agreed, we will keep it with `cta` here and update the buttons to use the same name

emeryro (approved):

PR #1191, opened 2019-07-08
chore(website): enhance performance - INNO-1617

yhuard:
Route-based code splitting, meta data (frontmatter) preloaded, many small tweaks here and there in order to reduce the bundled size...

## Bundles

Was:

image

Now:

image

The next step will be to reduce the number of request by merging the small (<4kB) chunks. To be done later.

## Performance score

Was:

image

Now:

image

500ms to 800ms saved! 😎 

## Data (first load)

Was:

image

Now:

image

## Cache

Before, when 1 page was changed, the whole EC or EU JS file had to be downloaded again. Now, when a user visits a page the has been changed, only the JS for this page will be download (a few kB, comparent to hundreds of kB before). The browser's cache is thus now more efficiently used :)

## Other possible enhancements

- Preload/prefetch https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules
- better `optimization.splitChunks` options (https://webpack.js.org/guides/caching/, https://webpack.js.org/plugins/split-chunks-plugin/)
- find use of https://github.com/numical/script-ext-html-webpack-plugin

## UPDATE 12/08/2019

now, v2-dev is:

image

with current changes: 

image

Back to 1 big chunk for all the pages (1 for EC pages, 1 for EU pages).

All it takes to split it to 1 chunk per page is to change `src/website/src/routes/Ec.jsx`: `/* webpackMode: "lazy-once" */` to `/* webpackMode: "lazy" */`.


HAL-Patch-INNO:

Built with commit d8acfd7e5a273ac4b91202b619d9c79a6bdaf1a6

https://deploy-preview-1191--europa-component-library.netlify.com

emeryro:
I have mixed feelings here.
On one hand, the initial loading is indeed faster (800ms according to your stats), on the other hand there is now a loading time for each other pages (duration may vary, but I'd say 200 to 800ms). Of course this is only for the first visit, when the pages are in cache, there is no real differences anymore.

I get the global idea: manage each page separately and not reload the whole site each time, which is fine by me.
But with the current implementation it "feels" slower when browsing the site (at least with my setup).

I guess that proposed enhancement may improve things (preloading of pages for instance).

Currently, the time we gain at startup is somehow "lost" when we browse the site. This may change in the future if the site gain weight.

In the end I'd suggest to wait before merging this PR, and try to apply one enhancement or another

emeryro (approved):
Navigation on the website is definitely faster now.
I didn't see any drawback. Good job!

PR #1197, opened 2019-07-11
refactor(list): split component - INNO-1564

emeryro:
# PR description

Split `list` component into 3 packages
- `list-description`
- `list-ordered`
- `list-unordered`

Notes (it also apply for EU):
- package `ec-react-component-list` has been removed as it was used only internally 
- package `ec-component-list` has been put in legacy
- package `ec-specs-list` has to be kept as is, as it is used by ecl-twig. Once twig has been updated, this package can be removed
- templates using list have been updated


HAL-Patch-INNO:

Built with commit 8ebe8daf4b37901c5af0f8fc47ff51fada90c2fd

https://deploy-preview-1197--europa-component-library.netlify.com

src/systems/ec/implementations/react/components/list-description/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "private": true,
+  "name": "@ecl/ec-react-component-list-description",
```
yhuard: may I suggest we rename it `@ecl/ec-react-component-description-list`? 👼

emeryro: I put it that way to keep all packages related to list together (when ordered alphabetically)

yhuard: in ECL Twig, we have this https://github.com/ec-europa/ecl-twig/tree/master/src/ec/packages

- ec-component-description-list
- ec-component-ordered-list
- ec-component-unordered-list

if we want to keep it consistent, we should follow the same pattern, shouldn't we?

emeryro: well, ok. I didn't remember we already did it the wrong way on twig :)

yhuard (approved):
LGTM 👍

PR #1213, opened 2019-07-16
docs(text-field): update documentation of text field - INNO-1641

yhuard:
https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Input+-+Text+Field+-+ECL+Page+-+final

### DIFF
--- a/src/website/src/pages/ec/components/forms/text-field/docs/usage.mdx
+++ b/src/website/src/pages/ec/components/forms/text-field/docs/usage.mdx
@@ -78,4 +78,4 @@ import { Paragraph, Anatomy, Link } from '@ecl/website-components';
 
 - <Link to="/ec/components/forms/radio/usage/">Radio button</Link>
 - <Link to="/ec/components/forms/select/usage/">Select</Link>
-- <Link to="/ec/components/forms/text-area/usage/">Text area</Link>
+- <Link to="/ec/components/forms/text-area/usage/">Textarea</Link>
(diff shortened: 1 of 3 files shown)


HAL-Patch-INNO:

Built with commit 8e146ca32751e45f347a72ee9c2aa0e67bef2fb1

https://deploy-preview-1213--europa-component-library.netlify.com

planctus:
@yhuard , i see a few differences from the wiki page:

1. this is actually reproducing what we have in the page, we have two "links" to the text-area component but their label is once "text area" and then "Textarea", might look irrelevant but it's worth defining a vocabulary that is consistently used while writing guidelines for others, i think.

2. "when not to use" section, in the wiki we have "Dropdown select", we wrote "Select", not sure if it's intentional, but it is consistent with the "related component" link where in the wiki page we have "Dropdowns", did we agree with the folks that we would simply call it "Select"? :)

3. List of related components apparently is not matching, although it is unclear to me what "Generic inputs page" might refer to, so maybe it's just that we miss "checkboxes" here, but this is on purpose i believe because we don't have the component yet.

planctus (commented):
Just these few inconsistencies with the wiki page that i put in the comment.

yhuard:
@planctus thank you for the review, it made me realise I updated the text area .md file instead of the text field's one 😱 

1. they (usability) use both "text area" and "textarea", I agree it's confusing. Feel free to put a comment on their wiki page 👼 
2. they call the "select" component a "dropdown select" or, even worse, just "dropdown". Wesley and I strongly disagree with the naming, that might explain why I've replaced "dropdown" with "select" :)
3. indeed! I don't know what "Generic inputs page" is, and we don't have the checkbox yet, this explain the differences

planctus (approved):

PR #1223, opened 2019-07-19
chore(website): enhance sidebar behaviour - INNO-1659

yhuard:


HAL-Patch-INNO:

Built with commit c9f7b16e77d0ccbeeb55f7a3f7b11c9178c9c792

https://deploy-preview-1223--europa-component-library.netlify.com

emeryro (changes requested):
it is indeed much better this way.
I only spotted an issue with the close button
image
Using `right: 0` instead of fixed left should do the trick
image

yhuard:
Updated with `right: 0` ✅

emeryro (approved):

PR #1235, opened 2019-07-31
fix(label): use inline display - INNO-1704

emeryro:
# PR description

Update label display to not use display block


HAL-Patch-INNO:

Built with commit 174edd1e8dde8d6c0ea30393ee79e4afdf990515

https://deploy-preview-1235--europa-component-library.netlify.com

yhuard (changes requested):
one more thing: `.ecl-radio__group--binary .ecl-radio` should have `display: inline-flex;` instead of `display: inline-block;` (there's a 4px difference 😱)

yhuard (changes requested):
Binary radio groug is now broken :'(

yhuard (approved):
I can't find any issues :'(

PR #1248, opened 2019-08-08
feat(events): add event details template - INNO-1719

emeryro:
# PR description

See ticket for details: https://webgate.ec.europa.eu/CITnet/jira/browse/INNO-1719


HAL-Patch-INNO:

Built with commit e45354e853f774681b734b8286774cc7e40d864d

https://deploy-preview-1248--europa-component-library.netlify.com

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
```diff
@@ -0,0 +1,815 @@
...
+                  European Year of Culture Heritage
+                </div>
+              </div>
+
+              <div className="ecl-row ecl-u-type-m ecl-u-mt-m">
+                <div className="ecl-col-12 ecl-col-md-4 ecl-u-type-bold">
+                  Website
+                </div>
+                <div className="ecl-col-12 ecl-col-md-8 ecl-u-mt-2xs">
+                  <Link label="see website" href="/example" />
```
yhuard: would be nice to use an address as a label, as on the mockup

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
```diff
@@ -0,0 +1,815 @@
...
+            <BreadcrumbItem label="Events" href="/example" />
+            <BreadcrumbItem label="Fair of European Innovators in Cultural heritage" />
+          </Breadcrumb>
+          <div className="ecl-page-header__meta-list">Conference</div>
+          <div className="ecl-u-d-flex ecl-u-flex-column ecl-u-flex-md-row">
+            <h1 className="ecl-page-header__title ecl-u-flex-grow-1">
+              Fair of European Innovators in Cultural heritage
+            </h1>
+            <div className="ecl-u-align-self-end ecl-u-mt-l ecl-u-mt-md-none">
+              <Button variant="call" label="Register" />
```
yhuard: "Register" are links, not buttons

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
```diff
@@ -0,0 +1,815 @@
...
+            </div>
+          </div>
+        </div>
+      </div>
+
+      <main className="ecl-u-pv-2xl ecl-u-pv-md-3xl">
+        <div className="ecl-container">
+          <div className="ecl-row">
+            <div className="ecl-col-12 ecl-col-md-6 ecl-u-order-md-last">
+              <UnorderedList variant="no-bullet">
```
yhuard, line 53: I don't see the value of using a list here

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
```diff
@@ -0,0 +1,815 @@
...
+                </div>
+              </div>
+
+              <div className="ecl-row ecl-u-type-m ecl-u-mt-m">
+                <div className="ecl-col-12 ecl-col-md-4 ecl-u-type-bold">
+                  Social media
+                </div>
+                <div className="ecl-col-12 ecl-col-md-8 ecl-u-mt-2xs">
+                  {/* :'( Custom social media follow */}
+                  <SocialMediaFollow
```
yhuard: can you fix `SocialMediaFollow` so it doesn't render `ecl-social-media-follow__description` when there's no description please?

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
```diff
@@ -0,0 +1,815 @@
...
+
+              <div className="ecl-row ecl-u-type-m ecl-u-mt-m">
+                <div className="ecl-col-12 ecl-col-md-4 ecl-u-type-bold">
+                  Entrance fee
+                </div>
+                <div className="ecl-col-12 ecl-col-md-8 ecl-u-mt-2xs">Free</div>
+              </div>
+            </div>
+            <div className="ecl-col-12 ecl-col-md-6">
+              <img
```
yhuard: shouldn't it be a webtools widget here?

yhuard (changes requested):
globally it's interesting :) it shows that some specs are unrealistic

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
```diff
@@ -0,0 +1,815 @@
...
+                </div>
+              </div>
+            </div>
+
+            <div className="ecl-col-12 ecl-col-md-6 ecl-u-mt-l ecl-u-mt-md-none">
+              <h3 className="ecl-u-type-heading-3 ecl-u-mt-none ecl-u-mb-m ecl-u-mb-md-l">
+                Press contact
+              </h3>
+
+              <div className="ecl-row ecl-u-type-m">
```
yhuard: you should not nest rows

degliwe:
What about using address for contact info? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address
What about hcard? http://microformats.org/wiki/hcard

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
emeryro: that's possible but we need extra utilities to set `overflow-wrap` otherwise the design is broken

yhuard: then let's break it!

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
emeryro: I don't see why we shouldn't use a list... That's a "serie" of several items related to the same topic (event information). So to me that's clearly a list

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
emeryro: actually that's an open question. Theoretically the description is mandatory. So this case should not exist.

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
emeryro: on a real implementation, yes. In a template... maybe.
but I have no idea where to find information about this widget

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
emeryro: why not? We don't really have other options to set consistent column width design

src/systems/ec/implementations/react/templates/pages/event-detail/examples/Default.jsx
emeryro: I tried to add the webtools widget, but to allows it to work we would have to send a request for domain whitelisting (and that's probably not something we want to do for a template)

PR #1252, opened 2019-08-12
fix(cards): use aria-label instead of alt - INNO-1727

yhuard:
And other fixes:
- `aria-label` and `role="img"` are used on the image `<div>`
- `level` is not propagated to `<Link>` anymore
- `<section>` is replaced by a `<div>` because there's no heading in it
  > Each section should be identified, typically by including a heading (h1-h6 element)as a child of the section element.

  — https://www.w3.org/wiki/HTML/Usage/Headings/Missing#section_element_advice
- remove `alt` on videos (not supported)

tested with https://validator.w3.org/nu/

### DIFF
--- a/src/systems/ec/implementations/react/components/card/src/Card.jsx
+++ b/src/systems/ec/implementations/react/components/card/src/Card.jsx
@@ -27,7 +27,8 @@ const Card = ({
     imageMarkup = (
       <div
         className="ecl-card__image"
-        alt={image.alt}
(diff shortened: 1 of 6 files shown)


HAL-Patch-INNO:

Built with commit 2e9b69d141f908fe39993b439aac033bd8b81b47

https://deploy-preview-1252--europa-component-library.netlify.com

emeryro (approved):

PR #1273, opened 2019-08-26
refactor(eu-footer): update EU footer - INNO-1740

yhuard:
The changes are as small as possible. In order to use the new footer, the user has to:
- add a few classes: `ecl-footer__custom-v1908"` ("August 2019"), `ecl-footer__publication-date` 
- do a small change as shown here: https://github.com/ec-europa/europa-component-library/pull/1273/files#diff-42f68ea12cee02e8134529171e1bc59fR97

see https://github.com/ec-europa/europa-component-library/pull/1273/files#diff-42f68ea12cee02e8134529171e1bc59f

These changes have a minimal impact on the old markup, only the height of the blue border will change on old markup with the new styles (no a big deal)


HAL-Patch-INNO:

Built with commit 745f3c80e1a654746218ffa5a2a2e63d209138ae

https://deploy-preview-1273--europa-component-library.netlify.com

emeryro:
I don't really get the point of this `ecl-footer__custom-v1908`. Why didn't you just update the existing css rules?
Yes, the display will be slightly different if the user get that latest version of the css, but that's the whole point of it.
That won't be a breaking change.

emeryro (approved):

PR #1294, opened 2019-09-02
fix(content-item): fix image css - INNO-1761

emeryro:
# PR description

Remove forced height on the images in content-item template, as it already uses a ratio (so width is enough)

Note: utilities will be used instead of inline style when media utilities are fully ready


HAL-Patch-INNO:

Built with commit c11a3d852171a80f95f98e510ffb6683e45c04b9

https://deploy-preview-1294--europa-component-library.netlify.com

degliwe (approved):

PR #1320, opened 2019-10-03
fix(links): change link with icon display - INNO-1829

yhuard:
Please double check there's no side effects

### DIFF
--- a/src/systems/ec/implementations/vanilla/packages/ec-component-language-list/ec-component-language-list.scss
+++ b/src/systems/ec/implementations/vanilla/packages/ec-component-language-list/ec-component-language-list.scss
@@ -1,6 +1,6 @@
 /**
  * Language list component
- * @define language-list
+ * @define language-list; weak
  */
 
 // Import base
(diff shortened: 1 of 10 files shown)


HAL-Patch-INNO:

Built with commit 502c163e942750839c1d9ba24c0e511a8a532ca9

https://deploy-preview-1320--europa-component-library.netlify.com

emeryro (changes requested):
Display of links on several lines is indeed better:
before
image

after
image

However, this has some side effect in several locations. Some of them are minor but still noticable (file download, footer, language list)
image

But in some cases it is breaking (before on left):
social media follow
image
social media share
image
event content types
image

So we could fix this at the component/template level, but there may be other broken cases on sites using ECL.

emeryro (approved):

planctus:
I have to say that in many of the components, the alignment of the icon is still questionable, i'd say that in some of the cases playing with that single pixel movement of the element doesn't really change the feeling that something is not in the right place, but i believe that in most of these situation the problem comes from the "wrong" proportions between text and icon (typically it happens with arrows, see breadcrumb or pagination).
Anyway, i tried to check all the components including a link implementation and i couldn't spot anything clearlt broken by this implementation.. ;)

PR #1331, opened 2019-10-10
docs: document harmonised group 1 - INNO-1850

yhuard:
Ref: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Harmonised+templates+Group+1-+Usage

### DIFF
--- a/src/website/src/pages/ec/core-template/index.mdx
+++ b/src/website/src/pages/ec/core-template/index.mdx
@@ -1,7 +1,6 @@
 ---
 order: 4
 title: Core template
-dividerBefore: true
 ---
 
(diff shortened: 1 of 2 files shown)


HAL-Patch-INNO:

Built with commit e7b0e8ab9ef7fb4d1b84a41f869074f28e4e2af1

https://deploy-preview-1331--europa-component-library.netlify.com

planctus (approved):

planctus:
@yhuard , it's just ok, the only thing is that when going mobile the table looks a bit odd, but that's valid for all these usage pages i guess...The other thing i noticed is that the dropdown we get in mobile it seems is mixing the groups and header and footer together, but nothing regarding this ticket.

PR #1368, opened 2019-10-22
feat: Escape attributes values in Ecl Twig v1

LOBsTerr:
# PR description

Escape attributes's values, because they can contain values not suitable for attributes for example:
```
{"disabled":{":input[name='override_permissions'":{"value":1}}}
```


HAL-Patch-INNO:

Built with commit 98b7a7dc47021cc12df814130170da9518c95017

https://deploy-preview-1368--europa-component-library.netlify.com

planctus:
Hello @LOBsTerr , well, we are very sensitive on this topic, especially because it's part of the decisions we are making also regarding ecl-twig, so the twig templates for ECL 2., you can see for instance this: https://github.com/ec-europa/ecl-twig/issues/165.
The thing here is that the use case you have, it's really drupal specific and i don't know why ecl is involved, this is because you are using conditional fields on the front end? Maybe you could handle this differently, but still we are about to consider this request also for ecl-twig, so that a common decision and a consistent behavior can be expected between the two.
You know how tricky is to define the right escape strategy and we'd like to avoid rushing on a solution, from this regard twig has also a specific "scope" for escaping html attributes, e(html_attr), we do some testing and we come back to you, ok? :)

LOBsTerr:
Hi @planctus, 
Yes, I've provided a Drupal specific example, but it is just a point, that ECL component can't handle correctly string with special characters passed to attributes. Of course, I can write custom js code, but it doesn't solve the initial problem.

BTW, Drupal handles this case correctly :) So, problem is not in the Drupal

planctus:
Hello again @LOBsTerr , i do wonder a bit about what you mean by "handling special characters" in the html attributes, what i mean is that the room for the extra_attributes we have for each template is not meant to do this. At the same time it's clear that being this something we receive as an input we can't control it (or we are just not doing it at the moment) and therefore an escape strategy has to be in place to avoid nasty situations.

I made some testing and i've go this results rendering via twing:

Attr:
--------------------------------------------------------
```
{ name: 'json', value: '{ json_value: "a value" }' },
{ name: 'html', value: '<a>value with tag</a>' },
{ name: 'js', value: '<script>alert("script")</script>' }
```

The resulting string is then rendered without autoescaping it. |raw

Value not explicitely escaped:
-------------------------------------------------------

  ```
html="<a>value with tag</a>"
  js="<script>alert("
  json="{ json_value: "
  script=""
  script")<=""
  value"=""
  }"=""
```


Escape with |e
-------------------------------------------------------
```

  html="<a>value with tag</a>"

  js="<script>alert(\\"script\\")</script>"

  json="{ json_value: \\"a value\\" }"
```


Escape with |e('html')
--------------------------------------------------------

```
  html="<a>value with tag</a>"

  js="<script>alert(\\"script\\")</script>"

  json="{ json_value: \\"a value\\" }"
```


Escape with |e('html_attr')
---------------------------------------------------------

```
  html="<a>value with tag</a>"

  js="<script>alert(\\"script\\")</script>"

  json="{ json_value: \\"a value\\" }"
```


Escape with |e('js')
----------------------------------------------------------

```
  html="\\\\u003Ca\\\\u003Evalue\\\\u0020with\\\\u0020tag\\\\u003C\\\\/a\\\\u003E"
  js="\\\\u003Cscript\\\\u003Ealert\\\\u0028\\\\u0022script\\\\u0022\\\\u0029\\\\u003C\\\\/script\\\\u003E"
  json="\\\\u007B\\\\u0020json_value\\\\u003A\\\\u0020\\\\u0022a\\\\u0020value\\\\u0022\\\\u0020\\\\u007D"
```

Although no difference can be noticed between the different "context" used (only js renders differently) it seems to me that being this an attempt to specifically set an escaping strategy for the html attributes values, the right approach is indeed to use |e('html_attr')
I tested this also on ecl-twig with no issues, so we are going to do the same also there, most likely.
I based this investigations on:
https://twig.symfony.com/doc/1.x/filters/escape.html

If you agree with this, you can apply these change and we are then going to merge the pr and release this in ECL 1 in the next available slot. Thanks!

**Comment by LOBsTerr** [comment]
@planctus I think you are right we have to use 

```
e('html_attr')
```

Presented new changes


what I noticed also. In many twig files this code used again and again:
```
{% if extra_attributes is defined %}
  {% for attr in extra_attributes %}
    {% set _extra_attributes = _extra_attributes ~ ' ' ~ attr.name ~ '="' ~ attr.value|e('html_attr') ~'"' %}
  {% endfor %}
{% endif %}
```
It could be refactored and replace with more reusable solution.

planctus:
ciao @LOBsTerr , you might be surprised by what i'm saying but after testing this on ecl-twig we noticed that in our scenario the simple `|e` could have been better than `|e('html_attr')`  but more than that, `|e('html_attr')` was only introduced in twig 1.9.1 and drupal uses a much older version.
Therefore we can't use it,  [ecl-twig 2.15.0](https://github.com/ec-europa/ecl-twig/releases/tag/v2.15.0 ) was released with the` |e` for  html attribute values, so i guess we can do the same here in ECL 1. ;)

Regarding the "duplication" of the code related to extra_attributes you are right, it's a bit crazy to have it repeated in each template (the same happens in ecl-twig), at the same time we kept the components "independent" from internal dependencies by not relating them or relying on blocks and child templates...we'll see :)

LOBsTerr:
@planctus Should revert the last commit then ?

planctus:
Yep, @LOBsTerr , revert it to |e,  this would align the ECL v1 twig templates with the ones in ECL v2 that we already published using this approach. Thanks! ;)

LOBsTerr:
@planctus done

planctus (approved):

PR #1383, opened 2019-10-29
fix(page-header): meta uppercase - INNO-1915

emeryro:
# PR description

Use uppercase for meta data.  
The solution chosen here is to use utilities, so that we can still have lowercase when needed.


HAL-Patch-INNO:

Built with commit 871234ae06fc8cb0b38ff503dac0429825e6e9fc

https://deploy-preview-1383--europa-component-library.netlify.com

yhuard (dismissed):

yhuard (approved):

PR #1391, opened 2019-10-31
feat(fact and figures): add component - INNO-1927

emeryro:
# PR description

specs: https://webgate.ec.europa.eu/CITnet/confluence/pages/viewpage.action?pageId=896647792


yhuard (changes requested):
overall, it looks good but it "breaks" when you activate the `#test-fake-global-rules` ;-)

yhuard (changes requested):
"View all metrics" is still affected by `#test-fake-global-rules`:/

yhuard (approved):

PR #1395, opened 2019-11-05
fix(file): amendments - INNO-1939

emeryro:
# PR description

apply https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/File+-+amendments


src/systems/ec/implementations/vanilla/packages/ec-component-file/ec-component-file.scss
```diff
@@ -19,10 +19,11 @@
 ) {
   .ecl-file {
     margin: 0;
+    max-width: 823px;
```
degliwe: Where is this coming from? (can't find it in the specs)
better way to define this?

degliwe (changes requested):

src/systems/ec/implementations/vanilla/packages/ec-component-file/ec-component-file.scss
emeryro: this value has been provided by the designers, in private message. I asked to put it in the specs too, but it hasn't been done yet.
image

yhuard (approved):

PR #1396, opened 2019-11-05
feat(utilities): add dimension and box-sizing utilities - INNO-1940

emeryro:
# PR description

Add utilities to manage some display attributes (height, width, box-sizing)


degliwe (approved):
follow up ticket created https://webgate.ec.europa.eu/CITnet/jira/browse/INNO-1947

degliwe (approved):

PR #1403, opened 2019-11-11
fix(footers): set default rules - INNO -1955

emeryro:
# PR description

Enforce default css to prevent overlap


yhuard (approved):
perfect ✅

PR #1404, opened 2019-11-11
fix(inpage-nav): remove sticky nav when out of viewport - INNO-1948

yhuard:
Solution based on IntersectionObserver.

To make use of it, simply add `data-ecl-inpage-navigation-container` to the "real" container of the nav (e.g. the ".ecl-row" parent, and not simply its direct parent), and that's it


emeryro (changes requested):
It works well, but we should add a small explanation about it on https://5dc975c5164c083fea75bb6b--europa-component-library.netlify.com/ec/components/navigation/inpage-navigation/code/ (with the js part)

emeryro (approved):

PR #1412, opened 2019-11-18
feat(event): allow to select template - INNO-1963

emeryro:
# PR description

Add radio buttons in storybook to select template for event content types


degliwe (dismissed):
pages banner at the bottom is merged with footer in some case (blue on blue) we should switch to a grey page banner in those cases

kalinchernev (commented):
Changes look fine to me in terms of visualization and knobs behavior.

To be honest, I am not able to provide a deep feedback on the possible misplacement and miscoloring of elements on the page because I don't know what the difference between core/standatized/harmoznied is. I managed to see and can verify visualization changes in the playground, but couldn't see actual difference between Standardised and Harmonised group 1 options and I don't know if that is a correct behavior in terms of concept requirements.

Here's a suggestion discussed with @yhuard for a follow-up: possibly split examples to make them easier to follow and update, for instance:

```js
import CoreEventAgendaPageExample from '../examples/Core';
import StandardisedEventAgendaPageExample from '../examples/Standardised';

storiesOf('Templates|Pages', module)
  .addDecorator(withKnobs)
  .add('Event agenda', () => {
    const template = radios(
      'Template',
      {
        Core: 'core',
        Standardised: 'standardised',
        'Harmonised group 1': 'harmonised-g1',
        'Harmonised group 2': 'harmonised-g2',
      },
      'core'
    );

    if (template === 'core') return <CoreEventAgendaPageExample />;
    if (template === 'standardised') return <StandardisedEventAgendaPageExample />;
    ...
  });
```

This way it won't be needed to have all the logic in 1 file, but rather separate focused templates.

yhuard (approved):

PR #1415, opened 2019-11-19
refactor(card): allow meta array - INNO-1967

emeryro:
# PR description

Allow use of array to define card meta (string are still allowed too)

### DIFF
--- a/src/systems/ec/implementations/react/components/card/src/Card.jsx
+++ b/src/systems/ec/implementations/react/components/card/src/Card.jsx
@@ -37,7 +37,11 @@ const Card = ({
   // Meta
   let metaMarkup = '';
   if (meta) {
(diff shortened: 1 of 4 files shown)


planctus (commented):
@emeryro , just a question, the fact that you defined the prop as string OR array is meant to "ideally" support existing implementations that are using a string for the meta? Because for me this is always an array, even with a single item and that's how we have it defined on the twig side, for instance.

planctus:
@emeryro , well, it looks like at the end of the day we have to keep this "original sin" of defining the meta for the card as a string in the specs. ECL in the end is not showing parameters anywhere, it only shows a knob where the meta is a string, up to the implementation to deal with it. On the twig side it is defined as an array only, though.

PR #1421, opened 2019-11-22
feat(site-header-core): add login - INNO-1970

emeryro:
# PR description

specs: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Core+Site+header+-+Design+Specifications+update


src/systems/ec/implementations/react/components/site-header-core/src/SiteHeaderCore.jsx
```diff
@@ -1,13 +1,17 @@
...
 import classnames from 'classnames';
 
 import Icon from '@ecl/ec-react-component-icon';
+import Link from '@ecl/ec-react-component-link';
 import SearchForm from '@ecl/ec-react-component-search-form';
 import { LanguageListOverlay } from '@ecl/ec-react-component-language-list';
 
 const SiteHeaderCore = ({
   logo,
+  logged,
```
kalinchernev, line 12: Variable communicating a state would be probably more easier to figure as `isLogged` without the documentation.

src/systems/ec/implementations/react/components/site-header-core/src/SiteHeaderCore.jsx
```diff
@@ -52,6 +56,53 @@ const SiteHeaderCore = ({
             />
           </a>
           <div className="ecl-site-header-core__action">
+            {!!(
+              loginToggle &&
+              Object.values(loginToggle).length >= 1 &&
+              loginBox
```
kalinchernev, line 62: Not as easy to follow, and a bit tricky with the "bang-bang" (`!!`). I think it's ok to skip the "truthy" typecasting at all or use Boolean if required.

The check for number of items in `loginToggle` is brittle.Two properties will make it pass, but won't be sufficient for later use. Similar concern for the `loginBox` object which can easily be truthy without having sufficient properties to be useful.

I think a longer and more verbose check can be ok and if any of the properties is missing it'd be better to skip displaying the login part at all.

src/systems/ec/implementations/react/components/site-header-core/src/SiteHeaderCore.jsx
```diff
@@ -52,6 +56,53 @@ const SiteHeaderCore = ({
...
+                      <Icon shape="general--logged-in" size="s" />
+                      {loginToggle.labelLogged}
+                    </a>
+
+                    <div
+                      id={loginBox.id}
+                      className="ecl-site-header-core__login-box"
+                      data-ecl-login-box
+                    >
+                      <p className="ecl-site-header-core__login-description">
```
kalinchernev: The div can be conditionally displayed based on presence of `loginBox.description`, if empty string is passed, the div will be rendered empty.

src/systems/ec/implementations/react/components/site-header-core/src/SiteHeaderCore.jsx
```diff
@@ -1,13 +1,17 @@
...
 import Icon from '@ecl/ec-react-component-icon';
+import Link from '@ecl/ec-react-component-link';
 import SearchForm from '@ecl/ec-react-component-search-form';
 import { LanguageListOverlay } from '@ecl/ec-react-component-language-list';
 
 const SiteHeaderCore = ({
   logo,
+  logged,
+  loginToggle,
+  loginBox,
```
kalinchernev, line 14: An additional `onLogout` handler executed on clicking on the log out would be appreciated. This way the consumer of the component can pass down a custom handler from parent component which will kill session (side effect) and still keep this component pure.

kalinchernev (dismissed):
The display and the behaviors are working fine. I've added a few comments on the react component which is not a blocker as it's only used for the style guide.

yhuard:
@emeryro you will get the new anatomy image and an update of the usage guidelines soon

update: or maybe not soon... in any case, can you update the table to add the optional login? thanks!

src/systems/ec/implementations/react/components/site-header-core/src/SiteHeaderCore.jsx
emeryro: true, but I kept the naming that was in place for other site headers.
If we change the parameter name, that would be considered as a breaking change (unless ke keep the two parameters)

src/systems/ec/implementations/react/components/site-header-core/src/SiteHeaderCore.jsx
emeryro: ideally, every properties should be checked, indeed.
But as react packages are not released (they all are set to private), it is only for internal use. So we kind of control the data the pass.
So I'd consider it as a nice to have, when we have time (there are probably lots of similar cases in other components)

src/systems/ec/implementations/react/components/site-header-core/src/SiteHeaderCore.jsx
emeryro: As far as I know, login and logout will be handled by EU login, so it will probably be a link with some parameters.
I agree that adding handlers could be a good improvement (not only for this component), but I'd rather create a nice to have ticket for this

kalinchernev (approved):
Updated usage pages as well as images for Anatomy.

PR #1426, opened 2019-11-25
feat(datepicker): Add component - INNO-1599

emeryro:
# PR description

First step: [RFC](https://github.com/ec-europa/europa-component-library/blob/feat/datepicker-INNO-/docs/decisions/004-datepicker.md)

Second step: implementation
specs: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Date+picker+-+Design+specifications


kalinchernev:
My suggestion would be to spend the least amount of time on customization for the following reasons:

- Avoid making the same mistakes library authors have already solved.
- Frameworks consuming ECL will most probably not use this date picker at all, but a framework-specific one. I'm pretty sure jquery datepicker is still used more than pikaday in target platforms.
- We don't really have the capacity at the moment.

Pikaday was chosen for v1 because of its theming capabilities and good API. Looking at the spec, there aren't any problems keeping it, but please correct me if I'm wrong and maybe we can search for solutions to specific issues.

emeryro:
A first implementation has been made, waiting for feedback from design team (see comments on the wiki page)

emeryro:
Component has been validated by management, ready for technical review

kalinchernev (dismissed):
The technical implementation looks good to me :+1: 

Making the review, a few questions came to mind, they are not blockers:
- Missing API section for documenting JavaScript usage of the wrapping component
- Do we need to maybe update the landing page of the components' listing to have a simplified image of the new component?
- Where does width (small/medium/large) option come from? Maybe there will be full width 100%?
- Shall we allow for options' spread from wrapper ECL component down to pikaday? This could potentially allow for the user to modify pikaday options directly. This can potentially open new work to be done on styling features which are not used.

emeryro:
Thanks for the review.
- I added the jsdoc
- We need to update the landing page indeed. I think that several components are missing. I'll check with the designers (could be done in another task)
- The width are coming from the form input. After discussion with the designers, width has been forced to "S"
- There are already a few options available to control datepicker parameters. If needed we could expose some more later or find better ways to deal with it.

emeryro:
design and behavior has been approved by the designers: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Date+picker+-+Design+specifications?focusedCommentId=907437570#comment-907437570

kalinchernev (approved):

kalinchernev:
@emeryro thanks much for the following up through the whole process of having this component in!

PR #1427, opened 2019-11-25
chore(website): attempt to solve infrastructure mishaps - INNO-1979

kalinchernev:
# Changes summary

1) Breadcrumbs and Footer partials updated

In a way to not lead to github pages any more. This was noticed while working on documentation which was not accessible any more because not taking into account the specific version of the site.

2) index.html file mirror of each html file

With the aim to have same working pattern as `/ec` root
Breaking down the working example: `/ec` -> `/ec/` => `/ec/index.html`, shown as `/ec/`

3) Updated paths

After having a change of folder, mirror `index.html` have to have some links updated: scripts, link tags, links, iframes, etc.

## Existing example

https://ec.europa.eu/component-library/v1.15.0-beta.3/

## Testing

Because the fixes suggested in this pull request are specific to how website is prepared before going to production, testing will have to happen in 2 stages with possible repetitions and tweaks of this existing code.

To test locally, set your environment as production:

```
export NODE_ENV=production
```

Build the site:

```
yarn dist
```

Then see site locally

```
http-server dist/website
```

Expected is that one can navigate through left side bar from component to component without any problems. Also, previews show well and raw/preview templates are linked correctly from internal display of components/temlates/utilities, etc.

When code changes are approved, the only way we can validate whether changes are ok for the actual infrastructure (not possible to reproduce locally) we make a new temporary tag to build and deploy a test.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -93,6 +93,15 @@ pipeline:
   # Tag ===================================
   # =======================================
 
+  prepare-hosting:
+    image: node:dubnium
+    environment:
+      - NODE_ENV=production
+    commands:
+      - node scripts/prepare-europa-hosting.js
+    when:
+      event: tag
+
   archive-website:
     image: alpine:3.10
     commands:
--- a/.drone.yml
+++ b/.drone.yml
@@ -95,8 +95,6 @@ pipeline:
 
   prepare-hosting:
     image: node:dubnium
-    environment:
-      - NODE_ENV=production
     commands:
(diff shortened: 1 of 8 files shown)


papegaill (dismissed):
Successfully tested on my side, Able to navigate from component to component and all previews/raw are correctly rendered and linked to displays.

kalinchernev:
Thanks @papegaill for the review! I made a small update for fixing the lint issues and merged to test the next step!

PR #1430, opened 2019-11-28
feat(menu): add component - INNO-1981

emeryro:
# PR description

Add standardised and harmonised group 1 menu.  
Menu are provided as separate packages, but demo as part of the site header on the website.

specs: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Menu+-+Design+Specifications


kalinchernev (commented):
Didn't find any issues with code or re-implementation. Can't quite understand a few points however, I'm sharing my concerns and questions below.

In `src/systems/ec/implementations/vanilla/packages/ec-component-menu-legacy/ec-component-menu-legacy.scss` implementation is based on `@mixin ecl-menu-legacy()`, whereas `src/systems/ec/implementations/vanilla/packages/ec-component-menu-harmonised/ec-component-menu-harmonised.scss` doesn't have this approach.

Both standardised and harmonised site headers make use of a single menu, i.e no differences.

New menu(s) not included in the website whereas `src/website/src/pages/ec/components/dropdown-legacy/index.md` is `hidden: true`

I'm not saying the duplication is wrong, but the reasoning is not obvious and implementation (`usage`) guidelines are not accessible which makes me wonder what's the goal of this iteration?

I'm not knowledgeable of the "harmonised", "standardised" concept, however if I'm to be on the implementers' side, I'd definitely appreciate a note about what's the difference between those and when to choose which, given the fact they are currently identical. Not having the wrapping mixin which can potentially make styles reusable for other systems is deeming the current state of results to be 100% certainly reworked quite a lot.

emeryro:
In short:
- I'm not sure to understand what you mean about the different approach for scss. Both files define a mixin. The difference is that the menu legacy exposed some variables in the mixin. That was done in hope of different color set for ECL ("themes"), but this kind of thing has been abandoned.
- Menu is indeed the same for standarised and harmonised group 1
- At first I wanted to add a dedicated page in the website for menu, but menus are already part of the site header documentation (in the screenshots). If later we have a dedicated documentation it would make sense to add a page, but for now I'm not sure.
- Concerning the differences between the families... I never understood them either :) That's a business decision.

kalinchernev (approved):

kalinchernev:
@emeryro thanks for the clarifications. Given the unknown future of harmonized and standardised, etc. components, I merged and hope that if will still be better like this potentially accommodating having a bigger difference in implementation.

Thanks also for the clarification on the obsolete theming concept, that's what I was wondering about, good to know.

PR #1438, opened 2019-12-02
fix(ie11): fix js error - INNO-1988

emeryro:
# PR description

Fix javascript error on some component, on IE11 (it was only visible on the website).  
Error message: Object doesn't support property or method 'values'  
Components: 
- fact and figures
- site header core
- site header harmonised
- site header standardised


yhuard (approved):
I confirm the bug is fixed

PR #1440, opened 2019-12-03
feat(department): add content type - INNO-1993

emeryro:
# PR description

specs are on [the ticket](https://webgate.ec.europa.eu/CITnet/jira/browse/INNO-1993)


kalinchernev (approved):
- code check: no issues
- components check: no mismatch found between versions of standard, harmonised, etc. (including groups)
- deps checked: no issues
- knobs in storybook: no issues

I got the feeling the screens (specs) in the ticket are not using same fonts, colors, etc., but comparing existing ECL implementations, no mismatch, so the ECL implementations for site/page headers, etc. are the reference for me when checking the template.

PR #1448, opened 2019-12-06
fix(site-headers): Add arrow icon to login link - INNO-1996

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


src/systems/ec/implementations/vanilla/packages/ec-component-site-header-core/ec-component-site-header-core.scss
```diff
@@ -78,6 +78,12 @@
     }
   }
 
+  .ecl-site-header-core__login-toggle {
+    .ecl-icon--xs {
```
kalinchernev: Doesn't look quite right to me to be honest. Styling an icon inside current component is kinda scope leak. Although it's been quite long since my last work on these, I think it can be better either to hide elements with utilities or if necessary make a new sub-component element such as `ecl-site-header-core__subelement` to avoid the leak of specificity between components.

Edit: Seems quite ok to apply transform on sub-element: https://github.com/ec-europa/europa-component-library/blob/5310d9dd62989a81f1be8521acc38809203d0aa2/src/systems/ec/implementations/vanilla/packages/ec-component-file/ec-component-file.scss#L152

papegaill: Speaking here, apparently it's necessary to avoid utilities classes in components.
Also, there is already a general css rule for the left _logged in_ icon that act on all icons inside that link: https://github.com/ec-europa/europa-component-library/blob/a23635a46490c09fee01cc017d95e10f6b80de0d/src/systems/ec/implementations/vanilla/packages/ec-component-site-header-core/ec-component-site-header-core.scss#L214 this is why I needed to declare properties for xs icon.

emeryro: Although we had to apply css rules on sub elments sometimes, it is indeed better to use a dedicated css class for this.
In this case, we could have something like
`<Icon shape="ui--corner-arrow" size="xs" className="ecl-site-header-standardised__login-icon" />` or something similar. same for the other icon (I don't remember why it has been done like this, but this is a good opportunity to clean it a little).
The trick is that we have to keep the old css rules, to prevent breaking change, but we can mark them as deprecated and remove them in a few version

papegaill:
@kalinchernev, @emeryro, Thank you for the review and suggestions. I'm now using a specific class for the new icon as suggested. I also changed the size of the existing icons at the markup level `s` to `m`, rather than overwrites theme via css.

src/systems/ec/implementations/vanilla/packages/ec-component-site-header-core/ec-component-site-header-core.scss
```diff
@@ -78,6 +78,12 @@
     }
   }
 
+  .ecl-site-header-core__login-toggle {
```
kalinchernev, line 81: I'd personally suggest this to be `ecl-site-header-core__login-toggle-arrow` or `ecl-site-header-core__login-arrow`, there are many `icon`-s already. Just a suggestion, not a blocker.

kalinchernev: Btw, the code after `{logged && ()}` is a candidate for a separate component by itself, as `logged` prop is wrapping quite some logic already.

src/systems/ec/implementations/vanilla/packages/ec-component-site-header-core/ec-component-site-header-core.scss
```diff
@@ -205,10 +211,21 @@
 
     .ecl-site-header-core__login-toggle,
     .ecl-site-header-core__language-icon {
-      .ecl-icon {
-        height: $ecl-icon-m;
+      .ecl-icon--m {
         margin-right: $ecl-spacing-xs;
```
kalinchernev, line 216: Reading this line and imagining `ecl-site-header-core__language-icon` and `ecl-icon--m` specificity inside, I'm wondering: what about the other icons? What if someone changes the size, then there will be impact on this. Can we avoid it? Maybe both icons can take the `ecl-site-header-core__icon` which hold the size and can be somehow styled this specifically?

src/systems/ec/implementations/vanilla/packages/ec-component-site-header-core/ec-component-site-header-core.scss
```diff
@@ -205,10 +211,21 @@
...
-        width: $ecl-icon-m;
+      }
+    }
+
+    .ecl-site-header-core__login-toggle {
+      .ecl-site-header-core__login-icon {
+        display: inline-block;
```
kalinchernev: `.ecl-site-header-core__language-selector` and `.ecl-site-header-core__login-toggle` are `display: flex;` so I think this should be `display: inline-flex` but please double-check.

kalinchernev (commented):
Looks quite nice, I placed a few comments mostly in thinking of changes in time, mostly about the sizes. In general it's ok for me

src/systems/ec/implementations/vanilla/packages/ec-component-site-header-core/ec-component-site-header-core.scss
papegaill: @kalinchernev thanks for the suggestion, updated to `ecl-site-header-core__login-arrow`

src/systems/ec/implementations/vanilla/packages/ec-component-site-header-core/ec-component-site-header-core.scss
papegaill: You are right, better to keep forcing size on these icons. I'm now using a `ecl-site-header-core__icon` class for general headers icons and forcing `m` size with css

src/systems/ec/implementations/vanilla/packages/ec-component-site-header-core/ec-component-site-header-core.scss
papegaill: All headers icons are displayed as `block`, therefore I updated in this way.

papegaill:
@kalinchernev, many thanks for the review, I updated the PR, can you have a new review?

kalinchernev (approved):
@papegaill thanks for the updates!

PR #1453, opened 2019-12-12
feat(footer-harmonised): add group 3 - INNO-2016

emeryro:
# PR description

specs: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Harmonised+Footer+Group+3+-+Design+Specifications


papegaill (dismissed):
Hi @emeryro the label "_Joint partnership_" should not be a link, confirmed with Filipa. I also see that this label is not correctly centered as requested on the mocks >=480px. Also I think there is an issue when we'll have multiple logo on mobile version, currently they stay on a line, I think we should plan the case with several logos, see tickets for more information.

kalinchernev:
Following up on https://github.com/ec-europa/europa-component-library/pull/1453#pullrequestreview-331900056

- "the label "Joint partnership" should not be a link" => changed
- "label is not correctly centered as requested on the mocks >=480px" => checked and it looks ok

Until (rougly) 700px width, it's in this case

```
Mobile version < 480 px
--
If the items exceed the width of the view port
```

Then it's `align-items: center;` and ok as far as I can see. If you delete one image place holder to have only 1 image and 1 logo, it does match behavior to up to 480 when elements have enough space.

- "I think there is an issue when we'll have multiple logo on mobile version, currently they stay on a line, I think we should plan the case with several logos, see tickets for more information."

Not quite clear, please clarify

papegaill:
@kalinchernev PR updated, please check my last comment on the ticket for updates information.

kalinchernev (approved):
@papegaill thanks for the updates! (and the explanations in the ticket!)

PR #1461, opened 2019-12-19
docs(facts-figures): add documentation - INNO-2033

emeryro:
# PR description

reference: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/2019-11+Facts+and+figures+-+ECL+Page+-+final

### DIFF
--- a/src/website/src/pages/ec/components/fact-figures/docs/usage.mdx
+++ b/src/website/src/pages/ec/components/fact-figures/docs/usage.mdx
@@ -20,11 +20,11 @@ import { Anatomy, Paragraph } from '@ecl/website-components';
     items: [
       {
         color: '#404040',
-        label: 'Mandatory',
+        label: 'mandatory',
       },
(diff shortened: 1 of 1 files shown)


kalinchernev (changes requested):
Please update the following two sections:

Screenshot_20191219_103748

The labels should be lowercase: optional, mandatory.

Elements should link to existing ECL components when there are such.

emeryro:
I updated the labels.
But I don't think that there should be links to components in this table. Elements here are just part of the current component, not necessarily a component as such (or maybe an altered version of it).
To me, if there are links to other components, they should be part of the "Related components" section at the end of the page. But this is empty here on the reference wiki (or is it up to us to list them?)

kalinchernev (approved):

kalinchernev:
@emeryro regarding links to existing components/elements: not blocker.

PR #1471, opened 2019-12-20
docs(button): update usage page - INNO-2038

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


src/website/src/pages/ec/components/button/docs/usage.mdx
```diff
@@ -3,70 +3,310 @@ title: Usage
...
+# Search Button
+
+<Paragraph size="lead">
+  The search button is only used with the search form.
+</Paragraph>
+
+## Anatomy
+
+<Anatomy
+  image="https://inno-ecl.s3.amazonaws.com/media/images/EC/Buttons/button-search-anatomy-desktop.png"
```
emeryro, line 251: Images for mobile and desktop version are different here, so we should use an imgset (see https://github.com/ec-europa/europa-component-library/pull/1484/files#diff-af5dd2c1617b773729988b532636b636R21 for instance)

emeryro (changes requested):
Just a small modification for the images, but apart of that this is good

src/website/src/pages/ec/components/button/docs/usage.mdx
papegaill: Thank you for the review @emeryro, PR updated. That said, I don't quite understand the approach of using imgsrc .. so if you browse the style-guide on mobile you get mobile images .. doesn't make any sense.. even on desktop, I should be able to see mobile images and directive, what do you think?

kalinchernev (approved):

src/website/src/pages/ec/components/button/docs/usage.mdx
kalinchernev: When we looked at the images on Friday with Alexis, there wasn't any difference at this moment between desktop and mobile. For me the PR is ok

emeryro: We would probably need a better way to display mobile/desktop version, I agree. Let's consider this as a temporary solution

emeryro (approved):

PR #1484, opened 2019-12-23
docs(dateblock): add documentation - INNO-2033

emeryro:
# PR description

reference: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/2019-12+Date+Block+-+ECL+Page+-+in+progress


planctus (changes requested):
The page looks good, we just miss the "bolded" text in the main description of the component, then i read your comment about the "notes" section, not sure you received an answer about that but i'm fine with it.

emeryro:
I remember that Filipa asked us to remove all inline bold text from documentation (after a report from accessibility team).
I'll try to find some trace of that...

planctus:
Not that i would have anything against the idea of removing those bolded things, but since we received these tickets with this style we just introduced many new use cases of bolded text in our docs paragraphs right in these days. If they just gave us "specs" like this, i would not consider a comment given in the past by one single designer enough to change them, otherwise let's try to get an official decision about this and we do it confidently and consistently, imho.. :)

planctus (approved):

planctus (approved):

PR #1494, opened 2019-12-24
fix(person-content-type): reorder element on website - INNO-2063

emeryro:
# PR description

Change display order of elements in website's navigations.
Also allows three levels of navigation


kalinchernev (commented):
Changes are ok for me, I'm only not sure whether Job/Person is one item or not. The question has been addressed in the ticket. Please keep updates for levels' styling regardless.

kalinchernev (approved):

PR #1496, opened 2019-12-31
refactor(content-item): use template - INNO-2019

emeryro:
# PR description

Allows content item to be used directly in other templates.

- content-item composition has been updated to allow several parameters, and different variant (an example has been added for each one)
- content types using content items have been updated to use the template directly


src/systems/ec/implementations/react/templates/pages/commemorative-coin/src/CommemorativeCoinPage.jsx
```diff
@@ -111,194 +119,122 @@ const CommemorativeCoinPage = ({
...
-                style={{
-                  backgroundImage:
-                    'url("https://inno-ecl.s3.amazonaws.com/media/examples/example-image.jpg")',
-                  backgroundRepeat: 'no-repeat',
-                  backgroundSize: 'contain',
-                  width: '7.5rem',
-                }}
-              />
+            <ContentItem
+              meta={{
```
papegaill, line 124: Compared to the current version the following classes are missing 
`className={"ecl-u-mt-m ecl-u-mt-lg-l"}`

src/systems/ec/implementations/react/templates/pages/department/src/DepartmentPage.jsx
```diff
@@ -260,72 +263,60 @@ const DepartmentPage = ({ siteHeader, footer, template }) => (
...
+                  href: '/example',
+                  label: 'Annual activity report',
+                }}
+                description={{
+                  label:
+                    'Department achievements, resources used during previous year',
+                }}
+              />
+
+              <ContentItem
```
papegaill, line 310: on current version there is no bottom border on this last article, but it seems good to me like this.

src/systems/ec/implementations/react/templates/pages/department/src/DepartmentPage.jsx
```diff
@@ -373,105 +364,102 @@ const DepartmentPage = ({ siteHeader, footer, template }) => (
...
-                <div
-                  role="img"
-                  aria-label="Example image"
-                  className="ecl-u-flex-shrink-0 ecl-u-mr-s ecl-u-media-a-s ecl-u-media-bg-size-contain ecl-u-media-bg-repeat-none"
-                  style={{
-                    backgroundImage:
-                      'url("https://inno-ecl.s3.amazonaws.com/media/examples/example-image-square.jpg")',
-                  }}
-                />
+              <ContentItem
```
papegaill, line 367: much better like this, without left and right padding that we have on current version ;)

src/systems/ec/implementations/react/templates/pages/job-person-list/src/PersonListPage.jsx
```diff
@@ -228,181 +229,173 @@ const PersonListPage = ({ siteHeader, footer, template }) => (
...
+              </h3>
+              <p className="ecl-u-type-paragraph ecl-u-type-uppercase ecl-u-type-color-grey-75 ecl-u-mt-none ecl-u-type-s">
+                Deputy Head of Cabinet
+              </p>
+              <DescriptionList variant="horizontal">
+                <DescriptionTerm>Phone number</DescriptionTerm>
+                <DescriptionDefinition>+32 2 29 64816</DescriptionDefinition>
+                <DescriptionTerm>Responsibilities</DescriptionTerm>
+                <DescriptionDefinition>
+                  <UnorderedList className="ecl-u-mv-m">
```
papegaill: This is already the case, but here we must remove top margin that we don't have on the previous and next row's lists. Should be replaced by `ecl-u-mb-m`

On all the lists, I would also force left padding to better align it with the content: `ecl-u-pl-m`

src/systems/ec/implementations/react/templates/pages/main-policy-hub/src/MainPolicyHubPage.jsx
```diff
@@ -783,130 +717,119 @@ const MainPolicyHubPage = ({ siteHeader, footer, template }) => (
...
+              <ContentItem
+                title={{
+                  href: '/example',
+                  label: 'Aurel CIOBANU-DORDEA',
+                }}
+                description={{
+                  className: 'ecl-u-type-s',
+                  label: 'Director',
+                }}
+                images={{
```
papegaill, line 759: much better like this, profile image and without left and right padding ;)

src/systems/ec/implementations/react/templates/pages/research-area-hub/src/ResearchAreaHubPage.jsx
```diff
@@ -293,55 +294,47 @@ const ResearchAreaHubPage = ({ siteHeader, footer, template }) => (
...
+                    }}
+                    description={{
+                      label:
+                        'Marine research related calls may be found in this programme, particularly in the European maritine and fishieries fund.',
+                    }}
+                  />
                 </div>
               </div>
 
               <div className="ecl-row">
```
papegaill, line 325: Every second row below have an `ecl-u-mt-m` classes, this one doesn't. Personally, I don't understand why a margin-top is defined on the following rows. This creates a larger visual gap at the top on all following rows.

papegaill (changes requested):
Nice job here, I reviewed content-types according to the currently published version. Some spacing etc ... some remarks may need to be the subject of a new ticket.

src/systems/ec/implementations/react/templates/pages/search/src/SearchPage.jsx
```diff
@@ -189,145 +189,119 @@ const SearchPage = ({ siteHeader, pageHeader, footer, template }) => (
...
-                For the first time, in 2014, EU countries had to take the
-                results of the elections into account when proposing a candidate
-                for European Commission President.
-              </p>
-            </article>
+
+            <ContentItem
+              title={{
+                href: '/example',
+                label: 'Jobs, growth and investment',
```
papegaill, line 196: link is no more underline like current version?

src/systems/ec/implementations/react/templates/pages/search/src/SearchPage.jsx
```diff
@@ -189,145 +189,119 @@ const SearchPage = ({ siteHeader, pageHeader, footer, template }) => (
...
-                label="Justice and fundamental rights"
-                className="ecl-u-type-prolonged-m ecl-u-type-bold"
-              />
-              <p className="ecl-u-type-paragraph-m ecl-u-type-color-grey ecl-u-mv-none">
-                Every EU citizen enjoys the same fundamental rights based on the
-                values of equality, non-discrimination, inclusion, human
-                dignity, freedom and democracy.
-              </p>
-            </article>
-            <hr className="ecl-u-mv-none" />
```
papegaill, line 289: much better like this, by removing the `<hr>` which was not stylish.

papegaill (changes requested):

src/systems/ec/implementations/react/templates/pages/department/src/DepartmentPage.jsx
emeryro: as there are border bottom for all other elements on the page, I guess we can keep it this way

src/systems/ec/implementations/react/templates/pages/research-area-hub/src/ResearchAreaHubPage.jsx
emeryro: I agree, these margins are not needed

src/systems/ec/implementations/react/templates/pages/search/src/SearchPage.jsx
emeryro: well, it is not underlined for all other content items. We could make an exception here, but I think we should be consitent. (though I agree that links should be underlined in theory; it may be updated in content item template directly)

emeryro:
Thanks for the detailed review.
PR updated accordingly

papegaill (approved):

PR #1505, opened 2020-01-14
feat(storybook): expose stories' markup in dedicated paths - INNO-2068

kalinchernev:
# PR description

Because the ECL website's `Playground` component is used within MDX it's not easy to extract rendered output of src/website/src/website-components/Playground.jsx without rewrite of how the documentation is composed. (`Playground` items to be taken out of the mdx to be separate entities)

Instead of reworking ECL website, I decided to try to provide the stories markup directly in Storybook. This way one won't need to write logic of matching ECL website to storybook paths, because paths in storybook will already be matching target list for comparison with Twig.

The addon used: https://github.com/whitespace-se/storybook-addon-html
Please let me know if this solution is feasible or if additional formatting should be applied to the markup?

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/systems/ec/implementations/react/storybook/.storybook/addons.js
+++ b/src/systems/ec/implementations/react/storybook/.storybook/addons.js
@@ -3,5 +3,4 @@ import '@storybook/addon-knobs/register';
 import '@storybook/addon-viewport/register';
 import '@storybook/addon-a11y/register';
 import '@storybook/addon-cssresources/register';
-import '@whitespace/storybook-addon-html/register';
 import '../../../../../../tools/storybook-addon-code/register';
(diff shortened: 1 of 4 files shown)


planctus:
@kalinchernev, we already show the html in storybook trough a custom plugin: 
https://github.com/ec-europa/europa-component-library/tree/v2-dev/src/tools/storybook-addon-code
(Not sure we still need to use our custom one, since this "official" plugin is available and we could most likely add syntax highlighting also to that).
Anyway, regarding the way we retrieve the html of the story from ecl-twig this would not help or change much, we would still have to reach the page with an headless browser and click on the link to reveal the html. 
Maybe what we could do (but is much less relevant) is to provide a toggler for switching from the syntax highlighted format to the raw in the current plugin we use, this way we would not have to clean and play with the html when we retrieve it, but as i was mentioning in the ticket, for how sub-optimal it might be, it works, so it is not really necessary.
Thanks for investigating the possibility, anyway, making diffs with the ecl stories is really an added value on the twig side, it's a bit tricky and we can't integrate in the browser, but we are already happy having the command line script ;)

kalinchernev:
Thanks @planctus as I thought, but had to validate and not go too deep before having ensuring understanding. Will continue on the path to have markup in a separate html files next to the playground cases.

PR #1506, opened 2020-01-14
fix(storybook): unset sidebar customization - INNO-2006

kalinchernev:
# PR description

I checked github history for the reason of having the code saving sizes and couldn't find a matching ticket or request, must have been aimed at previewing stories with handles with shorter labels.

## QA Checklist

- [ ] I have provided tests


emeryro:
After discussion, it appears that the behavior here is still the same: right panel on storybook has a default size, but can be resized by the user. This new width is then saved and used on page reload or when browsing other components.

An improvement could be to have a min-width for the panel, so it can't become unreadable. To be investigated

kalinchernev:
Further investigations can be summed as following:

- Sizes of panels are managed via theming which is based on settings stored in localstorage and later applied in a way that no native CSS can work solve the issue.
- Existing values for sizes from localstorage are now ignored desired sizes are set at each page load.

emeryro (approved):

PR #1507, opened 2020-01-16
refactor(content-types): remove logic - INNO-1977

emeryro:
# PR description

Remove logic from content type templates, and use different files to manage possible displays


kalinchernev (changes requested):
Thank you for the massive refactoring!

Could you please double-check `Standardised` and `Harmonised group 1` variants? All pages in the storybook are having differences in current versions in https://ec.europa.eu/component-library/playground/ec/?path=/story/templates-pages--call-for-proposals (for instance)

emeryro:
I checked these two variants, but the only difference i can see is the presence of the site name, which has been added previously in https://github.com/ec-europa/europa-component-library/pull/1504 (but not released yet).
That would be convenient to be able to see current v2-dev branch, to compare...

kalinchernev (approved):

PR #1513, opened 2020-01-20
fix(data): fix data file name - INNO-2081

emeryro:
# PR description

Use double dash for data file name, instead of simple dash


planctus (dismissed):

planctus:
I was about to merge this, @emeryro, i just noticed that in EU we still have this inconsistency, it doesn't affect twig since the EU components are simply not there, but let's align also that, what do you think?

planctus (approved):

PR #1518, opened 2020-01-23
feat(content-types): manage optional items - INNO-1878

emeryro:
# PR description

Add checkboxes in storybook to display/hide optional elements of templates (menu, login, search, ...)
image
The idea is to be able to see templates in all possible configuration, and spot potential issues.

Notes:
- As the status of the optional items for the footer is still unclear, it hasn't been handled here. That could be added later in another PR
- There are some display issues when some optional elements are not displayed (mostly missing margins). That will be handled case by case in another PR, to keep this one small enough


papegaill (approved):
Really nice feature, as discussed we'll see if we can add harmonised G3 in the future.

PR #1520, opened 2020-01-24
chore(storybook): upgrade to 5.3 - INNO-2075

kalinchernev:
# PR description

Ensure clear re-installation. (`rm -rf node_modules` + `yarn install`)

Local development with `yarn start:ec` and `yarn start:eu` should have no issues, as well as build (validated already by the CI)


emeryro (commented):
overall it seems fine, I just have an issue when running `yarn start:ec`: instead of opening the localhost url in the browser (as it is on v2-dev), I get this error
image
("Windows does not find '\http://localhost:6006/\'. Check you entered the correct name and try again.")
That's not a blocker, but if you have any idea how to solve it that may be fine

kalinchernev:
@emeryro I managed to reproduce the issue in WSL 1 and debugging it, turns out that storybook's start script uses `opn` which is now `open` under the hood.

There are known issues with Windows and I tried to patch the version of opn and open to have [latest fixes](https://github.com/sindresorhus/open/commit/d8a508be1b23c80b53d0089b6f1e487e84f53a19) but these don't help. 

Also, there is an open issue with relation to the exact issue here https://github.com/sindresorhus/open/issues/154

In summary, the issue is not coming from storybook itself, but `open` package and how the host OS opens default browser.

In my case, this code

```js
const open = require("open");

(async () => {
  // Opens the URL in the default browser.
  await open("http://localhost:6000");
})();
```

Worked well in Linux, but didn't do anything in WSL, because it's treated as Linux but there seems to be no link between host and subsystem.

papegaill (approved):
no issue running `yarn start:ec`, successfully tested on IE11 and Edge.

PR #1522, opened 2020-01-28
feat(eu-grid): implement grid - INNO-2078

emeryro:
# PR description

Implement fixed and flexible grid, based on https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/3.+Grids

playground:
- https://5e2ffbb6050bb36c673d251e--europa-component-library.netlify.com/playground/eu/?path=/story/layout-grid--fixed-grid
- https://5e2ffbb6050bb36c673d251e--europa-component-library.netlify.com/playground/eu/?path=/story/layout-grid--fluid-grid


papegaill (approved):
@emeryro For me the grid is good and perfectly matches the specs. My only concern, is on the 4-column mobile grid. It can makes a major difference with the EC grid. It will also not be possible to use 4 and 8 columns classes on the mobile version? Because here by changing the classes, I'm able to use them on mobile:

Screenshot 2020-01-28 at 14 17 05

Screenshot 2020-01-28 at 14 19 35

I don't really understand the principle.. why display a 4 cols grid, if it is actually a 12 cols grid like EC, it will cause confusion in general. I think we should display a grid of 12 like the one used in ECL, what do you think?

emeryro:
My idea is as follow:
On a technical point of view, we keep a 12 columns grid for all breakpoint. But the guidelines will enforce the usage of only 4 columns on mobile (so `ecl-col-3`, `ecl-col-6`, `ecl-col-9` and `ecl-col-12`)
So indeed, it it technically possible to use different set of columns on mobile, but that would not be supported by the guidelines.
That's the best I could find to follow the specs (of course if we could convince everyone to use 12 columns for all breakpoint, as it is now, that would be even better)

papegaill:
Ok thank you for clarification. We’ll need to discuss this with them, so that they understands the principle. I think we really have to stay on a grid of 12 (as it is) and just make necessary recommendations in guidlines.

src/systems/eu/implementations/vanilla/packages/eu-layout-grid/mixins/_grid.scss
```diff
@@ -2,19 +2,19 @@
...
 
-  @each $breakpoint in map-keys($gutters) {
+  @each $breakpoint in map-keys($margins) {
     @include ecl-media-breakpoint-up($breakpoint) {
-      $gutter: map-get($gutters, $breakpoint);
+      $margin: map-get($margins, $breakpoint);
 
-      padding-left: ($gutter / 2);
-      padding-right: ($gutter / 2);
+      padding-left: $margin;
```
kalinchernev, line 16: Not quite getting this part of the implementation in using margin to padding. What's the idea behind?

emeryro: `margin` is just the naming they use on the specs; I kept is as is. That's the space around the grid (left and right)

kalinchernev: it's it that the gutter is the spacing between grid items?

steddybear:
> @emeryro For me the grid is good and perfectly matches the specs. My only concern, is on the 4-column mobile grid. It can makes a major difference with the EC grid. It will also not be possible to use 4 and 8 columns classes on the mobile version? Because here by changing the classes, I'm able to use them on mobile:
> 
> Screenshot 2020-01-28 at 14 17 05
> 
> Screenshot 2020-01-28 at 14 19 35
> 
> I don't really understand the principle.. why display a 4 cols grid, if it is actually a 12 cols grid like EC, it will cause confusion in general. I think we should display a grid of 12 like the one used in ECL, what do you think?

In principle, the purpose of a grid system is to support the design ensuring the best possible organization of content and consistency across different layout and devices (keeping into account the reflow of different components, for example: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Cards+reflow+and+spacing+-+EU+website).  Considering that the EU design has a different need compared to EC I don't see the point to maintain 12 columns on mobile devices and not to explore the same approach adopted by Material Design and Carbon Design systems, but it's something we can further discuss.

papegaill:
Here is the principle used by [material.io grid](https://material.io/develop/web/components/layout-grid/): 

> You can set the cells span by applying one of the span classes, of the form `mdc-layout-grid__cell--span-{columns}`, where `{columns} `is an integer between 1 and 12. **If the chosen span size is larger than the available number of columns at the current screen size, the cell behaves as if its chosen span size were equal to the available number of columns at that screen size.** If the span classes are not set, `mdc-layout-grid__cell` will fallback to a default span size of 4 columns.

material-io-mobile-grid-layout

emeryro:
To me, the proposed implementation offers what is required:
- we follow the specs
- we don't reinvent the wheel and base on what is existing
- class names are consistent and users of ECL EC won't have to learn a new syntax

Using 4 columns or not on mobile is mostly a question of guidelines. So it may even evolve later if we discover new needs.

papegaill:
@emeryro implementation approved during today ecl weekly meeting.

PR #1525, opened 2020-01-31
fix(ec-editor): fix list and paragraph - INNO-2104

emeryro:
# PR description

fix small visual inconsistencies in ec-editor for lists and paragraphs


src/systems/ec/implementations/vanilla/packages/ec-editor/_ec-editor--typography.scss
```diff
@@ -58,6 +58,7 @@
    * Paragraph
    */
   .ecl-editor p {
+    color: $ecl-color-grey-100;
```
kalinchernev: All looks ok to me, except I'm not sure about the color. Checking content types which I guess are going to be an example to compare for the requester in the case, it's `ecl-u-type-color-black` in addition to `ecl-u-type-paragraph`

From this and mapping back the black from [the loop](https://github.com/ec-europa/europa-component-library/blob/7f867ec68a8fb04fa65594ef9a42b422d87d1c00/src/systems/ec/implementations/vanilla/packages/ec-utility-typography/ec-utility-typography.scss#L137-L144) black from src/systems/ec/themes/default/exports/colors.scss is `'black': rgb(0, 0, 0),` and not grey 100 which is `'grey-100': rgb(64, 64, 64),`

emeryro: Yes, the situation about the colors is not very clear to me too.
Indeed, most paragraphs in content types are displayed in black. But text in several components is in grey, so display is not consistent.
I tried to get feedback from DG COMM on this, but got no answers yet.

kalinchernev: Maybe it'll be safer for now to not put a color and put a comment in the ticket that this particular request is not accommodated because it's meant to be like this (inheriting color) rather than putting a rule which is not consistent. It's my opinion at least

kalinchernev (approved):

PR #1529, opened 2020-01-31
docs(page-header-harmonised-1): update doc - FRONT-144

emeryro:
# PR description

Update documentation


kalinchernev (approved):

pitommasi:
Why lots of components, such as Site header, page header, and so on don't have a sample code to be used for the actual implementation?
How am I supposed to add such components to a website?
Other than copying the html from the playground?

PR #1543, opened 2020-02-07
fix(footer): handle optional elements - FRONT-467

emeryro:
# PR description

Ensure that display is not broken if some optional elements are not present.

What has been done:
- update of the markup to put some data in the same container (needed to keep an homogenous display if elements are not the same height)
- update of data structure to use named objects instead of generic numbers array
- add knobs to storyboook, to display/hide optional elements of footers
- cleanup code and add comments

All changes should be backwards compatible (markup and data)


papegaill (commented):
Changes are perfectly matching order defined in https://webgate.ec.europa.eu/CITnet/confluence/pages/viewpage.action?pageId=907441470 

However at markup level, we now have section 2 and 3 (with 2 sub-sections not numbered). 
We are now going from section 3 to 6. Won't it be a bit confusing in the end for integrators? 

`.ecl-footer-harmonized__section` is missing for sections 2 and 3. 

I understand the fact that we have to leave section 6 container even if it's empty in order to keep a separation border. But I find it confusing in the end, for upper sub-sections it's enough to remove container and it's content, but here container must be kept.

I wonder if it would not be clearer to have a rows structure: 

1. ecl-footer-harmonized__row1 
    - ecl-footer-harmonized__col 
    - ecl-footer-harmonized__col 
    - ecl-footer-harmonized__col 

2. ecl-footer-harmonized__row2 (with top border > `.ecl-footer-core__section--separator`) 
    - ecl-footer-harmonized__col 

3. ecl-footer-harmonized__row3 (with top border > `.ecl-footer-core__section--separator`) 
    - ecl-footer-harmonized__col 
    - ecl-footer-harmonized__col 
    - ecl-footer-harmonized__col 

Wouldn't it be clearer for end users? 
Sorry to challenges markup, but do we really need to number each section after all?
What do you think?

emeryro:
At the beginning, the footer was supposed to be a fixed structure, with all elements mandatory. So we ended up using a css grid. Also we wanted to have a similar markup for all possible footer as they were relatively close.
Now the specifications have evolved, footers are quite different, with optional elements. So we could indeed think about it again.

Your proposal seems completely doable, but the main issue here is that we have to keep backward compatibility; the old markup has to work with the new css, which can be tricky.
I will have a deeper look at it.

Thanks for the feedback!

emeryro:
So, I did some tests to see how we could integrate the proposed markup structure, and in the end I think that it will be really complex to do it without losing backward compatibility.
There are lots of specific cases concerning the spacing and border, so we would have to put a dedicated css class on each section anyway.

I don't say that it isn't possible, but that would probably required lots of time.
So I suggest to open a new ticket for this, to keep things clean, and keep the current markup for now.

papegaill (approved):
FRONT-494 created :smirk:

PR #1551, opened 2020-02-25
feat(menu): implement menu v2 - FRONT-811

emeryro:
# PR description

specs: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Menu+v2+-+Design+Specifications

several specific behavior have been added to follow the specs (mobile display on tablet regardless of screen size, swipe to open/close the menu, sticky header, ...)

Also display has been slightly adapted on IE (grateful degradation) with a forced full width mega menu


planctus:
salut @emeryro, thanks for the hard work.
I've made a simple test by changing the labels of the first level menu item to be 16 characters long (specs are ambiguous, in a page i've found 17 in another 16)
The result is that:
- The menu is not fitting in the container, which means that if everything is correct this "editorial guidelines" are not effective
- but it also means that our calculations to collapse the menu are not necessarily correct.

image

planctus:
sub-items have an editorial recommendation to be not longer than 32 characters but there is no "reaction" foreseen for such a use case, at the moment with a longer sub-item i see this happening:

image

planctus:
But it seems that the width of the list item in the dropdown is not correct, even with a 32 chars long string, the list item is "shorter" than the link.
image

papegaill:
On desktop there is a horizontal scrollbar, something is pushing. 
Scrollbar disappears if you hover some menu items.

Screenshot 2020-03-03 at 16 43 16

papegaill:
Based on specs mobile background should be full-height (100vh)

mobile-menu-bg-blue-

papegaill:
I think we need to avoid scrolling the body, when scrolling menu on mobile. It can become difficult if you close the menu and therefore get lost..

mobile-body-scrool

papegaill:
Why we have this default focus on first sub-items? Is this coming form the specs?

default-mobile-focus

papegaill:
Any off-canvas transition planned for mobile menu?

captured

papegaill:
on desktop, hovering delay between each item to open dropdown seem long?

emeryro:
Thanks for the reviews!

Now, to give some answers to the comments:
- The desktop menu collapse is done on page load (similar to what is done for the breadcrumb for instance). So indeed if you edit the label with the inspector, display will be broken, but I don't think this is a real use case. If you had this behavior on page load, that's indeed a bug. Although on my side it seems to work (see the "ecl-menu--forced-mobile" class added)
image
- I will fix the display of long sub links (I guess it is safe to put them on 2 lines in this case)
- for the horizontal scrollbar, that's probably because the calculation to see if there is enough space on the right or not is done on hover. I'll see if I can fix it.
- I'll check the mobile menu background to make it full height
- I'll check if that won't be too complex to prevent body scrolling when the menu is open
- The default focus on first item is because you clicked on this item to go to this "screen". Technically, that's the same DOM element as before, display has just been changed with css and js. If needed I can force the focus out of it with js, but I'm not sure that would be a good idea.
- There is nothing about animation for the menu in the specs. If needed we could add it, but it has to be clearly defined (duration, speed, ...)

planctus:
@emeryro , the test i made was done by editing the specs, not using the inspector. 
I also wonder about who is supposed to enforce the behaviour of this menu, if any of the label of the first level is longer than 17 characters (in english) it should go hamburger, It's this a functionality dg comm expects from ecl?
Same for "if there are more than 7 items in the first level" it should go hamburger. who is supposed to do it? I think we have a class for these use cases but not the code to set it when these conditions are met, right?

emeryro:
@planctus I'll double check this menu width issue
Concerning the check on the data, to me it is quite clear: this is not up to ECL to check this kind of things. If data have to follow specific pattern, it has to be put in the guidelines. So if display is not "optimal" on a site because of bad data use, that would be their responsibility

planctus:
@emeryro , i think here is different, these are just editorial guidelines, not editorial rules, you are not forbidden to include a longer menu item, for instance, what the specs say is that the menu would collapse, in that case. 
Same for the number of items, if you have more than seven they all get in the hamburger, is not that you violated some rules and your implementation is not valid. 
So, this is not a matter of respecting or not the guidelines, it's a behaviour that has to be created, somewhere, if ECL doesn't do it, it also means that the component cannot be correctly implemented if an "integration" is not present, which i think is wrong. The vanilla component should behave as such, and ECL is the one that defines it. 
Having said this i'm not happy at all at the idea of enforcing these behaviours in the library, but if they are "valid" it has to be possible to "reproduce" them with the ECL component only.

planctus:
@emeryro , reading the specs you find this: 

> For all tablets, regardless of their screen size, display the hamburger menu.

I haven't looked at the code, but testing the component in storybook with ipad pro 12.9 inches (which i suppose is a tablet) the menu is shown in the "desktop" version. As we know, being this based on "hover", this requirement becomes essential, if we render this menu in the desktop version on a tablet it would be practically unusable.

papegaill:
@emeryro There is an issue switching from mobile to desktop menu. If you open menu in mobile and then go on desktop format, then the menu stay the hamburger menu.

captured (1)

papegaill:
Dropdown menu for the latest item are not correctly right aligned on storybook preview. 
But it works fine on full-page preview.

Screenshot 2020-03-06 at 15 02 54

planctus:
If we add a seventeenth sub-item it goes to a fifth column, while it should be added to the first column in the fifth place, basing on the specs.
image

planctus:
It's a minor thing, but i see a "strong" border while hovering a menu item with no sub-items, not sure this is a requirement defined somewhere :
image

planctus:
Not sure we fixed the problem with the collapsing on the menu based on the fact that it it fits or not the container, @emeryro (the selected area is the ecl-container) 
image

papegaill:
> It's a minor thing, but i see a "strong" border while hovering a menu item with no sub-items:
> image

@planctus I think, it’s defined like that in the specs, but I also find it strange.

planctus:
I think there is a problem with the delay of the hovering, it's understandable that we wait a bit to check the user real intentions, but the styles are inconsistent, imho, the underline appears immediately while the "item" wrapped becomes grey after a bit. Especially on items with no children this is kind of noticeable.

papegaill:
Some strange offset spacing and highlight for the open mobile menu on IE11

Screenshot 2020-03-06 at 15 29 11

planctus:
About the switch of the direction of the dropdown (i know, this should have need highlighted already), i think what we do is to "reverse" the direction when the dropdown doesn't fit the viewport, but this means that if the concerned item is the last and it has room on the right it will open outside the "container" and i think this would very unhandy, so i guess that we should actually use the container to decide whether to open the dropdown toward the right or the left.

emeryro:
I'll have a look at your comments, but one thing that may make things clearer: 
Calculus for menu display are done on page load. It concerns display or not of hamburger and position of sub menu (right or left). So if you resize the screen, or change the display in storybook (wich is not considered as a page load), you may have inconsistent display as reported.
Now, resizing of screen have been considered edge cases in the past (for instance the breadcrumb also switch display on page load only). We may reopen the discussion if needed, but for now this is not taken into account.

planctus:
@emeryro , for what concerns my tests, as it was from the first moment, i'm changing the specs and loading the page, no window resizing on my side ;)

papegaill:
@emeryro testing the off-canvas transition:
- resizing the window(desktop to tablet) we can see closing transition of the mobile menu
- opening mobile menu, the overlay is shown before the end of menu transition, therefore we see a white area, can the overlay be full-width?

screencast-live browserstack com-2020 03 11-09_54_37

emeryro:
@papegaill for the menu being briefly displayed on resize, I notice it. But I'm not sure I will be able to do anything about it...
For the overlay, on my side it is working fine
overlay

papegaill:
> @papegaill for the menu being briefly displayed on resize, I notice it. But I'm not sure I will be able to do anything about it...
> For the overlay, on my side it is working fine
> overlay

@emeryro Overlay gap on IE11, working fine on Chrome & FF

planctus:
@emeryro , i tried to see whether the problems i highlighted in the previous reviews were solved or tackled and it seems to me that most of them are still there, beside the glitches in styles or behaviours what i tried to focus on is the overall capability of this implementation to fulfill the requirements because the usability of the component itself heavily relies on these requirement.

- I still see the desktop layout applied to the ipad pro 12.9 inches, for instance, and i see this as a major issue.

- I tried to add a seventeenth item again and i see that something has changed, but not the way it should have, maybe, i see now the item hidden by a display:none in desktop and i don't know the reason why.

- I see instead the dropdown behaviour improved, the concern i had was with the left/right opening with items already on the "right" side and it looks like the dropdown is now considering the container and not anymore the viewport, so it always open on the left side for the items on the right which is the right behaviour in my opinion.

- In overall it looks good to me, this is not a detailed review neither about the styles nor its browser compatibility (i only tested with chrome on linux).
Thanks for the hard work ;)

planctus (commented):
It's maybe worth spending a bit of time understanding whether the issues i reported are a common concern or not ;)

emeryro:
thanks @planctus for the feedback

- detection of tablet is done using a third part library in javascript (see https://github.com/ec-europa/europa-component-library/blob/feat/menu-v2-FRONT-811/src/systems/ec/implementations/vanilla/packages/ec-component-menu/ec-component-menu.js#L240). As it seems, it doesn't work perfectly on your side. I haven't a tablet available here, so I could only try with simlations (which worked), but maybe that library is not good enough
- for the 17+ items, that's the best answer I could offer for now. That's an implementation of the limitation to 16 items max displayed. It won't be possible to have them displayed below in some custom order.
- for the first level, there is no hard limit of 7 items. as long as there is enough space, the destop menu is displayed, and if that's not the case, the hamburger menu replaces it. That's what I understood from the specs

planctus:
@emeryro , you are right about the requested behaviour for the first level, the way it behaves is correct, if it finds space it places the additional item horizontally otherwise it goes hamburger, sorry for the confusion.

> Important note: If the above-mentioned limitations cannot be imposed, allow for the following expansion:
Place the additional items (e.g. Item 8, Item 9, ...) horizontally; once the menu concerts to its Hamburger display, place them vertically, under the preceding item number.

About the other points i'm glad to know that we are trying to detect a tablet as such, regardless of the viewport size, in my case it's not working (in storybook, with ipad 12.9 selected, with chrome on linux, at least) but this clearly needs more tests than mine and if the library we use is not working properly we can look for something else, but the important thing is that we are handling this request.

The additional items in the dropdown (from the seventeenth) is then the technical issue we are facing that prevent us from implementing one of the requirements. If we don't find a solution i guess we have to make it explicit and ask comm to change the "recommendation" (limit to 16 items) into a strict "requirement".

PR #1557, opened 2020-03-11
feat(ecl): provide watch mode including website presets - FRONT-934

kalinchernev:
# PR description

To test, run `yarn run` from project's root folder. In addition to the existing `start:ec`, there are a few new commands. The one to use is `start:ec:dev`. It's reusing existing setup within the website preset.

I've been trying to figure out a way to include the preset's folder within the storybook's preview webpack watch mode, but couldn't. Actually testing existing behavior of `start:ec`, there's no automatic reload on change neither.


papegaill (approved):
Works like a charm, many thanks @kalinchernev

papegaill (changes requested):
@kalinchernev maybe we should add small note on this for the documentation?

kalinchernev:
@papegaill docs added for the new command. Noticed a possible set of improvements in the docs in general and touched a few things as well, but the bigger picture needs big improvement which I left for later/other tasks.

papegaill (approved):

PR #1559, opened 2020-03-11
docs(release): improve release documentation - FRONT-940

emeryro:
# PR description

Add some more explaination to the release procedure


.github/CONTRIBUTING.md
```diff
@@ -18,13 +18,16 @@ Please fill in the description following the template.
 
 4.  Create a change log entry for the release:
 
-    - You'll need an [access token for the GitHub API](https://help.github.com/articles/creating-an-access-token-for-command-line-use/). Save it to this environment variable: `export GITHUB_AUTH="..."`
+    - You'll need an [access token for the GitHub API](https://help.github.com/articles/creating-an-access-token-for-command-line-use/). This token should have all the "repo" permission granted (no need for other permissions). Save it to this environment variable: `export GITHUB_AUTH="..."`
```
kalinchernev: I was not sure about the scope/permissions part here, so I checked what are the minimum requirements, which is best to have in terms of security.
Turns out that all repo are needed for private repos, whereas only 1 will suffice for public repos, such as the ECL one.
I think it'd be best for the token explanation note to refer to this https://github.com/lerna/lerna-changelog#github-token This way we are safe in time regardless whether the repo remain public or not, etc.

.github/CONTRIBUTING.md
```diff
@@ -34,14 +37,24 @@ Please fill in the description following the template.
...
+10. To publish the latest changes run `npm run publish` :warning: **Do not run `npm publish`!**
 
 11. The CLI will ask for a confirmation about the new package versions. Please verify them carefully before accepting.
 
 12. Once you have accepted, the script will start publishing the packages. Go to your `npm` user profile and organisation `@ecl` and double-check packages have actually been published.
 
 13. Push the locally-synched `v2` to the remote. This marks the release as complete.
 
-14. Finally, create a GitHub Release with the same text as the changelog generated at step 3. Add "useful links" section. Congratulations!
+14. Finally, create a GitHub Release from branch `v2` or `v2-dev` (they are equal at this point), with the following text:
```
kalinchernev: It's true that the branches are equal at this point, however it'd be probably better to mention v2, as it should be the release branch.

kalinchernev (commented):
Tiny suggestions

Please update confluence wiki page with the latest version of this file

kalinchernev (approved):

PR #1560, opened 2020-03-16
chore(deps): upgrade dependencies - FRONT-929

kalinchernev:
# PR description

- All dependencies are up to date to today 16/03/2020
- `Link` website component has been refactored to accept `label` similarly to ECL's Link component. This way, many ignore rules of prettier have been removed and future confusions of using `children` instead of `label` will hopefully be avoided.
- Many templates for content types in storybook/playground have been refactored to use `pageHeaderClassName` directly, because results of `classnames` are strings. Discovered while testing the switching of knobs on pages such as https://ec.europa.eu/component-library/playground/ec/?path=/story/templates-pages--call-for-proposals 
- `ecl-builder` had to be refactored to use promise-based patterns of `mkdirp` instead of continuation passing style.
- `parent.finder.removeChild(child)` has been refactored to `child.remove()` because of updated linting rules
- `size-limit` binary is not peer or direct dependency to the used big-lib preset, thus had to be added explicitly

## Notes

Unfortunately, `yarn audit` still fails, because of 2 packages:
- `acorn`: I was unable to solve by [`yarn resolutions`](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) to none of the published versions higher than peer of depth, because of missing `walk`.
- `decompress` package is not patched yet. (https://github.com/kevva/decompress/issues/71 & https://github.com/kevva/decompress/pull/73)


emeryro (approved):
thanks for the update and cleanup

PR #1565, opened 2020-04-03
fix(inpage navigation): fix long label and design update - FRONT-899

emeryro:
# PR description

up to date specs: https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Inpage+v2+-++Design+Specification

### DIFF
--- a/src/systems/ec/specs/components/inpage-navigation/demo/data.js
+++ b/src/systems/ec/specs/components/inpage-navigation/demo/data.js
@@ -7,7 +7,7 @@ module.exports = {
     },
     {
       href: '#inline-nav-2',
-      label: 'Heading 2',
+      label: 'Heading 2 with a long title going on several lines',
     },
     {
       href: '#inline-nav-3',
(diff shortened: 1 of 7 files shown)


src/systems/ec/specs/components/inpage-navigation/demo/data.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     },
     {
       href: '#inline-nav-2',
-      label: 'Heading 2',
+      label: 'Heading 2 with a long title going on several lines',
```
papegaill, line 10: good for validation but I'm not sure that we need to have this label on production, what do you think?

papegaill (commented):
Nice work, inpage look & feel is really better on mobile. 
while testing, I notice that in mobile the anchor is done in the wrong place. the title is hidden by the navigation bar. It's not specifically related to this ticket. but I think we should fix that too.
Already saw a previous ticket related to this FRONT-175, seems that @kalinchernev was not able to reproduce the issue, @emeryro do you have same issue?

emeryro:
Indeed, when clicking on the link the heading is hidden by the blue bar. But I'm not sure that it is an issue, as the heading label is displayed on this bar (and so replace the "real one" on the screen.
It can be discussed (preferably in a dedicated ticket), but I'm not sure that it would be easy to improve

papegaill (approved):
Ok, thank you @emeryro, let's discuss this during next meeting.

PR #1580, opened 2020-04-16
feat(label): add component - FRONT-1006

emeryro:
# PR description

specs: 

- https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Label+Component
- https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/2020-03+Label+-+ECL+Page+-+final

### DIFF
--- a/src/systems/ec/implementations/react/components/label/index.js
+++ b/src/systems/ec/implementations/react/components/label/index.js
@@ -1,0 +1,1 @@
+export default from './src/Label';
--- a/src/systems/ec/implementations/react/components/label/package.json
+++ b/src/systems/ec/implementations/react/components/label/package.json
@@ -1,0 +1,20 @@
+{
+  "private": true,
+  "name": "@ecl/ec-react-component-label",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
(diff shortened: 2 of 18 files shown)


papegaill (changes requested):
Everything ok to me 👍

The only remark would be the documentation images that seems more blurry than the ones provided by the designers?

Screenshot 2020-04-16 at 13 26 05

vs 

Screenshot 2020-04-16 at 13 28 25

emeryro:
I removed the shadow under images. It should be better

papegaill:
@emeryro Thank you but it's still the same, but I see the problem, you resized down images and as I have retina display these images seems more blurry.. I don't know images rules on documentation part? retina or not?

emeryro:
There are no rules for images.
Indeed, I had to resize them. If that really is an issue, I can display them at full size, but then It won't be accurate.
To me it is fine as it is now

papegaill (approved):
sure, this is the case for all the images for the doc, so this is another topic.

emeryro:
component has been validated by designers:
image

PR #1583, opened 2020-04-24
docs(ECL): 🧹 major grooming since v1 => v2 changes - FRONT-955

kalinchernev:
Many parts of the docs have been removed:
- updating docs on github: never used, no need to keep it
- eu system specifics: moving target

The rest is mostly simplified or updated to v2.


docs/conventions/components.md
```diff
@@ -0,0 +1,62 @@
...
+
+Please refer to the [dedicated conventions section regarding SCSS](./scss.md).
+
+## JavaScript
+
+Please refer to the [dedicated conventions section regarding JavaScript](./javascript.md).
+
+## Links
+
+If a component contains links, they should always lead to an internal example page instead of blank link (`#`) or external links: `../../example.html#{component_name}`
```
emeryro: that's not very clear which links we are talking about.
I guess that's for the demonstration of the component, right? If so, it may be part of a section dedicated to "specs" files

docs/README.md
```diff
@@ -1,19 +1,15 @@
...
 
-Welcome to ECL documentation landing page. Within these documentation pages, one can find the following types of information:
+ECL v2 consists of 2 systems (EC & EU) composed of more than 540 packages organized in 2 implementations within this repository - Vanilla and React. [Twig](https://github.com/ec-europa/ecl-twig) implementation is a separate project with nearly 60 packages in the 2 systems.
 
-- [Conventions](./conventions/README.md): ECL's collection of best practices for developers.
-- [ECL's design patterns](atomic-design-patterns.md): especially useful when you need to know how to name things within ECL.
+To maintain these it's important to have a good grasp of few concepts before diving in the code of the packages.
+
+- [Conventions](./conventions/README.md): collection of best practices.
+- [Core concepts](core-concepts.md): useful read for anyone interested in the fundamentals of ECL.
```
emeryro: link here is broken

emeryro (commented):
Thanks for this documentation update!
I had a quick first look at it, and spotted a few things already:
There are a few conventions that are not really followed on our side for now:
- github commit message. Should we start using these conventions? Should we display warning during commit if they are not foloowed?
- theming parmeters in scss files; some components have them, some don't. There are no themes for now on ECL, so that's not really an issue

Also I guess that the page dealing with ECL EC and EU may evolve in the future, as the shape of ECL EU is quite uncertain for now

kalinchernev:
Pity I can't reply to comments of the review, adding them here instead.

> github commit message. Should we start using these conventions? Should we display warning during commit if they are not foloowed?

I asked myself the same and came to the conclusion that although git commit conventions are not enforced through hooks and are not a blocking issue, it's still good to keep the documentation for future references as a suggestion and best practice. In real life, it will most probably referred mostly for giving an example of the `feat`, `fix`, etc. categories as they are still enforced automatically through the pull requests conventions.

> theming parmeters in scss files; some components have them, some don't. There are no themes for now on ECL, so that's not really an issue

What is your suggestion: shall we maybe remove this section to avoid confusion or rather re-phrase things in a way that this idea used to be valid in the past with an explanation but no longer, etc.?

emeryro (approved):
Seems good to me, good job!
I keep the PR open, so someone else can have a look at it

PR #1584, opened 2020-04-28
feat(legacy-page-header): add background image - FRONT-1130

emeryro:
# PR description

Add a variant to allow background image on legacy page header

demo: https://5ea932019d7a378c696b1ec2--europa-component-library.netlify.app/playground/ec/?path=/story/deprecated-page-header-ecl-2-14-0--background-image


planctus (approved):
So ok, let's go like this, in twig i will add also a boolean, since the idea is that you could only want to get the generic css for the page-header-background and then go writing your own css for your breakpoints and support for retina screens in each implementation, but here it's fine like this

PR #1590, opened 2020-05-18
feat(select): add multiple select - FRONT-1142

emeryro:
# PR description

Add custom design and behavior for multi select

Specs: https://citnet.tech.ec.europa.eu/CITnet/confluence/pages/viewpage.action?spaceKey=NEXTEUROPA&title=Selects+%28droplist%29+Multi-select+Specs


kalinchernev:
I think not all branch names are valid aliases and that's why preview is not showing actual content

planctus:
There is something i don't necessarily find a requirement for, but i guess that the width of the text input should always be 100% of the container. when the selected width is "L", the text field doesn't expand, instead.

image

planctus:
I see that the point we raised about how to handle the text inside the select when multiple selections are made was not really answered, but it's a crucial one, how can this be just ignored? 

image

planctus:
About the selected items and how they are reflected in the select box, it's also a bit surprising to have them sorted alphabetically, so that the last selected item might not be the last selection as reported in the box. But clearly this might change once the overall issue will be handled.

planctus (approved):
I left a few comments, @emeryro, but maybe only the issue with the text-input size is relevant, i leave it to your evaluation.
The other comments are about something that apparently we tried to highlight but that has been left unhandled.

emeryro:
thanks for the review @planctus 
it would indeed maked sens to apply the width to the text field too. I just fear that it would be too big as we have extra borders here. To be seen...

the behavior of the component, including the text order, has been validated by COMM as it. We could reopen some discussion, but I'd rather have a first working version and improve it later if needed

PR #1591, opened 2020-05-18
feat(select): add js file - FRONT-1142

kalinchernev:
Preview available at https://select-multiple-dev--europa-component-library.netlify.app


src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -0,0 +1,196 @@
...
+    this.selectTextDefaultDataSelector = selectTextDefaultDataSelector;
+    this.selectTextSearchDataSelector = selectTextSearchDataSelector;
+
+    // Private variables
+    this.id = 0;
+    this.input = null;
+    this.search = null;
+    this.select = null;
+    this.selectIcon = null;
+    this.textDefalt = null;
```
emeryro: small typo: defalt

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -0,0 +1,196 @@
...
+    root.ECLSelect = select;
+    return select;
+  }
+
+  constructor(
+    element,
+    {
+      selectIdSelector = 'select-multiple',
+      selectDataSelector = 'data-ecl-select-multiple',
+      selectMultipleCssSelector = 'ecl-select__multiple',
```
emeryro: this parameter seems to be a css prefix rather than a selector (we try to avoid to use css classes for selectors; data attributes are used instead when needed)
I'm not sure this parameter has to be exposed, as it won't be displayed correctly if another value is used here.

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -0,0 +1,196 @@
...
+        );
+      });
+    }
+
+    this.select.parentNode.parentNode.insertBefore(
+      this.selectMultiple,
+      this.select.parentNode.nextSibling
+    );
+
+    this.select.style.display = 'none';
```
emeryro: we should use a dedicated css class to hide the element; adding `display: none` would prevent data from being sent to the form.
Ideally this class should be added to the parent container of the select, so that I could remove extra margin too (something like `.ecl-select__container--hidden`)

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -0,0 +1,196 @@
...
+
+    this.search = document.createElement('input');
+    this.search.classList.add('ecl-text-input', 'ecl-text-input--m');
+    this.search.setAttribute('id', `select-multiple-search-${this.id}`);
+    this.search.setAttribute('type', 'text');
+    this.search.setAttribute('placeholder', this.textSearch || '');
+    this.searchContainer.append(this.search);
+
+    this.searchContainer.append(
+      createCheckbox('select-multiple-all', 'Select all')
```
emeryro: I will provide the text for select all in a data attribute `data-ecl-select-all`

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -0,0 +1,196 @@
...
+    );
+
+    if (this.select.options && this.select.options.length > 0) {
+      this.select.options.forEach(option => {
+        this.searchContainer.append(
+          createCheckbox(`select-multiple-item-${option.value}`, option.text)
```
emeryro: we may have conflict for ids if there are several multi-select on the page
I'd suggest using the id of the initial select element as prefix

emeryro (commented):
Looks good so far!
The generated markup seems to be correct, which was the main challenge here.

I noted a few next steps to be done on both side:
on css/html side:
- hide select without display none
- remove extra spacing on select
- set position absolute for overlay
- add data for select all data-ecl-select-all

on js side:
- show/hide overlay on click
- handle select all
- add value to the select element when ticking a box

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -0,0 +1,289 @@
...
+    this.search.setAttribute('placeholder', this.textSearch || '');
+    this.searchContainer.appendChild(this.search);
+
+    const selectAll = this.createCheckbox(
+      'all',
+      this.textSelectAll,
+      this.selectMultipleId
+    );
+    selectAll.addEventListener('click', this.handleClickSelectAll);
+    this.searchContainer.appendChild(selectAll);
```
emeryro: shoudn't it be selectContainer instead of searchContainer?

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -0,0 +1,289 @@
...
+
+    this.select = queryOne(this.selectMultipleSelector);
+
+    if (this.select.nextSibling.classList.contains('ecl-select__icon')) {
+      this.selectIcon = this.select.nextSibling;
+    }
+
+    this.selectMultiple = document.createElement('div');
+    this.selectMultiple.classList.add('ecl-select__multiple');
+    this.selectMultiple.addEventListener(
```
emeryro: currently, this event is triggered when clicking a checkbox, so it is not possible to select mutliple ones without reopening the overlay every time.
I guess that putting the click event on the first ecl-select__container would be better

Note: there is a small video in a comment here, which shows expected behavior: https://citnet.tech.ec.europa.eu/CITnet/confluence/pages/viewpage.action?spaceKey=NEXTEUROPA&title=Selects+%28droplist%29+Multi-select+Specs

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -0,0 +1,289 @@
...
+      }
+    });
+
+    this.updateInputValue();
+  }
+
+  /**
+   * @param {Event} e
+   */
+  handleClickSelectAll(e) {
```
emeryro, line 340: improvement for the select all: when the select all checkbox is checked, if user uncheck another checkbox, the select all checkbox should also becomes unchecked

emeryro (commented):
thanks for the updates. We are getting closer to completion
I found a few points below that could be updated/fixed.

Apart of that, I think that the remaining points are:
- try to fix errors on IE
- handle search. For this point, I may see with COMM if it could be added on a second phase (if they want to have a running version of the component earlier)

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
kalinchernev: I double-checked, the markup at the moment is matching the one from the other pull request with [reference markup](https://5ecb7eefd2193ec6ca8e002c--europa-component-library.netlify.app/playground/ec/iframe.html?id=components-forms-select--multiple) where `ecl-select__multiple-dropdown` contains search input and checkboxes which is named searchContainer in the JS implementation

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
kalinchernev: Nice suggestion, thanks! Implemented.

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -184,6 +180,28 @@ export class Select {
...
+    this.search.addEventListener('keyup', e => {
+      const keyword = e.target.value;
+      if (keyword.length > 2) {
+        this.checkboxes.forEach(checkbox => {
+          if (
+            !checkbox
+              .getAttribute('data-select-multiple-value')
+              .includes(keyword)
+          ) {
+            checkbox.style.visibility = 'hidden';
```
emeryro: here we should toggle display instead of visibility (`display: none` or `display: flex`).
As the checkboxes are not used in the form itself, it can be done, and it will improve the design

src/systems/ec/implementations/vanilla/packages/ec-component-form-select/ec-component-form-select.js
```diff
@@ -184,6 +180,28 @@ export class Select {
     this.search.classList.add('ecl-text-input', 'ecl-text-input--m');
     this.search.setAttribute('type', 'text');
     this.search.setAttribute('placeholder', this.textSearch || '');
+    this.search.addEventListener('keyup', e => {
+      const keyword = e.target.value;
+      if (keyword.length > 2) {
+        this.checkboxes.forEach(checkbox => {
+          if (
+            !checkbox
+              .getAttribute('data-select-multiple-value')
```
emeryro: as a small improvement, we may ignore font case (in the example, searching for "bel" won't keep option "Belgium" enabled)

emeryro (commented):
Seems good, thanks
I assume that the search will be discussed more in detail when needed.
I just spotted two small improvements below, with that I think we will be able to ask for a first validation

emeryro (approved):
To me, the implementation is good.
Let's wait for the validation from COMM before we merge it

kalinchernev:
Merging after confirmation in Jira

PR #1596, opened 2020-05-27
feat(gallery): improve gallery - FRONT-1168

emeryro:
# PR description

Add improvement to gallery component, as specified here: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Gallery+updates

Mainly:
- add a counter to display number of items
- limit dipslay to 3 items on mobile, and 2 lines on desktop
- add a "view all" button

### DIFF
--- a/src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
+++ b/src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
@@ -317,14 +317,14 @@ export class Gallery {
     if (!this.viewAll) return;
 
     if (this.isDesktop) {
-      const galleryY = this.element.getBoundingClientRect().y;
+      const galleryY = this.element.getBoundingClientRect().top;
       let hiddenItemIds = [];
       // We should browse the list first to mark the items to be hidden, and hide them later
       // otherwise, it will interfer with the calculus
       this.galleryItems.forEach((galleryItem, key) => {
         galleryItem.parentNode.classList.remove('ecl-gallery__item--hidden');
         if (
-          galleryItem.getBoundingClientRect().y - galleryY >
+          galleryItem.getBoundingClientRect().top - galleryY >
           this.imageHeight * 2
         ) {
           hiddenItemIds = [...hiddenItemIds, key];
--- a/src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
+++ b/src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
@@ -9,6 +9,7 @@ import { queryOne, queryAll } from '@ecl/ec-base/helpers/dom';
  * @param {String} options.descriptionSelector Selector for gallery description element
  * @param {String} options.metaSelector Selector for gallery meta info element
  * @param {String} options.closeButtonSelector Selector for close button element
+ * @param {String} options.allButtonSelector Selector for view all button element
  * @param {String} options.overlaySelector Selector for gallery overlay element
  * @param {String} options.overlayHeaderSelector Selector for gallery overlay header element
  * @param {String} options.overlayFooterSelector Selector for gallery overlay footer element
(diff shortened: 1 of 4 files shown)


planctus:
It looks good to me @emeryro, a collected just a few comments, something also unrelated to this work:

- the caption can be smaller than the icon, in the mocks they look the same when there is only one line of text, this is already the case in production.

video_icon

planctus:
- on resize the "third" line is appearing before being hidden

glitch

planctus:
After a refresh (ctrl+r) on chrome:
image

If after that i hard refresh (ctr+shift+r):
image


But this is affected by the size of the "viewport", with a smaller one (like with knobs enabled on the side), when i use ctrl+shift+r i get the two rows back.

planctus:
i get the same result of the refresh by clicking "enter" in the url bar, one single image thumbnail being shown, (i modified the default viewport to be desktop), but also an important glitch with the same image appearing like "in mobile", full width, before that.

planctus:
Nothing specific about this implementation, i guess, but the "view all" link hovering is so difficult to perceive that you have the impression the link is not "active".

planctus:
I did not know getBoundingClientRect(), i read a couple of articles and here it was recommending not to use "y" due to lack of browsers support. https://www.digitalocean.com/community/tutorials/js-getboundingclientrect

i checked on Edge and it works, this is IE11:

image

src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
```diff
@@ -408,6 +515,31 @@ export class Gallery {
     }
   }
 
+  /**
+   * Invoke listeners for on click events on view all.
+   *
+   * @param {Event} e
+   */
+  handleClickOnViewAll(e) {
```
planctus, line 523: I noticed that this handler is the same as the standard handler for the click on an element, couldn't we re-use that by simply calling it from `handleClickOnViewAll()`?

planctus (changes requested):

emeryro:
Thanks for the review.
To address every point:
- caption size: indeed, visually it does seems to be different. We didn't recieve any specification for that, so I don't know it we should do something about it. Let's wait for DG COMM's feedback
- third line appearing when resizing: that's because the resize event is only called once (for obvious performance matter). So I'd say it is expected bahavior (plus resizing is already an edge case)
- for the chorme display issue: I noticed it, indeed. But that's a wrong interaction between chrome and storybook, similar to the one we noticed on the menu. Displaying the gallery in real cases won't present issues. I'd love to fix this behavior, but as for the menu I don't think that's something we can handle
- view all hover: indeed, the difference is quite subtile. It is using the default button, we could report it but as it has been validated like this long time ago I'm not sure the designers wish to update it
- getBoundingClientRect().y: I'll check if I can find an easy way to improve this. By the way, display is really different on IE11 anyways (graceful degradation)

src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
emeryro: it isn't 100% the same. The only difference is that the view all button displays the first item in any case (`this.updateOverlay(this.galleryItems[0]);`) and the item click depends on the item selected (`this.updateOverlay(e.currentTarget);`)

planctus (dismissed):

src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.js
```diff
@@ -250,6 +290,73 @@ export class Gallery {
...
+  /**
+   * Hide several gallery items by default
+   * - 2 "lines" of items on desktop
+   * - only 3 items on mobile
+   */
+  hideItems() {
+    if (!this.viewAll) return;
+
+    if (this.isDesktop) {
+      const galleryY = this.element.getBoundingClientRect().y;
```
planctus: about this @emeryro , the article i was referring to was only warning us about the usage of the `y` in getboundingclientRect(), `top` is more broser compatible, he said, and it does the same thing.

> getBoundingClientRect is reliably supported on all desktop and mobile browsers (with the exclusion of the x and y values which is unstable in IE/Edge and Safari)

PR #1597, opened 2020-06-04
chore(deps): treat vulnerabilities - FRONT-1172

kalinchernev:


emeryro (changes requested):
Seems like there are some issues with netlify preview (tried to restart the build, but same thing)

kalinchernev:
Unfortunately, I haven't been able to downgrade by neither one-by-one, nor by group of deps in a working way. What's why I made a new one only with the required upgrades for the security here https://github.com/ec-europa/europa-component-library/pull/1601 

I tried to update the yarn.lock and combination of downgrades in order to make this existing PR work, but somehow yarn does not work well with resolutions and downgrades in a consistent ways and I had to start from a new.

PR #1598, opened 2020-06-04
feat(table): add sorttable - FRONT-1155

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


src/systems/ec/implementations/react/components/table/stories/Index.jsx
```diff
@@ -1,14 +1,38 @@
...
+      </StoryWrapper>
+    ),
+  ],
 };
 
-export const Default = Table;
+export const DefaultTable = Table;
 
-Default.story = {
+DefaultTable.story = {
```
kalinchernev: The `Default` here is on purpose, it is consistently such throughout all stories in order to keep stories' paths compatible with ecl-twig's comparison tool which scrapes such paths https://5ed9193f8dee528866f37b7e--europa-component-library.netlify.app/playground/ec/?path=/story/components-table--default which now would become https://5ed9193f8dee528866f37b7e--europa-component-library.netlify.app/playground/ec/?path=/story/components-table--default-table and break the patter.

Please revert this change.

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
```diff
@@ -0,0 +1,105 @@
+/* eslint-disable unicorn/consistent-function-scoping */
+import { queryOne, queryAll } from '@ecl/ec-base/helpers/dom';
+
+/**
+ * @param {HTMLElement} element DOM element for component instantiation and scope
+ * @param {Object} options
+ * @param {String} options.sortSelector Selector for toggling element
+ * @param {Boolean} options.attachClickListener Whether or not to bind click events on toggle
```
kalinchernev: I don't think this one should be optional. If it is off, then the behavior will be unnecessarily loaded without being able to do anything.

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
```diff
@@ -0,0 +1,105 @@
...
+    table.init();
+    // eslint-disable-next-line no-param-reassign
+    root.ECLTable = table;
+    return table;
+  }
+
+  constructor(
+    element,
+    {
+      sortSelector = '[data-ecl-tablesort-toggle]',
```
kalinchernev: I'm not sure `data-ecl-table`  https://github.com/ec-europa/europa-component-library/pull/1598/files#diff-6b789af6f8f8d406236aed4ca6aae98aR20 is necessary given this option is actually used for internal references within the class.

The `data-ecl-auto-init="Table"` referes to `root`/`element` which is enough I think.

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
```diff
@@ -0,0 +1,105 @@
...
+   */
+  init() {
+    this.toggles = queryAll(this.sortSelector, this.element);
+
+    // Bind click event on toggles
+    if (this.attachClickListener && this.toggles) {
+      this.toggles.forEach(toggle => {
+        toggle.addEventListener(
+          'click',
+          this.handleClickOnSort.bind(this, toggle)
```
kalinchernev: Please don't forget to `removeEventListener` with the same signature from `destroy()` to avoid memory leaks.

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
```diff
@@ -0,0 +1,105 @@
...
+    const comparer = (idx, asc) => (a, b) =>
+      ((v1, v2) =>
+        v1 !== '' && v2 !== '' && !Number.isNaN(+v1) && !Number.isNaN(+v2)
+          ? v1 - v2
+          : v1.toString().localeCompare(v2))(
+        (asc ? a : b).children[idx].textContent,
+        (asc ? b : a).children[idx].textContent
+      );
+
+    const table = toggle.closest('table');
```
kalinchernev: General/global polyfill not yet available in ECL centrally, so probably IE will complain :)

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
```diff
@@ -0,0 +1,105 @@
...
+    const tbody = queryOne('tbody', table);
+
+    [...queryAll('tr', tbody)]
+      .sort(
+        comparer(
+          // eslint-disable-next-line unicorn/prefer-spread
```
kalinchernev: Please remove this line when https://github.com/ec-europa/europa-component-library/pull/1591/commits/cc104fdc5e545de3e2cb91e51964098d197ef8ae goes in. These have been repeated unnecessarily imposed rules and `Array.from` is definitely better in this conversion.

kalinchernev (changes requested):
Works well, nice job :)
I've added comments on what I see as possible set of improvements, but in general it's good result in my opinion.

src/systems/ec/implementations/react/components/table/stories/Index.jsx
papegaill: Ok, thank you!

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
papegaill: ok, done, thx for explanation!

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
papegaill: done 👍

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
papegaill: done by changing the logic to retrieve the correct column index with the support of colspan

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
papegaill: I don't really understand your comment... being optional or not? I took example of the accordion and it seems same for the menu .. I think I need to more explanation here?

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
papegaill: Polyfill added 👍  why don't we put this to the helpers? 
I need to test this on IE..

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
kalinchernev: You're right the accordion(s) and the expandable have this option to optionally attach a listener or not. My comment is related this https://github.com/ec-europa/europa-component-library/pull/1598/files#diff-9c8d63a91d8421e5c83211d9d55da702R93 
Is it really that the user adding javascript to sort values in the table will want to put `false` on this flag to not be able to use the behavior? I don't see possible use cases for such options.

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
kalinchernev: There's FRONT-1187 for improving how polyfills are used indeed.

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
papegaill: Ok sure it make sense now, thx for explanation, code updated ;)

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
```diff
@@ -0,0 +1,153 @@
...
+    if (this.sortHeadings) {
+      this.sortHeadings.forEach(tr => {
+        const sort = document.createElement('span');
+        sort.classList.add('ecl-table__arrow');
+        /* eslint-disable-next-line unicorn/prefer-node-append */
+        sort.appendChild(Table.createSortIcon('ecl-table__icon-up'));
+        /* eslint-disable-next-line unicorn/prefer-node-append */
+        sort.appendChild(Table.createSortIcon('ecl-table__icon-down'));
+        /* eslint-disable-next-line unicorn/prefer-node-append */
+        tr.appendChild(sort);
```
papegaill, line 94: wanted to use `prepend` here, but it's not IE compatible, so indeed of having a new Polyfill.. 
I use `appendChid` as the element is absolute positioning.

kalinchernev: Yes, `appendChild` is better for IE compatibility in my opinion as well.
I've actually removed this eslint plugin `unicorn` here https://github.com/ec-europa/europa-component-library/pull/1591/commits/cc104fdc5e545de3e2cb91e51964098d197ef8ae because its suggestions are too frequently irrelevant to the ECL code base patterns.

kalinchernev: When you have multiple occurrences of a rule which is irrelevant for a given file, you can disable the rule as following from the top of the file

```
/* eslint-disable unicorn/prefer-node-append */
```

src/systems/ec/implementations/react/components/table/examples/TableSort.jsx
```diff
@@ -0,0 +1,55 @@
...
+            {row.map(({ label, ...cellProps }, cellIndex) => (
+              <TableHeader
+                {...adaptCellProps(cellProps)}
+                key={`header-row${rowIndex}-cell${cellIndex}`}
+                {...(label &&
+                  !cellProps.colspan && {
+                    'data-ecl-table-sort-toggle': true,
+                  })}
+              >
+                {parse(label)}
```
kalinchernev:
```suggestion
                {label}
```

src/systems/ec/implementations/react/components/table/examples/TableSort.jsx
```diff
@@ -0,0 +1,55 @@
...
+      </TableHead>
+      <TableBody>
+        {demoContentDefault.rows.map((row, rowIndex) => (
+          <TableRow key={`body-row${rowIndex}`}>
+            {row.map(({ label, ...cellProps }, cellIndex) => (
+              <TableCell
+                {...adaptCellProps(cellProps)}
+                key={`body-row${rowIndex}-cell${cellIndex}`}
+              >
+                {parse(label)}
```
kalinchernev:
```suggestion
                {label}
```

src/systems/ec/implementations/react/components/table/examples/TableSort.jsx
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable import/no-extraneous-dependencies, react/no-array-index-key  */
+import demoContentDefault from '@ecl/ec-specs-table/demo/data--multi';
+
+import React from 'react';
+import parse from 'html-react-parser';
```
kalinchernev: Not necessary as the JSX will parse strings without any issues. I noticed other components use it too, but I honestly don't see any benefit from it.

I can guess the idea here is to allow the component user to input HTML and not only string label, but I'm personally not sure this is a good idea to allow.

src/systems/ec/implementations/react/components/table/examples/TableSort.jsx
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable import/no-extraneous-dependencies, react/no-array-index-key  */
```
kalinchernev: The `react/no-array-index-key` ignore rule is available in in 26 occurrences in 24 files which is a hint the rules does not help much in the code base. Could you please remove these ignore rules and add a global one after this one please? https://github.com/ec-europa/europa-component-library/blob/v2-dev/.eslintrc.js#L17

src/systems/ec/implementations/react/components/table/stories/Index.jsx
```diff
@@ -1,9 +1,33 @@
+/* eslint-disable react/react-in-jsx-scope, import/no-extraneous-dependencies */
```
kalinchernev, line 1: The `import/no-extraneous-dependencies` is 392 occurences in 391 files. We'll need to see the [options](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md#options) and tweak this, it's crazy to carry ignore rules like this along the whole project

kalinchernev: Created https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1247 for following up on all the ignore rules updates.

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.js
```diff
@@ -0,0 +1,169 @@
...
+      el = el.parentElement;
+    }
+    return null;
+  };
+
+/**
+ * @param {HTMLElement} element DOM element for component instantiation and scope
+ * @param {Object} options
+ * @param {String} options.sortSelector Selector for toggling element
+ * @param {Boolean} options.attachClickListener Whether or not to bind click events on toggle
```
kalinchernev: Thanks for removing also the comment for this option.

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.scss
```diff
@@ -162,6 +162,47 @@
...
+    cursor: pointer;
+    padding-left: calc(
+      #{$ecl-spacing-m} + #{$ecl-icon-2-xs} + #{$ecl-spacing-xs}
+    );
+    position: relative;
+  }
+
+  .ecl-table__arrow {
+    left: $ecl-spacing-m;
+    position: absolute;
```
kalinchernev, line 177: Isn't this parent element of absolute children supposed to be relative to hold them?

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.scss
```diff
@@ -162,6 +162,47 @@
...
+  }
+
+  .ecl-table__icon-up {
+    fill: $ecl-color-grey-50;
+    left: 0;
+    position: absolute;
```
kalinchernev: These are repeated throughout both up and down elements, a candidate for `.ecl-table__icon` if you ask me

src/website/scripts/docs.js
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+/* eslint-disable no-console */
```
kalinchernev: Why? It's a warning, not blocking

src/website/src/website-components/Icon/IconCard.jsx
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable import/no-extraneous-dependencies */
+/* eslint-disable import/no-extraneous-dependencies, react/no-danger */
```
kalinchernev: Also a warning, no need to ignore the `react/no-danger`

kalinchernev (changes requested):

src/systems/ec/implementations/react/components/table/examples/TableSort.jsx
papegaill: Ok good to know 👍

src/systems/ec/implementations/react/components/table/stories/Index.jsx
papegaill: ok perfect, I'm leaving them for now.

src/website/scripts/docs.js
papegaill: ok

src/website/src/website-components/Icon/IconCard.jsx
papegaill: removed, it's my editor who complains about these.

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.scss
papegaill: done

src/systems/ec/implementations/vanilla/packages/ec-component-table/ec-component-table.scss
papegaill: parent container can be _absolute_ if needed.

src/systems/ec/implementations/react/components/table/examples/TableSort.jsx
kalinchernev: I wrongly suggested updating the root file. There are eslint config files for the react implementations which I updated to solve these boring issues

kalinchernev (approved):

PR #1600, opened 2020-06-08
fix(EU): add missing elements - FRONT-1190

emeryro:
# PR description

Add missing elements in ECL EU:
- design tokens
- eu-component-fact-figures
- eu-component-form-datepicker
- eu-component-label
- eu-component-menu
- eu-editor
- eu-preset-editor
- eu-utility-dimension
- eu-utility-print

Note: to access EU playground, go to EC playground and replace `ec` by `eu` in the url


src/systems/eu/implementations/vanilla/packages/eu-component-fact-figures/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@ecl/eu-component-fact-figures",
+  "author": "European Commission",
+  "style": "ec-component-fact-figures.scss",
```
planctus: These two are still "ec"

planctus:
In the ecl editor in storybook:
image

planctus:
in the print if i click on the css resources i don't see the same as in EC, meaning that the font size gets reduced there and not here, not sure it is supposed to be exactly the same, though.
print

planctus:
Not sure this is a problem either, but the order of the components in storybook is somehow "manually" handled in ECL? Yo usee the new components they are at the bottom, here:
image

planctus:
Are we sure it was a good idea to include the menu in the list of component we could clone as they are in EC?
I mean, it looks really EC, it has the blue background and it contains the site name in the demo due to the "madness" of the family headers that we still don't have in EU, i would maybe consider to include this in the list of things for which we expect some "specs", what do you think?

planctus:
the dist is failing, not sure how this is related to these changes:

image

planctus (changes requested):

planctus:
Ah, i haven't noticed because the website started, but i had errors while running start:website, it's a repeated one in all the components:
image

emeryro:
thanks for the review @planctus 

- eu-editor has been fixed (preset was not loaded correctly)
- print differences may come from css differences in existing components. It will be handled in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1204
- order of elements in storybook is due to the use of mixed syntax for stories. All stories in EU will be updated in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1205 so things should be fine
- for the menu, that's a fair point, it could be discussed with COMM. But anyways, we needed to have the component so if we have to change a color that would be quick, and probably done in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1203
- I can't reproduce the errors you have with the build (and drone seems to work fine too). It doesn't seems to be related to this task (lots of ec stuff). If that still happens we could investigate in another ticket

planctus:
This is not anything you set in this pr, i think, but in this file:
src/systems/eu/implementations/vanilla/packages/eu-auto-init/index.js
we `import { queryAll } from '@ecl/ec-base/helpers/dom';`

in src/systems/eu/implementations/react/components/dropdown-legacy/examples/Default.jsx
we `import demoContent from '@ecl/ec-specs-dropdown-legacy/demo/data';`

in src/systems/eu/implementations/react/components/description-list/examples/Description.jsx
we `import demoDescription from '@ecl/ec-specs-description-list/demo/data';`

emeryro:
I didn't check existing components yet, that will be handled in citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1204
But sure, that will be fixed with a full diff between components

planctus (approved):

PR #1606, opened 2020-06-15
fix(inpage-navigation): fix hidden attribute - FRONT-1170

emeryro:
# PR description

Fix a misuse of the "hidden" attribute in inpage navigation (element with hidden attribute were forced to visible with CSS).
It is now handled properly.


planctus (commented):
All good for me @emeryro , but don't we have to apply the same changes to the EU component as well?

emeryro:
@planctus I intend to do it in https://github.com/ec-europa/europa-component-library/pull/1605
the current EU inpage navigation is outdated, so it will be easier to do everything at once

planctus (approved):

PR #1609, opened 2020-06-23
feat(EU): add families components - FRONT-1203

emeryro:
# PR description

Add families related components:
- breadcrumbs (no modifications)
- page headers (no modifications)
- site headers 
- footers

It also includes a fix for the menu display in site headers, spotted during the component creation (display issue when opening the mobile menu). It could be seen on [actual implementation](https://ec.europa.eu/component-library/playground/ec/?path=/story/components-site-headers-standardised--default)

Preview link: https://eu-families-front-1203--europa-component-library.netlify.app/playground/eu/

Notes:
Some points are not updated yet in this PR, and will be handled in a separate ticket (they are not mandatory to start using the EU components):
- order of elements in storybook
- page templates
- group deprecated components


planctus:
Both logos are visible in desktop in the site-header-harmonised

image

planctus (commented):
I could only spot the issue with the logos in the site-header-harmonised, for the rest it look simply ok to me ;)

emeryro:
@planctus site header harmonised fixed

planctus:
Cool, the implementation seems absolutely ok. At the moment i cannot really foresee the way we are going to handle the footer in ecl-twig, there might be issues but at least we know that we can make any change we want, considering that those templates have never been used.

planctus (approved):

PR #1611, opened 2020-06-25
docs(media-container): update usage page - FRONT-1230 

gingapa:
Updating media container usage page


emeryro (dismissed):
Seems good to me.
We should get validation from designer once we have a direct link available

emeryro (approved):

PR #1613, opened 2020-06-30
feat(gallery): add footer link - FRONT-1252

gingapa:
Updating design specification on Gallery component


src/systems/ec/implementations/react/components/gallery/src/Gallery.jsx
```diff
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
 import Button from '@ecl/ec-react-component-button';
+import Link from '@ecl/ec-react-component-link';
+import Icon from '@ecl/ec-react-component-icon';
```
emeryro: you don't need to import the Icon here, it's already part of the Link

src/systems/ec/implementations/react/components/gallery/src/Gallery.jsx
```diff
@@ -32,6 +35,20 @@ export const Gallery = ({
 
       <Button label={viewAllLabel} variant="ghost" data-ecl-gallery-all />
 
+      <div className="ecl-gallery__footer">
+        <Link
+          href="#"
```
emeryro: the href should be a parameter too.
In fact, the icon is also optional (according to the designers), so I think that it woulbe be clearer if we used a parameter of type Link directly, instead of separated ones.
Something similar to https://github.com/ec-europa/europa-component-library/blob/v2-dev/src/systems/ec/implementations/react/components/expandable/src/Expandable.jsx#L41, but with a Link instead of a Button

src/systems/ec/implementations/react/components/gallery/src/Gallery.jsx
```diff
@@ -32,6 +35,20 @@ export const Gallery = ({
...
+          aria-label={footerLinkLabel}
```
emeryro: the aria-label should be different from the link label. So it requires a dedicated parameter (already provided in the Link component)

src/systems/ec/implementations/react/components/gallery/src/Gallery.jsx
```diff
@@ -32,6 +35,20 @@ export const Gallery = ({
...
+          label={footerLinkLabel}
+          icon={{
+            shape: 'ui--external',
+            size: 's',
+          }}
+          iconPosition="after"
+          className="ecl-link--standalone"
```
emeryro: here you should use the link variant "standalone", which will add the correct classes automatically

src/systems/ec/implementations/vanilla/packages/ec-component-gallery/ec-component-gallery.scss
```diff
@@ -106,6 +106,12 @@
     padding: $ecl-spacing-xs $ecl-spacing-s;
   }
 
+  .ecl-gallery__footer {
+    border-top: 1px solid $ecl-color-grey-10;
+    margin-top: calc(#{$ecl-spacing-xs} - 2px);
```
emeryro: we should not tweak the spacing values like this.
I guess that here the issue comes from the padding around the "view all" link, right? If there are inconsistencies like this, it should be reported to the designer (on the spec page or in the ticket)

emeryro (changes requested):
@gingapa here are a few remarks for this PR.
If that's not clear, just let me know

emeryro:
@gingapa the implementation seems fine to me.
There are just two remaining things to be done:
- add EU version
- fix indentation issue (reported by Drone). Do you use eslint and stylelint in your working environment? It should report and detect the errors automatically
image

emeryro:
@gingapa there are still lint errors reported by Drone
image

Did you install the linter on your environment? It should report and fix these issues

emeryro (approved):

PR #1618, opened 2020-07-02
feat(file-download): add thumbnail variant - FRONT-1202

emeryro:
# PR description

specs: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/File+thumbnail+variant


src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
```diff
@@ -62,42 +108,99 @@ export const FileDownload = ({
...
+                      label={item.download.label || download.label}
+                      variant="standalone"
+                      className={classnames(
+                        item.download.className,
+                        'ecl-file__translation-download'
+                      )}
+                      download
+                      {...(item.lang && { hrefLang: item.lang })}
+                    />
+                  </li>
```
papegaill, line 159: Bottom border of last item of the translation list is not 100% width:

Screenshot 2020-07-08 at 09 17 57

src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
```diff
@@ -1,44 +1,90 @@
...
+                {detailMeta && (
+                  <div className="ecl-file__detail-meta">{detailMeta}</div>
+                )}
+                <div className="ecl-file__title">{title}</div>
+                {description && (
+                  <div className="ecl-file__description">{description}</div>
+                )}
+              </div>
+
+              {image && (
```
papegaill, line 42: I think, image goes under the content too quickly (lg). Is it not possible to keep the image opposite the content longer up to sm format? 

Screenshot 2020-07-08 at 09 18 21

Seeing other variants, the same behavior with the download button, they could also stay in front of the content longer, what do you think?

Screenshot 2020-07-08 at 09 31 45

papegaill: Same for the preview on the showcase page, the download btn and the thumbnails goes below the content. Is it necessary to pass the contents below so early (lg)?

Screenshot 2020-07-08 at 09 38 10

papegaill (changes requested):

src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
emeryro: I will check with the designers.
using sm breakpoint would probably break the design at some point (if the title is long for instance), but maybe a md breakpoint could be fine.

emeryro:
After talking with the designer, the following modifications have been done:
- the image remain a small one even on mobile (no longer full width)
- breakpoint used are MD instead of LG

Fixes have been done already, specs will be updated soon

PR #1619, opened 2020-07-03
chore(deps): upgrade - FRONT-1193

kalinchernev:
# PR description

Applies minor and major upgrades to dependencies.

Highlights:
- size limit config updated to disable webpack on each instruction as otherwise webpack config got required
- eslint upgrades also winding major bumps in related projects
- changes in linting rules (more to be refactored at FRONT-1247)
- ecl-builder scripts changes: default browserlist removed as not matching root browserlist and not changing output of JS ECL lib
- ecl-builder scripts changes: also removed instructions on modules resolutions, left at defaults
- prettier: removed an instruction of config file now part of defaults
- prettier: reformatted the whole codebase to match updated rules
- copy-webpack-plugin configuration options migrated to configuration object
- webpack performance analysis module removed because unused
- resolutions from main package.json removed as current versions of these are now secure
- react helmet imports changed to selective rather than default although both are supported, the new one is suggested
- webpack process plugin removed again to make build steps less verbose. based on index for future-proof-ness

CI:
- node setup steps removed, part of ubuntu machine environment
- 2 preview links provided by drone: for branch alias (works sometimes) and deployment id (always works)
- github actions still provides a 3rd option with pattern of ecl-preview-pull-request-id


emeryro (approved):
seems good to me

PR #1622, opened 2020-07-08
fix(multi-select): check if data-ecl-select-all attribute is set - FRONT-1274

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


emeryro:
I'm a little confused here: the ticket was about fixing some potential js error, in the (unplanned) case where no label is provided for select all.
This is done in this PR, the js file.

But what about the css? I don't think that's necessary to add complexity to the markup (with an additional class), for a case that should never occurs.

So I'd keep only the extra test in js (for both EC and EU), and do not change the markup/css.

papegaill:
@emeryro I added this specific class only for the "select all" option because if that option is not displayed, then the first option will be styled with a bottom border, which is strange:

Screenshot 2020-07-17 at 15 01 18

I don't think it's that complex but if you want I can remove the selector and css part for it.

emeryro:
@papegaill ok, but then we should use a css class related to the select component, not the checkbox.
Here you added `ecl-checkbox--all`, which seems to be a variant of checkboxes (according to the name). It should be something like `ecl-select__multiple-all` instead

Also, please add the modification to EU system too

papegaill:
@emeryro thank you for your suggestion, I just updated the PR accordingly.

emeryro (approved):

PR #1647, opened 2020-08-11
fix(components): Apply small accessibility improvements - FRONT-1454 

gingapa:
Update ECL demo pages to fix small points spotted in the accessibility report.


src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
```diff
@@ -8,6 +8,7 @@ import Link from '@ecl/ec-react-component-link';
 export const FileDownload = ({
   variant,
   title,
+  ariaLabel,
```
kalinchernev, line 11: There are 2 lint violations to solve in this file

```sh
   11:3  error  'ariaLabel' is missing in props validation                          react/prop-types
  270:3  error  defaultProp "ariaLabel" has no corresponding propTypes declaration  react/default-props-match-prop-types
```

kalinchernev (changes requested):
There are a few styling violations reported by the CI: https://github.com/ec-europa/europa-component-library/pull/1647/checks?check_run_id=971670369

You can always reproduce CI reports locally with "yarn lint" (after yarn install)

Please check if you can have prettier formatter running on save in your editor and ensure you have git hooks running. Normally having a commit without the --no-verify should catch these issues.

If you prefer to not configure your editor to solve these issues automatically, you can run the following from the CLI:

yarn prettier --write "src/**/*.{md,mdx,html,json,yml,js,jsx,scss}"

kalinchernev (approved):

PR #1649, opened 2020-08-20
feat(label): add highlight variant - FRONT-1336

emeryro:
# PR description

Add new variant for label comonent (blue background and white text)
Also update corresponding usage page

Note: validated by COMM, can be merged

### DIFF
--- a/src/systems/ec/implementations/react/components/label/stories/Index.jsx
+++ b/src/systems/ec/implementations/react/components/label/stories/Index.jsx
@@ -8,6 +8,7 @@ const variant = {
   'low importance': 'low',
   'medium importance': 'medium',
   'high importance': 'high',
+  highlight: 'highlight',
 };
 
 export default {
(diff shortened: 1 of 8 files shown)


planctus:
I think this is a good example revealing the inconsistent approach we have when it's about showing "variants" for each component. Most of the times, the component shows the different variants as different stories, in this case we have a select to switch between the different variants and the story is a "default" one.
I don't think i have a strong opinion regarding what to prefer, ideally i see the variants as a single parameter switch, therefore the "select" element applied to a single story would make it more evident, while defining a different story gives the impression that there must be other differences justifying the "independent" preview, but at the same time i do see a "value" in having the variant naming the story in the storybook menu if we prepare separated stories, i just think we should be consistent, regardless of the approach we choose.

planctus:
In this case it also seems that a better naming could be found for this added variants , we already have "high importance", having "highlighted" seems a bit overlapping the other variant's meaning.

emeryro:
I agree with what you wrote.
There are indeed some cases where the variant is handled via a select list (usually it is when the differences between variants is very low).
I suggest that we take care of this consistency when dealing with new version of ECL (V3)

planctus (approved):

PR #1659, opened 2020-09-03
fix(file): add aria-label - FRONT-1573

gingapa:
Adding aria-label to file translations link, thumbnail variant.

### DIFF
--- a/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
+++ b/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
@@ -166,7 +166,7 @@ export const FileDownload = ({
                       }}
                       href={item.download.href || download.href}
                       label={item.download.label || download.label}
-                      ariaLabel={item.download.ariaLabel + ' ' + item.title}
+                      ariaLabel={`${item.download.ariaLabel} ${item.title}`}
                       variant="standalone"
                       className={classnames(
                         item.download.className,
@@ -213,7 +213,7 @@ export const FileDownload = ({
                         item.download.className,
                         'ecl-file__translation-download'
                       )}
-                      ariaLabel={item.download.ariaLabel + ' ' + item.title}
+                      ariaLabel={`${item.download.ariaLabel} ${item.title}`}
                       download
                       {...(item.lang && { hrefLang: item.lang })}
                     />
--- a/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
+++ b/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
@@ -166,7 +166,11 @@ export const FileDownload = ({
                       }}
                       href={item.download.href || download.href}
                       label={item.download.label || download.label}
-                      ariaLabel={`${item.download.ariaLabel} ${item.title}`}
+                      {...(item.download.ariaLabel
+                        ? {
+                            ariaLabel: `${item.download.ariaLabel} ${item.lang}`,
+                          }
+                        : {})}
                       variant="standalone"
                       className={classnames(
                         item.download.className,
@@ -208,12 +212,16 @@ export const FileDownload = ({
                       }}
                       href={item.download.href || download.href}
                       label={item.download.label || download.label}
+                      {...(item.download.ariaLabel
+                        ? {
+                            ariaLabel: `${item.download.ariaLabel} ${item.lang}`,
+                          }
+                        : {})}
                       variant="standalone"
                       className={classnames(
                         item.download.className,
                         'ecl-file__translation-download'
                       )}
-                      ariaLabel={`${item.download.ariaLabel} ${item.title}`}
                       download
                       {...(item.lang && { hrefLang: item.lang })}
                     />
--- a/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
+++ b/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
@@ -166,6 +166,7 @@ export const FileDownload = ({
                       }}
                       href={item.download.href || download.href}
                       label={item.download.label || download.label}
+                      aria-label={ariaLabel}
                       variant="standalone"
                       className={classnames(
                         item.download.className,
--- a/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
+++ b/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
@@ -168,7 +168,7 @@ export const FileDownload = ({
                       label={item.download.label || download.label}
                       {...(item.download.ariaLabel
                         ? {
-                            ariaLabel: `${item.download.ariaLabel} ${item.lang}`,
+                            ariaLabel: item.download.ariaLabel,
                           }
                         : {})}
                       variant="standalone"
@@ -214,7 +214,7 @@ export const FileDownload = ({
                       label={item.download.label || download.label}
                       {...(item.download.ariaLabel
                         ? {
-                            ariaLabel: `${item.download.ariaLabel} ${item.lang}`,
+                            ariaLabel: item.download.ariaLabel,
                           }
                         : {})}
                       variant="standalone"
--- a/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
+++ b/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
@@ -75,7 +75,7 @@ export const FileDownload = ({
               }}
               variant="standalone"
               className={classnames(download.className, 'ecl-file__download')}
-              aria-label={ariaLabel}
+              {...(ariaLabel && { ariaLabel })}
               download
             />
           </>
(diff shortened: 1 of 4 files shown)


src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
```diff
@@ -166,6 +166,7 @@ export const FileDownload = ({
                       }}
                       href={item.download.href || download.href}
                       label={item.download.label || download.label}
+                      aria-label={ariaLabel}
```
kalinchernev: Ticket description is not quite clear to me so i compared with the twig changes done recently.
- this `ariaLabel` is not the right aria label I think. Should be coming from the `item.download.link` I think https://github.com/ec-europa/ecl-twig/blob/ccd12cf2b964e34b4e527f9dd163402e71a128de/src/ec/packages/ec-component-file/file.story.js#L49
- if this is the right input to parameter match, this aria-label is not marked as required https://github.com/ec-europa/europa-component-library/blob/832240423e3b61102a5b28086fa4ee72ad7cdb4d/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx#L238 so it should be rendered conditionally similarly to this https://github.com/ec-europa/europa-component-library/blob/832240423e3b61102a5b28086fa4ee72ad7cdb4d/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx#L218

kalinchernev (commented):
Not able to say whether changes are correct, putting a comment based on my limited understanding of the ticket scope

src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
planctus: Just to clarify, the issue comes from an inconsistent markup for the file component variants, we have two variants including translations, in one of them "with translations" we print the aria-label for each item while in the "thumbnail" we don't, if this was fixing an accessibility issue and not just adding an "optional" element, i believe also the "thumbnail" should have the aria label set for the translations item.

This is the markup in the "with translations" variant for what concerns the "translations":

<details>
 <ul class="ecl-file__translation-list">
        <li class="ecl-file__translation-item">
          <div class="ecl-file__translation-info">
            <div class="ecl-file__translation-title" lang="bg">български</div>
            <div class="ecl-file__translation-meta">(15.7 MB - PDF)</div>
          </div><a aria-label="Download file State of the Union 2018 brochure" download="" hrefLang="bg" href="/example#bg" class="ecl-file__translation-download ecl-link ecl-link--standalone ecl-link--icon ecl-link--icon-after"><span class="ecl-link__label">Download</span>&nbsp;<svg focusable="false" aria-hidden="true" class="ecl-link__icon ecl-icon ecl-icon--fluid">
              <use xlink:href="dist/images/icons.865a18d2.svg#ui--download"></use>
            </svg></a>
        </li>
        <li class="ecl-file__translation-item">
          <div class="ecl-file__translation-info">
            <div class="ecl-file__translation-title" lang="es">español</div>
            <div class="ecl-file__translation-meta">(15.8 MB - PDF)</div>
          </div><a aria-label="Download file State of the Union 2018 brochure" download="" hrefLang="es" href="/example#es" class="ecl-file__translation-download ecl-link ecl-link--standalone ecl-link--icon ecl-link--icon-after"><span class="ecl-link__label">Download</span>&nbsp;<svg focusable="false" aria-hidden="true" class="ecl-link__icon ecl-icon ecl-icon--fluid">
              <use xlink:href="dist/images/icons.865a18d2.svg#ui--download"></use>
            </svg></a>
        </li>
        <li class="ecl-file__translation-item">
          <div class="ecl-file__translation-info">
            <div class="ecl-file__translation-title" lang="fr">français</div>
            <div class="ecl-file__translation-meta">(15.4 MB - PDF)</div>
          </div><a aria-label="Download file State of the Union 2018 brochure" download="" hrefLang="fr" href="/example#fr" class="ecl-file__translation-download ecl-link ecl-link--standalone ecl-link--icon ecl-link--icon-after"><span class="ecl-link__label">Download</span>&nbsp;<svg focusable="false" aria-hidden="true" class="ecl-link__icon ecl-icon ecl-icon--fluid">
              <use xlink:href="dist/images/icons.865a18d2.svg#ui--download"></use>
            </svg></a>
        </li>
        <li class="ecl-file__translation-item ecl-file__translation-description">Looking for another language which is not on the list? Find out why.</li>
      </ul>
</details>

This is the markup in the "thumbnail" variant:

<details>
<ul class="ecl-file__translation-list">
        <li class="ecl-file__translation-item">
          <div class="ecl-file__translation-detail">
            <div class="ecl-file__translation-title" lang="bg">Title bg</div>
            <div class="ecl-file__translation-description" lang="bg">Morbi fringilla turpis augue, et interdum ipsum egestas sed. Proin tristique, ante id aliquet malesuada, lorem dolor vulputate magna, a commodo purus ante nec massa.</div>
          </div>
          <div class="ecl-file__translation-info">
            <div class="ecl-file__translation-language">български</div>
            <div class="ecl-file__translation-meta">(15.7 MB - PDF)</div>
          </div><a download="" hrefLang="bg" href="/example#bg" class="ecl-file__translation-download ecl-link ecl-link--standalone ecl-link--icon ecl-link--icon-after"><span class="ecl-link__label">Download</span>&nbsp;<svg focusable="false" aria-hidden="true" class="ecl-link__icon ecl-icon ecl-icon--fluid">
              <use xlink:href="dist/images/icons.865a18d2.svg#ui--download"></use>
            </svg></a>
        </li>
        <li class="ecl-file__translation-item">
          <div class="ecl-file__translation-detail">
            <div class="ecl-file__translation-title" lang="es">Title es</div>
            <div class="ecl-file__translation-description" lang="es">Proin sagittis nisi hendrerit purus porta, at suscipit est hendrerit. Duis facilisis augue imperdiet, pharetra nisl sed, molestie nulla.</div>
          </div>
          <div class="ecl-file__translation-info">
            <div class="ecl-file__translation-language">español</div>
            <div class="ecl-file__translation-meta">(15.8 MB - PDF)</div>
          </div><a download="" hrefLang="es" href="/example#es" class="ecl-file__translation-download ecl-link ecl-link--standalone ecl-link--icon ecl-link--icon-after"><span class="ecl-link__label">Download</span>&nbsp;<svg focusable="false" aria-hidden="true" class="ecl-link__icon ecl-icon ecl-icon--fluid">
              <use xlink:href="dist/images/icons.865a18d2.svg#ui--download"></use>
            </svg></a>
        </li>
        <li class="ecl-file__translation-item">
          <div class="ecl-file__translation-detail">
            <div class="ecl-file__translation-title" lang="fr">Title fr</div>
            <div class="ecl-file__translation-description" lang="fr">Duis eget lacinia arcu. Nullam mattis ornare nibh. Proin tristique, ante id aliquet malesuada. Pellentesque porttitor commodo libero sed fringilla. Curabitur varius sodales elit, id tincidunt erat. Aenean tincidunt luctus molestie.</div>
          </div>
          <div class="ecl-file__translation-info">
            <div class="ecl-file__translation-language">français</div>
            <div class="ecl-file__translation-meta">(15.4 MB - PDF)</div>
          </div><a download="" hrefLang="fr" href="/example#fr" class="ecl-file__translation-download ecl-link ecl-link--standalone ecl-link--icon ecl-link--icon-after"><span class="ecl-link__label">Download</span>&nbsp;<svg focusable="false" aria-hidden="true" class="ecl-link__icon ecl-icon ecl-icon--fluid">
              <use xlink:href="dist/images/icons.865a18d2.svg#ui--download"></use>
            </svg></a>
        </li>
        <li class="ecl-file__translation-item ecl-file__translation-description">Looking for another language which is not on the list? Find out why.</li>
      </ul>
</details>

planctus: I think that the right thing to do is to align the two "examples", but starting from the specs defined. This is an attribute set for each of the items in a list, it;s pretty lazy to define this  only once and reuse it in the element in the translation list.
The property should be set in the specs for each of the items independently as @kalinchernev was pointing out, in the `item.download.link` object defining the translations item.
This is valid for both the variants showing translations, in twig we adapt what we receive already, after this fix will be applied on ECL we will have to adapt this to use the new specs properly.

planctus (changes requested):
As you can read in my comment, @gingapa, we probably have a better way to set this parameter in a list of items, both for the "with translations" and "thumbnail" variant, is a matter of adapting the specs file, then we go rendering the "right" aria-label instead of the one used for the main link.

planctus:
Ok, @gingapa, think we are close, just that the way you set the ariaLabel can be a bit simplified, i believe:

```
 {...(item.download.ariaLabel && {
   ariaLabel: item.download.ariaLabel,
 })}
```
should do the job, but similarly i would do the same for the main download button, the situation is the same, an "half" mandatory parameter, so apply the same check there, ok?

Another small thing is, to give these values a "meaning", would you put in the specs a different aria-label for the different items? otherwise it looks like we are reusing the same over and over, you can use the pattern we "thought" about, with the language "code" in place, so something like "download translation {language code}"...what do you think?

gingapa:
Absolutely yes, good tips, thanks!

planctus (approved):

PR #1660, opened 2020-09-03
fix(page-headers): adding missing aria-label - FRONT-1575

gingapa:
Adding missing aria-label on breadcrumb elipsis for all Page Headers.

### DIFF
--- a/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
+++ b/src/systems/ec/implementations/react/components/file/src/FileDownload.jsx
@@ -75,7 +75,7 @@ export const FileDownload = ({
               }}
               variant="standalone"
               className={classnames(download.className, 'ecl-file__download')}
-              aria-label={ariaLabel}
+              {...(ariaLabel && { ariaLabel })}
               download
             />
           </>
@@ -98,7 +98,7 @@ export const FileDownload = ({
               }}
               variant="standalone"
               className={classnames(download.className, 'ecl-file__download')}
-              aria-label={ariaLabel}
+              {...(ariaLabel && { ariaLabel })}
               download
             />
           </>
@@ -166,6 +166,9 @@ export const FileDownload = ({
                       }}
                       href={item.download.href || download.href}
                       label={item.download.label || download.label}
+                      {...(item.download.ariaLabel && {
+                        ariaLabel: item.download.ariaLabel,
+                      })}
                       variant="standalone"
                       className={classnames(
                         item.download.className,
@@ -207,12 +210,14 @@ export const FileDownload = ({
                       }}
                       href={item.download.href || download.href}
                       label={item.download.label || download.label}
+                      {...(item.download.ariaLabel && {
+                        ariaLabel: item.download.ariaLabel,
+                      })}
                       variant="standalone"
                       className={classnames(
                         item.download.className,
                         'ecl-file__translation-download'
                       )}
-                      aria-label={ariaLabel}
                       download
                       {...(item.lang && { hrefLang: item.lang })}
                     />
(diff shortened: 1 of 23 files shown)


kalinchernev (changes requested):
Please do not hard-code strings into templates. Values should always come dynamically through context which can support multi-language features in the result.

The way you can pass context data into the react templates is through storybook's examples (stories).
As the breadcrumb is a child component used in the page header, you can already make a search and see examples https://github.com/ec-europa/europa-component-library/search?p=2&q=ellipsisLabel&unscoped_q=ellipsisLabel

If you need to update default data structure which is passed along as context data, you'd need to update specs https://github.com/ec-europa/europa-component-library/blob/v2-dev/src/systems/ec/specs/components/breadcrumb-core/demo/data--simple.js or https://github.com/ec-europa/europa-component-library/blob/v2-dev/src/systems/ec/specs/components/breadcrumb-core/demo/data.js. This is not the case because the ellipsis is optional.

Please move the string to their corresponding "stories". 
Example:

```
src/systems/ec/implementations/react/components/page-header-core/src/PageHeaderCore.jsx
```
will have it's value coming from

```
src/systems/ec/implementations/react/components/page-header-core/stories/Index.jsx
```

You will need to pass the property value to the breadcrumb component similarly to one of the links shared above

```jsx
  <BreadcrumbCore
    {...breadcrumbProps}
    ellipsisLabel="Click here to expand"
    data-ecl-auto-init="BreadcrumbCore"
  >
    {items.map((item) => (
      <BreadcrumbCoreItem {...item} key={item.label} />
    ))}
  </BreadcrumbCore>
```

planctus:
@gingapa, didn't look carefully at this, but it looks like also in this case this value for the ellipsis button could and should come directly from the specs, if you set it this way, for instance, the info will not be available in twig, where we fetch the spec package for the component.

planctus:
@gingapa, i think my earlier comment is still valid, you've tried to add the property in the page-header specs and you noticed that it would not have much sense, but this is actually a breadcrumb property and should come from those specs.
So, as we import the specs for the breadcrumb in the page header the "elllipsisLabel" would be directly coming from there and you would not have to do anything in the page-headers.
Mind the fact that we have to deal also with EU, so normally the operation we perform on the EC component and specs should also be reproduced for the EU system.

@kalinchernev, i'm reading only now that you already presented this option to paulo, not sure the reason why you also said it would not be the right thing to do but sorry to have "recommended it" in this comment, we can agree about how to tackle this and then we give one single "suggestion", sorry for the confusion, @gingapa ;)

planctus:
@gingapa, we said to go this way, without changing the specs, but there is an issue:

- in the website the aria-label is still empty, the example below is taken from the page-header-core "showcase"
- as mentioned already this has to be "replicated" for the EU components as well.

```
        <li class="ecl-breadcrumb-core__segment ecl-breadcrumb-core__segment--ellipsis" aria-hidden="false"
          data-ecl-breadcrumb-core-ellipsis="true"><button type="button" class="ecl-breadcrumb-core__ellipsis"
            aria-label="" data-ecl-breadcrumb-core-ellipsis-button="true">…</button><svg focusable="false"
            aria-hidden="true" role="presentation"
            class="ecl-breadcrumb-core__icon ecl-icon ecl-icon--2xs ecl-icon--rotate-90">
            <use xlink:href="/dist/media/icons.5e3c0943.svg#ui--corner-arrow"></use>
          </svg></li>
```

planctus:
@gingapa , the example in the page-header-standardised "showcase" still shows the empty aria-label...

planctus (approved):
Thanks @gingapa, i've checked also the fixes you added for the site-headers in EU ;)

PR #1661, opened 2020-09-09
fix(file): fix translation detail on thumbnail variant - FRONT-1587 

gingapa:
Fixing issue with translation detail for short descriptions on File thumbnail variant


planctus (commented):
The fix is working, i think we could also change one of the descriptions coming from the specs for it to be short enough to "prove" that it works now regardless of the length of the description.
Then, as usual, there is the alignment with EU to be maintained, @gingapa, so the same changes have to be applied in the two systems, if you change the specs also in that case you will have to "mirror" the changes in the EU spec package.

planctus (approved):

PR #1662, opened 2020-09-10
feat(select): Adding support for disabled options and default value for the select - FRONT-1588

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/systems/ec/implementations/react/components/select/src/Select.jsx
+++ b/src/systems/ec/implementations/react/components/select/src/Select.jsx
@@ -82,7 +82,7 @@ const Select = ({
           className={classNames}
           disabled={disabled}
           required={required}
-          defaultValue
+          defaultValue={defaultValue}
           {...(multiple
             ? {
                 multiple,
@@ -93,21 +93,15 @@ const Select = ({
               }
             : {})}
         >
-          {options.map((option) => {
-            const attributes = {};
-            if (option.attributes) {
-              option.attributes.map((attr) => {
-                attributes[attr.name] = [];
-                attributes[attr.name] = attr.value ? attr.value : true;
-                return attributes;
-              });
-            }
-            return (
-              <option key={option.label} value={option.value} {...attributes}>
-                {option.label}
-              </option>
-            );
-          })}
+          {options.map((option) => (
+            <option
+              key={option.label}
+              value={option.value}
+              disabled={option.isDisabled}
+            >
+              {option.label}
+            </option>
+          ))}
         </select>
         <div className="ecl-select__icon">
           <Icon
@@ -123,7 +117,7 @@ const Select = ({
 };
 
 Select.propTypes = {
-  defaultValue: PropTypes.string,
+  defaultValue: PropTypes.oneOfType([PropTypes.array, PropTypes.string]),
   disabled: PropTypes.bool,
   groupClassName: PropTypes.string,
   helperText: PropTypes.node,
--- a/src/systems/ec/implementations/react/components/select/src/Select.jsx
+++ b/src/systems/ec/implementations/react/components/select/src/Select.jsx
@@ -6,6 +6,7 @@ import classnames from 'classnames';
 import Icon from '@ecl/ec-react-component-icon';
 
 const Select = ({
+  defaultValue,
   disabled,
   groupClassName,
   helperText,
@@ -81,6 +82,7 @@ const Select = ({
           className={classNames}
           disabled={disabled}
           required={required}
+          defaultValue
           {...(multiple
             ? {
                 multiple,
@@ -91,11 +93,21 @@ const Select = ({
               }
             : {})}
         >
-          {options.map((option) => (
-            <option key={option.label} value={option.value}>
-              {option.label}
-            </option>
-          ))}
+          {options.map((option) => {
+            const attributes = {};
+            if (option.attributes) {
+              option.attributes.map((attr) => {
+                attributes[attr.name] = [];
+                attributes[attr.name] = attr.value ? attr.value : true;
+                return attributes;
+              });
+            }
+            return (
+              <option key={option.label} value={option.value} {...attributes}>
+                {option.label}
+              </option>
+            );
+          })}
         </select>
         <div className="ecl-select__icon">
           <Icon
@@ -111,6 +123,7 @@ const Select = ({
 };
 
 Select.propTypes = {
+  defaultValue: PropTypes.string,
   disabled: PropTypes.bool,
   groupClassName: PropTypes.string,
   helperText: PropTypes.node,
@@ -127,6 +140,7 @@ Select.propTypes = {
     PropTypes.shape({
       value: PropTypes.string,
       label: PropTypes.string,
+      attributes: PropTypes.array,
     })
   ),
   width: PropTypes.string,
@@ -138,6 +152,7 @@ Select.propTypes = {
 };
 
 Select.defaultProps = {
+  defaultValue: '',
   disabled: false,
   groupClassName: '',
   helperText: '',
--- a/src/systems/ec/specs/components/select/demo/data--single.js
+++ b/src/systems/ec/specs/components/select/demo/data--single.js
@@ -3,6 +3,7 @@ module.exports = {
   helperText: "This is the input's helper text.",
   invalidText: 'This is the error message',
   requiredText: '*',
+  defaultValue: '2',
   optionalText: ' (optional)',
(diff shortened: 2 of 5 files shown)


src/systems/ec/implementations/react/components/select/src/Select.jsx
```diff
@@ -127,6 +134,7 @@ Select.propTypes = {
     PropTypes.shape({
       value: PropTypes.string,
       label: PropTypes.string,
+      attributes: PropTypes.array,
```
kalinchernev, line 137:
```suggestion
      isDisabled: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
```
Basing on the previously added `disabled={option.isDisabled}`

src/systems/eu/implementations/react/components/select/src/Select.jsx
```diff
@@ -111,6 +117,7 @@ const Select = ({
 };
 
 Select.propTypes = {
+  defaultValue: PropTypes.oneOfType([PropTypes.array, PropTypes.string]),
```
kalinchernev, line 120: I tried the following for array and none of them worked:

```jsx
defaultValue={['2', '3']}
```
```jsx
defaultValue={[2, 3]}
```

This worked well with setting France by default
```jsx
defaultValue="2"
```

I briefly checked [React docs](https://reactjs.org/docs/forms.html#the-select-tag) and found that when the select is multiple, you can set multiple values with `value` attribute which worked in our storybook example

```jsx
    <Select
      {...demoContentMultiple}
      value={['1', '2', '3']}
      id="select-multiple"
      options={demoContentMultiple.options}
      label={text('Label', demoContentMultiple.label)}
      helperText={text('Helper text', demoContentMultiple.helperText)}
      invalid={isInvalid}
      invalidText={text('Invalid text', demoContentMultiple.invalidText)}
      disabled={isDisabled}
      required={isRequired}
      requiredText={text('Required text', demoContentMultiple.requiredText)}
      optionalText={text('Optional text', demoContentMultiple.optionalText)}
      width={width}
      multiple
      data-ecl-select-default={defaultText}
      data-ecl-select-search={searchText}
      data-ecl-select-all={selectAllText}
    />
```

However, options are not containing any info regarding selected items 

screenshot-nimbus-capture-2020 09 11-07_37_05

src/systems/eu/implementations/react/components/select/src/Select.jsx
```diff
@@ -6,6 +6,7 @@ import classnames from 'classnames';
 import Icon from '@ecl/eu-react-component-icon';
 
 const Select = ({
+  defaultValue,
```
kalinchernev, line 9: Not necessary, native prop `value` which is closer to non-react implementation will work as well https://reactjs.org/docs/forms.html

Try setting `value="4"` on single select in the storybook, it works. `props` are spread over, no need to add anything to achieve this.

kalinchernev: Btw, our implementation can be considered as uncontrolled because we don't do state management.

src/systems/eu/implementations/react/components/select/src/Select.jsx
```diff
@@ -92,7 +94,11 @@ const Select = ({
             : {})}
         >
           {options.map((option) => (
-            <option key={option.label} value={option.value}>
+            <option
+              key={option.label}
+              value={option.value}
+              disabled={option.isDisabled}
```
kalinchernev, line 100: In twig we are not, but with react we are able to spread over additional properties coming from the option.
```
{...option}
```
I personally see a benefit of it. Having key, value, disabled is enough. We'll need to handle "selected" as well.

src/systems/eu/implementations/react/components/select/src/Select.jsx
```diff
@@ -127,6 +134,7 @@ Select.propTypes = {
     PropTypes.shape({
       value: PropTypes.string,
       label: PropTypes.string,
+      attributes: PropTypes.array,
```
kalinchernev, line 137: Please remove.

src/systems/eu/implementations/react/components/select/src/Select.jsx
```diff
@@ -138,6 +146,7 @@ Select.propTypes = {
 };
 
 Select.defaultProps = {
+  defaultValue: '',
```
kalinchernev, line 149: Not needed.

src/systems/eu/specs/components/select/demo/data--single.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
     {
       value: '6',
       label: 'Italy',
+      isDisabled: true,
```
kalinchernev, line 31: Can we handle `disabled` only as well?

src/systems/eu/implementations/react/components/select/src/Select.jsx
```diff
@@ -81,6 +82,7 @@ const Select = ({
           className={classNames}
           disabled={disabled}
           required={required}
+          defaultValue={defaultValue}
```
kalinchernev, line 85: Not needed.

src/systems/ec/specs/components/select/demo/data--single.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
     {
       value: '6',
       label: 'Italy',
+      isDisabled: true,
```
kalinchernev, line 31: Can we handle `disabled` only as well? If we go for spread of options as described below, passing any attribute key:value pair will be put directly.

src/systems/ec/implementations/react/components/select/src/Select.jsx
```diff
@@ -138,6 +146,7 @@ Select.propTypes = {
 };
 
 Select.defaultProps = {
+  defaultValue: '',
```
kalinchernev, line 149: Not needed.

src/systems/ec/implementations/react/components/select/src/Select.jsx
kalinchernev: Not needed.

src/systems/ec/implementations/react/components/select/src/Select.jsx
```diff
@@ -111,6 +117,7 @@ const Select = ({
 };
 
 Select.propTypes = {
+  defaultValue: PropTypes.oneOfType([PropTypes.array, PropTypes.string]),
```
kalinchernev, line 120: Not necessary.

src/systems/ec/implementations/react/components/select/src/Select.jsx
```diff
@@ -92,7 +94,11 @@ const Select = ({
             : {})}
         >
           {options.map((option) => (
-            <option key={option.label} value={option.value}>
+            <option
+              key={option.label}
+              value={option.value}
+              disabled={option.isDisabled}
```
kalinchernev, line 100:
```suggestion
              {...option}
```

src/systems/ec/implementations/react/components/select/src/Select.jsx
```diff
@@ -81,6 +82,7 @@ const Select = ({
           className={classNames}
           disabled={disabled}
           required={required}
+          defaultValue={defaultValue}
```
kalinchernev, line 85: Not necssary.

src/systems/ec/implementations/react/components/select/src/Select.jsx
```diff
@@ -6,6 +6,7 @@ import classnames from 'classnames';
 import Icon from '@ecl/ec-react-component-icon';
 
 const Select = ({
+  defaultValue,
```
kalinchernev, line 9: Not necessary.

kalinchernev (changes requested):
In the case of select multiple

```jsx
          {...(multiple
            ? {
                multiple,
                'data-ecl-select-multiple': true,
                'data-ecl-select-default': multiplePlaceholder,
                'data-ecl-select-search': multipleSearchText,
                'data-ecl-select-all': multipleAllText,
              }
            : {})}
```

please do update this part of the code as well

```jsx
          {options.map((option) => (
            <option
              key={option.label}
              value={option.value}
              disabled={option.isDisabled}
              {...option}
            >
              {option.label}
            </option>
          ))}
```

For setting `selected` on items set from `values`. The way select multiple javascript part works basing on the state of the DOM will not be able to figure this information otherwise.

PR #1666, opened 2020-09-14
fix(select): add disabled and selected parameters for options

emeryro:
# PR description

Allows to add "selected" and "disabled" paramters for individual options in select lists.

Note: even if the parameter is correctly added, the behavior does not correspond yet for multi select. It will be handled in a different PR

Note2: "selected" option is not visible in browsers's inspector (seems to be the standard behavior), but it works as expected


planctus:
@emeryro, if we use "selected" on the options we get this:
`warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.`

emeryro:
@planctus where do you see this warning? I have a "selected" paramter set on the examples, and I don't see anything in the inspector console

planctus:
image

I see it only locally, i had to deal with the same when i started on this, not sure we have to bother, these are not real components, anyway..

emeryro:
ok. Well I suggest that we see with Kalin for confirmation, but to me if it happens only locally on components used only on ECL website that should indeed not be an issue

planctus:
I agree :)

kalinchernev (approved):
Changes are ok for me
There are other type suggestions coming from react which are not followed so strictly, no need to do it at this stage.

PR #1667, opened 2020-09-14
fix(menu): use css class instead of data-attribute - FRONT-1563

emeryro:
# PR description

In menu component, a css rule was set based on the presence or not of a data-attribute.
Css and markup have been updated to use a dedicated class instead of this attribute.

The use of the data-attribute has been marked as deprecated, but it will still work to maintain backward compatibility.


gingapa (approved):
Everything's looking good!

PR #1676, opened 2020-09-18
chore(ecl): start code deletion - FRONT-1631

kalinchernev:


emeryro:
After discussion, it would be better to start with only one component, as example (button), and remove the other ones, as it was originally planned

kalinchernev:
Icon is the only one left because it's a dep of the button. All the other compositions, templates, etc have been removed from vanilla, react, specs and website.

emeryro (approved):

PR #1677, opened 2020-09-21
feat(page-header): add background image variant - FRONT-1333

emeryro:
# PR description

specs: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/Core+Page+header+-+Design+specifications

note: already validated by designers


src/systems/ec/specs/components/page-header-core/demo/data--background-image.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  title: 'Page title',
+  description:
+    'Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium',
+  meta: '<span class="ecl-u-type-uppercase">Meta info</span>',
```
planctus, line 5: I don't think we should ask users to pass an html string in general, particularly speaking if the goal is to simply set a style on an element. 
I see this approach has been already used for all these set of families and it should be changed for them all, in my opinion.
Also In twig this needs a fix, at the moment we strip out the html tag and we set the "utility" class in the template, which is wrong, components should be free of any utility, but this could be already fixed simply using a twig filter.
Not sure we want to take the chance of this implementation here for applying these changes to all the page-headers, i would not be against it, otherwise if you agree we can create a "follow up" ticket for taking care of them all.

src/systems/ec/implementations/vanilla/packages/ec-component-page-header-core/ec-component-page-header-core.scss
```diff
@@ -51,6 +51,31 @@
...
+
+  /*
+   * background image variant
+   */
+  .ecl-page-header-core--image {
+    background-position: 0 0;
+    background-size: cover;
+    position: relative;
+
+    &::before {
```
planctus, line 63: Wouldn't it be easier setting the gradient together with the image in the template instead of adding it through a pseudo selector..?
The line in the component would just look like this: 
backgroundImage: `linear-gradient(90deg,#004494, rgba(0,68,148,.7)), url(${image})`,

planctus (commented):
Just a couple of questions... :)

src/systems/ec/specs/components/page-header-core/demo/data--background-image.js
emeryro: I took the same logic as in the other examples.
But indeed, that may be updated. We would require clear specifications for this (first item always uppercase? other rules?)
Let's do it in a follow-up ticket

src/systems/ec/implementations/vanilla/packages/ec-component-page-header-core/ec-component-page-header-core.scss
emeryro: My idea is to put as little inline styling as possible, as this is a bad practice.
So I'd rather keep it as is (plus it is consistent with other gradient in banner, galleries, etc.)

PR #1678, opened 2020-09-21
fix(select): allows several multiple select lists - FRONT-1579

emeryro:
# PR description

Fixes a bug which prevented two or more multiple select lists to be displayed on the same screen
Examples have been updated too, to add missing auto init attribute

To test: use the provided html file. Just make sure to copy the css and js from the dist folder to the same place as the html file
[select.zip](https://github.com/ec-europa/europa-component-library/files/5255298/select.zip)

### DIFF
--- a/src/systems/ec/implementations/react/components/select/stories/Index.jsx
+++ b/src/systems/ec/implementations/react/components/select/stories/Index.jsx
@@ -109,6 +109,7 @@ export const Multiple = () => {
       data-ecl-select-default={defaultText}
       data-ecl-select-search={searchText}
       data-ecl-select-all={selectAllText}
+      data-ecl-auto-init="Select"
     />
   );
 };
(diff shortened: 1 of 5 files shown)


planctus:
Cool, i don't see drawbacks, or side effect of this. We both came to the same conclusion and i've been testing this also in ecl-twig with ecl-ec-preset-legacy-website.js as we get it in this branch. Good to go, i would only add an example in the search page, src/ec/implementations/react/templates/pages/src/searchPage.jsx, where we just added an example of a multiple select to "prove" a css fix we made, so we also have an example showing that this works. We could add a "topic" select, like this:

```
            <Select
              id="select-multiple-2"
              label="Topic"
              groupClassName="ecl-u-mt-m"
              invalidText="No results"
              width="m"
              multiple
              className="ecl-select"
              options={[
                {
                  value: '1',
                  label: 'Any topic',
                },
                {
                  value: '2',
                  label: 'Agriculture',
                },
                {
                  value: '3',
                  label: 'Climate action',
                },
                {
                  value: '4',
                  label: 'Environment',
                },
              ]}
              data-ecl-auto-init="Select"
              multiplePlaceholder="Placeholder text"
              multipleSearchText="Search"
              multipleAllText="Select all"
              data-ecl-select-multiple="true"
            />
```

planctus (dismissed):

PR #1681, opened 2020-09-22
fix(inPageNavigation): menu overlaping headers - FRONT-1248

gingapa:
Fixing navigation overlapping headers and adding "ecl-inpage-navigation__heading" class to target them.


kalinchernev (dismissed):
The fix looks good to me.
@emeryro do you think we need to sync with design team on the duplication of headings? After the fix, a heading is displayed twice: in the inpage navigation and the actual one.

kalinchernev:
On a separate note, I think we need to either add a note in the component documentation about the new configuration we use or add details in the release. It's not a breaking change, though eUI team will also need to update config of Gumshoe.

emeryro:
The behavior seems good to me.
But I have a question: do we need this 'ecl-inpage-navigation__heading' class at all? It seems that we can handle it with only the active class, and so users won't have to add any class on their side.

Also, yes we will have to report this change of gumshoe setting. I guess that writting it in the PR description wuold be enough

kalinchernev:
I think "ecl-inpage-navigation__heading'" has been added because there are no other bem specifiers on the headings. The others are utilities and generic elements which are not going to comply with our bem rules.
And also the active state from gumshoe is only .active which also did not comply with our own bem rules
Paulo, please correct me if i'm wrong

emeryro:
yes, I'm ok with the naming. But my point is: currently we ask the user to add a class to h2 tags: https://github.com/ec-europa/europa-component-library/pull/1681/files#diff-24e2e0d304c07adc21bf356b03a7a67fR128
if we put this css code into the ecl-inpage-navigation__heading--active class directly, it will be handle automatically.
In this case, that may not be 100% BEM complient, I agree, but I don't think that's a big issue.
Another possiblity could be to handle it at the utility level. Something like adding 'ecl-u-type-heading-2--active' (and putting the css in the utility. That way it would be BEM ok (but not very clean...)

kalinchernev:
@emeryro I agree that if it's possible to style the `ecl-inpage-navigation__heading--active` class directly it will bring more value than asking the user to modify the markup in addition. Didn't think about that.
@gingapa would it be possible to achieve this?

gingapa:
> @emeryro I agree that if it's possible to style the `ecl-inpage-navigation__heading--active` class directly it will bring more value than asking the user to modify the markup in addition. Didn't think about that.
> @gingapa would it be possible to achieve this?

I believe so, please check last commit. Thanks!

kalinchernev (approved):

PR #1685, opened 2020-09-25
chore: delete eu duplicate code - FRONT-1656

kalinchernev:
Documentation has been reset. FRONT-1691 has been created if eventually any of it would make sense in the future.


emeryro:
Locally it works fine. Drone complains about [missing paths](https://drone.fpfis.eu/ec-europa/europa-component-library/7130/14), but i don't know if that is something to be fixed now or if it will be part on another task.

Also I added a comment on [FRONT-1620](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1620) (not sure where to put it) concerning element order in css

emeryro (approved):

kalinchernev:
@emeryro thanks for the review
You're right, order of styles does matter. I updated the code to fix this issue as it's better to have it straight from the start.
I'm not sure whether there will be any way to go around this when it comes to CSS as it's always a matter of specificity regardless of the way we write it. Having the icon after button is wrong specificity in this case.
Regarding the CI, drone is breaking expectedly because i've removed the examples folder. In fact, I've updated the github actions accordingly. Because of the bureaucracy around drone, it's something to work on after everything else on the v3 track and i even don't plan to take it into account for v3-next branch pull requests. Rather, we can update the github actions to our help rather than spending hours and days updating a file which will surely change in the end.

PR #1688, opened 2020-09-28
Remove design tokens

kalinchernev:
Following up on FRONT-1630. Instead of putting lengthily information about how theo works and whether it's worth keeping it or not, I've made the following set of changes removing. This can be considered as the best way to demonstrate what the project would look like without theo and design tokens.

Notes on code changes:
- npm package containing tokens has been removed
- default theme has been kept and previously machine-generated files has been committed as a reference which should be the future way of managing themes when not having theo layer
- configuration files have been updated accordingly in order to make changes mergeable should there be no doubts in the reduced value of theo


src/themes/default/index.scss
```diff
@@ -1,13 +1,11 @@
...
-@import './exports/colors';
-@import './exports/colors-bg';
-@import './exports/font-size';
-@import './exports/font-size-print';
-@import './exports/media';
-@import './exports/spacing';
-@import './exports/spacing-print';
-@import './exports/widths';
-@import './exports/z-index';
+@import './maps/colors';
```
emeryro, line 3: what is the difference between variables in `variables` and the one in `maps/`?
do we need them both?

kalinchernev: Good question, asked myself the same.
`src/themes/default/variables/index.scss` is what was generated as a file with "flat" definition of variables. The index file is 1:1 to what has been previously "all.scss". Can be split into sub-files with specifics and that's why index file is left.
`src/themes/default/maps` seems to me to be the same set of information but organized in a different way. I briefly browsed the repo in v2 and there are a few cases where this approach has been used https://github.com/ec-europa/europa-component-library/search?q=map-get
As I'm not sure which approach will be preferable for the future I kept both.

emeryro: Personally, I find the definition done in maps better than the flat list. But that's something than can be discussed in the sass/css/theme track. We will keep only one variable definition for sure.
In any case, we will need a definition of unit variables (previous design tokens), and maybe some custom variables like composed font (with font family, font size, line height, ...).

kalinchernev: Ok, I don't mind having maps over flat variables. The [button](https://github.com/ec-europa/europa-component-library/blob/23a6ac254faf8e42fa87d79764b9c7bbe7c4b3d7/src/implementations/vanilla/packages/ec-component-button/ec-component-button.scss) component which is currently the only one left for a reference is using flat variables 
I don't know which are the unit variables and i'm not sure whether to do anything on it at this stage.
What is the suggested next step here? Shall we refactor things in a given way in this PR?

emeryro: Well I guess we could just keep the variables as they are at this stage, I would just add some comment (in the code or in a css related ticket) saying that the flat variables should be removed and the demo component updated

kalinchernev: Ok, created 2 new tickets to follow up on this. Reading current code, there is the question of theming https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1694 (inheritance vs overrides) on a component level.
And the one for refactoring the variables specifically: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1695
Maybe there is no need to re-evaluate SCSS approach for import checks and theming before this, but still could be helpful having answers before changing the whole set of files.

emeryro: ok then, let's merge

PR #1689, opened 2020-09-28
docs(list): update timeline usage page - FRONT-1335

gingapa:
Updating timeline usage page, adding "ecl-link--standalone to timeline items and changing collapse threshold to 3.


src/website/src/pages/ec/components/timeline/docs/usage.mdx
```diff
@@ -32,17 +32,17 @@ import { Paragraph, Anatomy, Link } from '@ecl/website-components';
 
 ### Default
 
-| Elements  | Mandatory | Description                                                       |
-| --------- | --------- | ----------------------------------------------------------------- |
-| heading   | yes       | heading of the item                                               |
-| timestamp | yes       | timestamp when the item happened (or is scheduled to)             |
-| details   | no        | a more detailed description. Can include text, links, media files |
+| Elements  | Mandatory Desktop | Mandatory Mobile | Description                                                       |
```
planctus: I think in all these use cases where we have the same for mobile and desktop we don't add a column for each of them, just a single one: "mandatory".

src/website/src/pages/ec/components/timeline/docs/usage.mdx
```diff
@@ -32,17 +32,17 @@ import { Paragraph, Anatomy, Link } from '@ecl/website-components';
...
+| --------- | ----------------- | ---------------- | ----------------------------------------------------------------- |
+| heading   | yes               | yes              | heading of the item                                               |
```
planctus: I was honestly expecting changes on these values, i see they are untouched instead, this reproduces the inconsistency that originated this request for updating this usage page, or more generically to align the demo with the usage page.
In the demo we don't have "titles" for many of the items, but here we say it's mandatory.

I do remember the conversations with the designers where they were actually convinced it was the usage page that needed to be "fixed", instead i see that we should fix the demo..

src/website/src/pages/ec/components/timeline/docs/usage.mdx
```diff
@@ -32,17 +32,17 @@ import { Paragraph, Anatomy, Link } from '@ecl/website-components';
...
 
-### Conditional
+### Conditional (over 12 items)
 
-| elements      | Mandatory | Description                                                                                                                                                          |
-| ------------- | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| expand button | yes       | actionable button to display the full list of items in the timeline - to be used **if** there are over 12 items in the timeline<br/>**Position**: under the 3rd item |
+| elements      | Mandatory Desktop | Mandatory Mobile | Description                                                                                                         |
+| ------------- | ----------------- | ---------------- | ------------------------------------------------------------------------------------------------------------------- |
+| expand button | yes               | yes              | if there are over 12 items in the timeline then Actionable button to display the full list of items in the timeline |
```
planctus: I wonder a bit were we retrieved this requirement of always showing the expandable button after the third element, i read the usage page and i don't see any reference to it, although the example shown in the "anatomy" places the button after the third, i would not assume it as a "requirement", maybe.

src/website/src/pages/ec/components/timeline/docs/usage.mdx
```diff
@@ -32,17 +32,17 @@ import { Paragraph, Anatomy, Link } from '@ecl/website-components';
...
+| timestamp | yes               | yes              | timestamp when the item happened (or is scheduled to)             |
+| details   | no                | no               | a more detailed description. Can include text, links, media files |
 
-### Conditional
+### Conditional (over 12 items)
 
-| elements      | Mandatory | Description                                                                                                                                                          |
-| ------------- | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| expand button | yes       | actionable button to display the full list of items in the timeline - to be used **if** there are over 12 items in the timeline<br/>**Position**: under the 3rd item |
+| elements      | Mandatory Desktop | Mandatory Mobile | Description                                                                                                         |
```
planctus: Also here no need to duplicate the columns.

src/website/src/pages/ec/components/timeline/docs/usage.mdx
planctus: I understand we received this in the "specs" for this update, but is this sentence meaning anything to you?
_if there are over 12 items in the timeline then Actionable button to display the full list of items in the timeline_
Anyway...if that's what they want.

Ah, Paulo, this is one of those prs for which we request approval by the "business". Once we get it the way we want we won't merge, we will ask first for an approval by "who created the issue" ;)

planctus (changes requested):
So, since the issue was about aligning the usage page with the demo, if the decision is that the "headings" are mandatory then we should show these "headings" for all of the items we set in the demo.

src/website/src/pages/ec/components/timeline/docs/usage.mdx
gingapa: There was a reference to it before:
**if** there are over 12 items in the timeline<br/>**Position**: under the 3rd item
Additionally they commented on the ticket:
_"(...)Also the "show more" button placement should be under the 3rd item (as seen on the usage page), not sure if it was mentioned."_

gingapa: I agree, it's not their best sentence...

planctus:
Thanks for the changes, @gingapa, i have a question still regarding the decision of fixing the "position" of the expandable. From the usage page it is impossible to find a requirement about this, i think we need to ask for clarifications, at the moment we are also showing two items after, in the "anatomy" there is only one, so if that was the key for this decision i would not understand why not using also for this.

Personally i don't feel confident with these changes, we should have been clarifying something thanks to this but i have more doubts that in the past.. :)

I'm asking for an alignment on the comm side, between ux and designers, let's maybe wait for this, can i just ask you to put some more realistic titles in the demo? Not a big deal, but it improves if we set there real examples, even a long title i would say,,Thanks

planctus:
Ah, there is some linter issue to be fixed, currently, @gingapa ;)

planctus (approved):

PR #1690, opened 2020-09-29
feat(exceptions): Handling two fake exceptions, different specs for button, different demo for icon - FRONT-1689

planctus:
# PR description

In this pr there is just a demo of two possible "exceptions" to be handled:
1) different input data for the two systems
2) different variants demoed in storybook

a third exception is not handled here, two different "styles" for the two systems, this doesn't seem possible at this stahe

The solution is the one in use in ecl-twig, basically:
A system variable that informs storybook about the system to "build", leading us to:
1) Being able to choose the right spec to use in a demo (story file)
2) Being able to select the right story in case there are differences in the variants to be demoed, for instance.

At the moment this is only working in a local environment, we currently (in next-v3) "distribute" only one instance of storybook therefore it is not possible to preview the differences in the deployed package.
The approach, following the ecl-twig implementation, is based on the idea of two different storybook instances, but it looks like it's too early to make such a decision, so for now i have not changed this, just adapted the main.js to show a conditional definition of the stories to be loaded.

**Not meant to be merged**, not real examples, here the changes are just meant to demo this "exceptions" handling. 

UPDATE: I've found a way to "demo" also the use case of a component with a different style defined in the two systems without changing things too much, but i still had to add a couple of packages in the vanilla folder.

## QA Checklist

- [ ] I have provided tests


src/specs/components/button/demo/eu-data--primary.js
```diff
@@ -0,0 +1,5 @@
+// Simple content for demo
```
kalinchernev, line 1: To my understanding, this file should be an example of what is referred to in the description of the pull request as "different input data for the two systems".
If this is true, I expect that there is actual difference in the data structure.

src/implementations/react/storybook/package.json
```diff
@@ -8,6 +8,7 @@
   "scripts": {
     "build": "build-storybook -c .storybook -o build",
     "start": "start-storybook -p 6006 -s ../../vanilla/packages/ec-preset-website/dist,../../vanilla/packages/ec-preset-editor/dist -c .storybook",
+    "start:eu": "STORYBOOK_SYSTEM=EU start-storybook -p 6007 -s ../../vanilla/packages/ec-preset-website/dist,../../vanilla/packages/ec-preset-editor/dist -c .storybook",
```
kalinchernev: This I map to "different variants demoed in storybook". I am ok with the approach of having an environment variable for setting an ec/eu context.

src/implementations/react/components/button/stories/Index.jsx
```diff
@@ -6,9 +6,13 @@ import demoContentSecondary from '@ecl/ec-specs-button/demo/data--secondary';
 import demoContentCall from '@ecl/ec-specs-button/demo/data--call';
 import demoContentGhost from '@ecl/ec-specs-button/demo/data--ghost';
 import demoContentSearch from '@ecl/ec-specs-button/demo/data--search';
+import euDemoContentPrimary from '@ecl/ec-specs-button/demo/eu-data--primary';
```
kalinchernev, line 9: Relying on systems separation in terms of file systems. It's the same as what we have in v2.

As mentioned here https://github.com/ec-europa/europa-component-library/pull/1690/files#r496654688 I think the example demonstrating data structure difference is not fully matching foreseen hardship in input data. At the moment, there are no differences in primary button data and there is no need for any split.

On the other hand, if there were a difference in the input data for the primary button between ec and eu, it shouldn't be handled in storybook in my opinion. Rather, it should be part of the checks inside the templates. (which is currently react, but will be twig)

I don't remember where was the document by Antonio regarding how the platform plans to load assets based on system context, but i doubt it's a layer before templates.

The topic of input data is separate from the number of storybook instances we can have in the repository.

src/implementations/react/storybook/package.json
kalinchernev: Regarding 
> a third exception is not handled here, two different "styles" for the two systems

Although style-related exceptions are not considered, it's worth considering change in styles for the scope of the exception handling evaluation.

This could be as simple as copy/paste of default theme and changing all blue colors to red in https://github.com/ec-europa/europa-component-library/blob/next-v3/src/themes/default/variables/index.scss

The same environment variable STORYBOOK_SYSTEM (or just SYSTEM) can be also used in the website builder config https://github.com/ec-europa/europa-component-library/blob/next-v3/src/implementations/vanilla/packages/ec-preset-website/ecl-builder.config.js#L30-L31 to compile ec/eu theme conditionally. (if file-per-system still)
I personally hope that passing such an environment variable to scss will also be possible, so that index import files have `$system` of some sort.
Please check our [builder options parameters](https://github.com/ec-europa/europa-component-library/blob/next-v3/src/tools/builder/scripts/styles.js#L35). Documentation says that the `data` parameter (which we'll need to add) supports passing data. See https://bl.ocks.org/Leimi/a9f125406c1efd916448 
Then all occurrences of `@import '@ecl/ec-theme-default/index';` can be system-aware and import the right theme based on an env variable.

src/implementations/react/components/icon/stories/EuIndex.jsx
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
```
kalinchernev, line 1: Let's please keep only the button for the POC in order to be able to keep the boilerplate reduced

src/implementations/react/storybook/package.json
kalinchernev: One of the ideas of storybook theme switchers in FRONT-1684 was basing on loading (let's say) 2 css files for ec+eu and storybook switching their loading through an environment variable too https://storybook.js.org/docs/react/configure/environment-variables

For the scope of this ticket we have 2 separate storybook configurations which only load 1 system file (ec or eu) and no need for a whole switcher invesitation at this stage

src/specs/components/button/demo/eu-data--primary.js
planctus: i kind of see what you refer to when you think of a different data structure, but that's something already handled in v2, i guess, by having "different" properties, in EU there are two logos for the site-headers, for instance, in EC only one, but that only means that there is an additional property to be passed when implementing the component in EU.
 
I don't personally have and i don't see any use case where the data structure would be screwed enough so that, i don't know, a "description" property would change into something else in the other system, or where all the properties would be re-shuffled, which has always been to my understanding what "the same data structure" meant for the core team and that is also the current situation.

The fact that some properties might only be meant to be used in a system in order to support reasonable differences in the two systems is already in V2, not sure we have to do anything about it, what is not the case in V2, instead, is the possibility, based on the same template, to alternatively show a certain "input data" or another based on the system in use, it's not a big deal as shown in here, but that is what this POC is mainly about. :)

src/implementations/react/components/button/stories/Index.jsx
planctus: _Relying on systems separation in terms of file systems. It's the same as what we have in v2._
But how is the same in V2... there you have two story files and you choose which file to import, here you have one story file that chooses the right one.

I wanted simply to handle different content (not only, but mainly) in the demos we make based on the system in use, which is a use case that we already have and we would not be able to reproduce in v3 without a solution for it, even easy as setting a variable is, i'm not trying to guess anything that i cannot imagine... :)

The twig template is already the same for the two systems in v2, core-team is already using the same template in the two systems, so in ECL v2 we are already handling and at the moment without informing the templates about the context, they are "system agnostic", as i mentioned before this is handled with "no additional logic in the templates" and i wouldn't change this unless there is a clear need for that, but also that is a possibility for V3, not excluding that.

Then we have the track about potential, "overall" changes in the definition of the data structure to facilitate the integration, if there is anything we can do for that, but that's on a different level, hope we can work on that even before making any "real" component here...

src/implementations/react/storybook/package.json
planctus: I did not "demo" any css difference because i've tried and while doing it i realised that many other decisions had to come before spending time on this.
I have no deep knowledge or understanding of the scss/css pipeline in ECL, i spent time in trying more than in digging into it, which is what you suggest me and by this you also give me a different approach i did not think about. 
Still it remains the feeling that is way too early to define "exceptions" at this specific level, because the overall process of generating a css might significantly change in V3 and it's probably better to evaluate this one we have clearer ideas about it.

src/implementations/react/components/icon/stories/EuIndex.jsx
planctus: Yeah, but i don't plan to merge anything from here, honestly, it's a very simple approach the one shown here and it depends on the decisions we'll make regarding all the other things, file structure, storybook configurations, build configurations..the ticket i think will still be in place for a while, surely not "fulfilling" its scope with this simple implementation.
Open to different opinions about this, anyway.  

Thanks a lot for the review

src/implementations/react/storybook/package.json
planctus: So, i've found a way to demo also the "exception" of a component with a different style in the two "systems", but for doing it i needed eu-preset-dev (where the single components scss is imported), then i "cloned" ec-preset-website and set up the logic in preview.js for storybook but with a single instance we have the issue that resources are also "hardcoded" in html files, where the env variable would not be usable.
I did not go for setting logic in the builder because it builds the packages one by one and out of the storybook environment, where the env variable is not available, the logic it has is still suitable for this use case, so i kept it as it is.
In general, i still imagine two storybook instances (.storybook folders..) in the end, but i did not want to touch this in this POC.

planctus: Ah, i focused on the "exception" case only, where is the component style that differs in the two systems, the themes were not relevant in this use case and i don't know much about the current "presets", nor about the future themes with storybook, i have to digg into this to get an understanding about it, not necessarily in the context of this POC.

src/implementations/react/components/button/stories/Index.jsx
kalinchernev: >>The twig template is already the same for the two systems in v2, core-team is already using the same template in the two systems, so in ECL v2 we are already handling and at the moment without informing the templates about the context, they are "system agnostic"

Ok. given template layer is solved, the demo here is showing a preparatory layer before templates selecting the right data based on a context of some sort. As I mentioned, if it's not breaking the way EWP is handling the data too, I don't mind having this selection logic on file-level in storybook

src/specs/components/button/demo/eu-data--primary.js
kalinchernev: Happy to see a confirmation to my observations. To my knowledge there is no (and should not be) reshuffling of data and we don't need to spend additional time on data handling layer.

src/implementations/react/storybook/package.json
kalinchernev: So, when I placed my suggestions above yesterday I thought the css layer on handling multiple themes/systems/whatever will be a straight-forward task with dynamic imports in scss. Wasn't the case.
Basically, yesterday I assumed many things regarding the CSS layer in ECL which were wrong. This layer is not my specialty neither. Agree there are several things to validate before being able to evaluate exceptions there. Let's leave it out for now. 

PS Please have a look at https://github.com/ec-europa/europa-component-library/pull/1692 and share your opinion

PR #1692, opened 2020-09-30
POC: multiple themes from 1 file and 1 config

kalinchernev:
This pull request attempts to illustrate and approach of generating multiple css themes based on a single entry.
Main goal being to make an attempt to provide dynamic theming without re-introducing code duplication.
It's possible that there are bugs in the overall picture and rough edges and maybe stupid ideas.

**Please do let me know if you have better ideas on how to make themes generation better without code duplication.** 

## Core concepts (and assumptions):
- `@ecl/themes` will be an npm package containing all theme variations. We can assume it will be publishe-able on npm for consumers to get all themes at once.
- `@ecl/builder` remains an abstraction on top of optimisation tools. (rollup bundler, babel transpiler, etc.) Users of the `@ecl/builder` (such as the platform team) should be able to use the tool interchangeably between v2 and v3 without many changes. Backwards compatibility is a plus.
- presets and other collections of packages (as well as oe_theme) are still able to have their own ecl-builder configurations override
- 'theme' is currently referring to 'ec' and 'eu' though these can be something different down the line
- a theme should be set as globally as possible (environment variable) and used throughout all building scripts consistently

For the sake of simplicity, the following prerequisites are in place:
- @ecl/ec-preset-website is taken as an example application having a builder config
- @ecl/ec-themes has a new "eu" theme which is basically what we currently have in the eu system in v2. Code follows the pattern based on https://github.com/ec-europa/europa-component-library/pull/1688 For quick validations and obvious demo of the difference, $ecl-color-blue-100 is set to red. Thus "eu" css primary button is red.

Desired functionality:

Builder's config can optionally have the following new settings:

```js
  themes: {
    list: ['ec', 'eu'],
    module: `${nodeModules}/@ecl/ec-themes`,
  }
``` 

`module` part instructs the new logic inside builder's script command on where to place theme markers.

When `yarn build:scripts` from the website preset or `yarn dist` from root, end result can optionally contain ec/eu separation in terms of bundled css assets.

These separated CSS assets can later be loaded either in storybook or other systems conditinally for a system/theme speicfic styling.

## Constraints

It is currently nearly impossible to have dynamic imports in sass. See comments in `src/tools/builder/scripts/styles.js` for details.
For this reason, `src/themes/index.scs` which is a central place for setting a theme becomes a file-based marker generated during build time.
The simplicity of having 1 file with 1 line setting a theme has the benefit of simplicity, but has the disadvantage of being a blocking approach. It's not possible to generate multiple themes at the same time because this marker file is also a lock file.

We can either decide to provide a default marker and commit/publish it, or not. When it's not present, any script building task will fail inthe absence of this file. The second approach is the currently implemented one. A failure on missing index file in a themes module seems to me an easy to track error of absence of a choice of a theme to build.

In an attempt to make themes management easier in long run, a new "theme" command has been added to the ecl-builder:
- `yarn ecl-builder theme`: shows current status
- `yarn ecl-builder theme --set (ec|eu)`
- `yarn ecl-builder theme --remove`

When environment variables are used for JavaScript assets generation and storybook, etc., these can also be managed through this command so that developers manage themes in 1 command for all target assets.


kalinchernev:
Added a few tweaks to potentially accommodate v2 ecl builder config in the new v3 context with log messages and the assumption that previously known systems will become themes.

Note that website preview panes and storybook demoes are not updated to link new stylesheets and components are not styled. If there is an internal agreement on the approach suggested in the pull request, i'll update these links of course.

planctus:
**THEMES** (for how they are defined together with comm) are about:
- Fonts
- Secondary Colors
- Border Radius
- Corner
- Shade

But it means that they only define a value for them, not the whole set of the possibilities, to be then propagated in the components, i see it as a single "file" with "constants" defined.

**THE BASE** (common or not?):

It seems that all the maps that we currently have in each theme could be actually common between the different themes, these are defining variables that would not change, what changes is a set of "constants" that needs to be applied potentially across all the css pipeline, but most likely in the presets, where components scss are included.

So:
- We could define a common base that is common to the two systems (instead of ec-base/eu-base, i'm confused about what the current default theme is and how that is different from the base package), but the sense here is that the "old" theme is actually the base for all the new themes, which are meant to differentiate the styles of the components, mainly. (the font global parameter is probably going into a _global.scss, instead)

- systems could be just an abstraction used in storybook to load different input data or defining different stories, in the end, the need for it in the base/theme/presets/css/js definition is only to package the right resources in a distributed package, it seems.

- These distributed packages should be defined and shipped separately, i believe, i imagine a "preset" package that statically includes one theme (one package per theme) when compiling the components. This is also the v2 way, basically, but without all the duplicated source files for resources defining the two different base packages.

- Although i'm fascinated by this kind of dynamic theme generation, and i see how this could have been even smarter if dynamic imports would be available, at the same time i may not see the need for it, what i imagine is probably very close to what we do in v2.

- Also having ec-themes and not a package per theme would be trickier from a consumer point of view, i think it's more straightforward to fetch a package which only requires the specific theme package and having a conf for a single set of resources generation (css/js) defined by it (the way it is) .

So, to mimic the real scenario we have:

@ecl/ecl-base -> all the common resources, mostly maps
@ecl/ecl-ec-logo -> Logo assets for EC
@ecl/ecl-eu-logo -> Logo assets for EU
@ecl/ecl-icons -> Icons resources (this is ultra simplified, there might be multiple icons set)
@ecl/ecl-theme-{ec-core} -> A set of variables only
@ecl/ecl-theme-{eu-core} -> A set of variables only
@ecl/ecl-theme-{ec-standardised} -> A set of variables only
@ecl/ecl-theme-{eu-standardised} -> A set of variables only
@ecl/ecl-theme-custom -> A set of variables only
@ecl/ecl-preset-{ec-core} -> Base + theme + components
@ecl/ecl-preset-{eu-core} -> Base + theme + components
@ecl/ecl-preset-{ec-standardised} -> Base + theme + components
@ecl/ecl-preset-{eu-standardised} -> Base + theme + components
@ecl/ecl-preset-custom-theme -> Base + empty theme + components (for a new theme generation on the client side)

It doesn't look like this would involve duplication of code (well, many presets mean also many packages lists in a index...), surely more packages and i see how these might increase in number as soon as new themes might be defined, but at the same time it feels more structured and plus we always said that this approach only made sense with a defined, limited amount of different themes, while the dynamic theme generation gives more flexibility than what is needed, maybe.

Thanks a lot for this work, i think i could start thinking "globally" at the picture only thanks to this, although it's still not easy, we are maybe still forced to assume too many things to see them clearly, i wrote, cancelled, wrote again then edited many of these considerations and if i read them again most likely i would re-start the process.. :)

_Indeed i had to try a bit to understand if this made sense, not sure it does still but in the `poc-v3-css-js-chain` branch there is a sort of an implementation of what i described, with the themes being an hyper minimalistic (too much) packages with an index.scss file only. Also storybook is configured for "EC" and "EU" and has the theme switch provided by css resources like in the romain's POC, but i find it really sub-optimal, hope we find something better. In that approach no change is needed about the builder, a preset would still build just a "set" of resources, corresponding to a single "theme"_

kalinchernev:
@emeryro could you please have a look and share your thoughts?
At this stage we have 3 POC-s suggesting on how to organize SCSS and theming.
Reading through @planctus feedback, I have the impression that so far that the approach of you and him so far solves it. (leaving the storybook topic aside)
If you confirm this, we can close this one as an irrelevant and confirm there is no need for changes in the toolchain related to SCSS and proceed with the question of organizing assets reusing existing toolchain.

emeryro:
First of all, thanks for the work done here!
If we start from the end, what is needed is to have one css file for each theme, and I think that all the approach fullfill this need.
I like the idea of having some global variables (defined in maps), and only theme related variables in each theme (basically some booleans or strings). We somehow keep the advantage of design tokens (defined only in one place), but without the overlay we had previously.

The main advantage I see in Kalin's proposal is that we would have access to environment variables to know what theme is currently in used. I think that we will need it at some point (to handle scripts, and maybe storybook).
Having the themes grouped in the same folder/package is not a bad idea, as it would make it very easy to add/remove theme in the future (for us, and maybe for clients). Having to handle 1 theme folder + 1 preset for each theme is not a big overload, but I'm not sure that it is justified.

So my proposal would be a mix of yours:
- @ecl/ecl-base -> all the common resources, mostly maps
- resources (icon, logo, ...), grouped in a single package or not
- components
- @ecl/ecl-theme -> Several set of variables (one per theme)
- @ecl/ecl-preset-{ec-core} -> Base + theme ec core + components
- @ecl/ecl-preset-{eu-core} -> Base + theme eu core + components
- @ecl/ecl-preset-{ec-standardised} -> Base + theme ec standardised + components
- @ecl/ecl-preset-{eu-standardised} -> Base + theme eu standardised + components
- @ecl/ecl-preset-{custom} -> Base + empty theme + components (for a new theme generation on the client side)

Now, depending on the needs we may also consider the option to put the resources in the theme directly (if the icon are different for instance).

PR #1696, opened 2020-10-05
fix(ecl-editor): fix css conflict - FRONT-1697

emeryro:
# PR description

Fix css specificity conflict with file component

### DIFF
--- a/src/systems/ec/implementations/vanilla/packages/ec-editor/_ec-editor--button.scss
+++ b/src/systems/ec/implementations/vanilla/packages/ec-editor/_ec-editor--button.scss
@@ -12,7 +12,7 @@
   $padding-vertical: $ecl-spacing-m
 ) {
   /* stylelint-disable selector-no-qualifying-type */
-  .ecl-editor button,
+  .ecl-editor button:not(.ecl-file__translation-toggle),
   .ecl-editor input[type='button'],
   .ecl-editor input[type='submit'] {
     appearance: none;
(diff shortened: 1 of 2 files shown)


kalinchernev (changes requested):
I tested the fix on the page given in the ticket description.
The hover state will also need the update

video

emeryro:
@kalinchernev the hover/focus state should be covered already: https://github.com/ec-europa/europa-component-library/blob/FRONT-1697-editor-conflict/src/systems/ec/implementations/vanilla/packages/ec-editor/_ec-editor--button.scss#L31

kalinchernev:
I think some tweaks are still necessary. The

```css
background-color: $ecl-color-blue-130;
``` 

make so that when hovering over the button it becomes unreadable.

emeryro:
yes, but my point is: in the compiled css it will be handled already with the proposed modification.
here is the scss:
```
.ecl-editor button:not(.ecl-file__translation-toggle),
.ecl-editor input[type='button'],
.ecl-editor input[type='submit'] {
  appearance: none;
  background: none;
  background-color: $ecl-color-blue-100;
  [...]

  &:hover,
  &:focus,
  &:active {
    background-color: $ecl-color-blue-130;
    text-decoration: none;
  }
```
So unless I'm wrong, this specific button won't have the blue background at all (including on hover)

kalinchernev (approved):

PR #1698, opened 2020-10-06
feat(specs-v3): Redefining specs for icons and button based on the template props - FRONT-1705

planctus:
# PR description

My proposal would also be to turn these spec packages into private ones.

## QA Checklist

- [ ] I have provided tests


src/implementations/twig/components/ecl-component-button/__snapshots__/button.test.js.snap
```diff
@@ -154,7 +154,7 @@ exports[`EC - Button Primary renders correctly 1`] = `
 exports[`EC - Button Search renders correctly 1`] = `
 <button
   class="ecl-button ecl-button--search"
-  type="submit"
+  type="button"
```
kalinchernev, line 157: Are you sure that this change is desired?

src/implementations/twig/components/ecl-component-button/button.test.js
```diff
@@ -1,14 +1,14 @@
 import { merge, renderTwigFileAsNode } from '@ecl/test-utils';
 
 // Import data for tests
-import dataPrimary from '@ecl/ec-specs-button/demo/data--primary';
-import dataSecondary from '@ecl/ec-specs-button/demo/data--secondary';
-import dataCall from '@ecl/ec-specs-button/demo/data--call';
-import dataGhost from '@ecl/ec-specs-button/demo/data--ghost';
-import dataSearch from '@ecl/ec-specs-button/demo/data--search';
+import dataPrimary from '@ecl/ecl-specs-button/demo/data--primary';
```
kalinchernev: The number of stories growing, please organize them like here https://github.com/ec-europa/europa-component-library/blob/v2-dev/src/systems/ec/implementations/react/components/page-banner/stories/Index.jsx

src/implementations/twig/components/ecl-component-icon/package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "@ecl/ec-twig-component-icon",
+  "name": "@ecl/ecl-component-icon",
```
kalinchernev: Renaming packages, files and folders is not in the scope of this ticket. It's in FRONT-1658
I don't mind having the two, but I do not agree with pattern provided. `@ecl/ecl-` as it's clearly an unnecessary repetition.
We can open the discussion on naming here or leave it for FRONT-1658. Up to you.

kalinchernev: I personally would like to keep implementation specific path in the pattern: i.e. do not remove `twig`. We'll surely need a separation of base/vanilla and thus removing removing implementation part is unnecessary step.

src/implementations/twig/storybook/package.json
```diff
@@ -15,10 +15,10 @@
     "@babel/preset-env": "7.11.0",
     "@babel/preset-react": "7.10.4",
     "@babel/runtime": "7.11.2",
-    "@storybook/addon-cssresources": "6.0.22",
-    "@storybook/addon-knobs": "6.0.22",
-    "@storybook/addon-viewport": "6.0.22",
-    "@storybook/html": "6.0.22",
+    "@storybook/addon-cssresources": "6.0.26",
```
kalinchernev: Why upgrading these at this pull request?

kalinchernev (changes requested):
Thanks for the updates in the specs. I noticed a few other changes which do not seem to be relevant here

src/implementations/twig/components/ecl-component-button/__snapshots__/button.test.js.snap
planctus: Yep, i've checked them one by one, i will then check also in ecl-twig the reason why of the "wrong" values for the type in the test, but here they are correct.

src/implementations/twig/components/ecl-component-icon/package.json
planctus: you are right two times, this was actually a mistake, the idea was to have them as ecl-{implementation}-component-{componentName} like i did for the button, i will revert those changes keeping the package names as they were.

src/implementations/twig/storybook/package.json
planctus: well, i wanted to fix the problem we have with the text knob in the current next-v3 branch, basically, i noticed that when we had two different major version of storybook in react and twig for some reason this triggers a "bug" and it breaks if you try to access the knobs. Then i noticed that the version was upgradable (0.22 to 0.26) and i took the chance. I can put it back to 0.22, in both the packages here but at the same time why not updating them?

src/implementations/twig/components/ecl-component-button/__snapshots__/button.test.js.snap
emeryro: well, in the case of a search form for instance, the button should remain a "submit" I think

src/implementations/twig/storybook/package.json
kalinchernev: No problem making an upgrade, sure. My thoughts were rather into the major bump in the [react one](https://github.com/ec-europa/europa-component-library/pull/1698/files#diff-74dfbd6d83f634d50878093b347f827aL19). First reaction was that we need to double-check more carefully after such a bump. But the second reaction is that react is going away, so no issue making whatever there.
If the updates are fixing the issue with knobs, good, keep the update pls

src/implementations/twig/components/ecl-component-button/__snapshots__/button.test.js.snap
planctus: But do we have to discuss here this? I mean, in ECL, in production, the search button example is a type="button"
image

I think it would be in case more relevant for V2 than V3, then, here we are for now just using these templates to play with the structure, i believe, not defining the new V3 version of them.
 
Whatever, for me, if you want to have it as as submit here, it's a line to be changed in the relative specs..

kalinchernev:
Changed to "changes requested" only for https://github.com/ec-europa/europa-component-library/pull/1698#discussion_r500734313
The rest is ok for me

src/implementations/twig/components/ecl-component-button/button.test.js
planctus: Mmhh..this surprises me, basically this would mean creating a file per variant, to export a single story, right? We said we would try to keep the duplication of code under control, this would introduce lots of it, no? Many more files to handle, at least.
Not sure, then, if we want to move to the current approach in ECL v2 it's worth maybe doing it in a separate ticket for both the "demo" components we have, but i'm not sure this would be an advantage, at least i think it would be worth discussing it before making that move.

src/implementations/twig/storybook/package.json
planctus: ok, it's fine with 6.0.22, also the private packages have a dep on storybook addons and it's set on 6.0.22. The react private packages are not aligned, instead, but as you said this will disappear together with react, so i'm not touching them, knobs are working in the preview.

src/implementations/twig/components/ecl-component-button/button.test.js
kalinchernev: What I meant was having 1 story file https://github.com/ec-europa/europa-component-library/blob/98f4647b41f805ce05a0bb65518ccca69ca669f1/src/implementations/twig/components/ecl-component-button/button.story.js of 127 lines of code and repetition being split into several smaller files. Duplication topic is WTF here. Rather it's a usual approach of having smaller files to work on, that's literally all i'm asking for.

kalinchernev: Skip it if you prefer btw, all the same for me

planctus: Yeah, but that's what i intended, lots of files to deal with, i'm not really discussing a general pattern in software development, neither expressing my preference about it, just considering the opportunity of making such a decision in the context of this pr and only for the button. I won't skip it, anyway.

PR #1699, opened 2020-10-06
chore(implementations): remove react - FRONT-1702

kalinchernev:
Highlights:
- react implementation has been removed in favor of twig (via twing)
- ecl website is reusing same twing loader and environment settings from twig implementation storybook
- spec and adapter files reused, although https://github.com/ec-europa/europa-component-library/pull/1698 might require rework on this

I haven't spent the time to update all demo variations on the website. They have been commented out for reference purposes.


emeryro:
Thanks @kalinchernev , this is working fine
I would just suggest to have at least one working example of button on the website, based on twig. That way we could use it as reference.
Idealy we should try to maintain the demo component up and running, but this could be updated in another PR if needed

src/implementations/twig/storybook/package.json
```diff
@@ -3,7 +3,7 @@
   "name": "@ecl/ec-twig-storybook",
   "author": "European Commission",
   "license": "EUPL-1.1",
-  "version": "2.31.1",
+  "version": "2.33.0",
```
emeryro: I guess we should put it to `2.34` already

src/website/package.json
```diff
@@ -41,6 +41,7 @@
     "svg4everybody": "2.1.9"
   },
   "devDependencies": {
+    "@ecl/ec-twig-storybook": "^2.33.0",
```
emeryro: here too

src/website/src/pages/ec/components/button/demo/index.js
```diff
@@ -0,0 +1,13 @@
+import demoContentPrimary from '@ecl/ec-specs-button/demo/data--primary';
+import demoContentSecondary from '@ecl/ec-specs-button/demo/data--secondary';
+import demoContentCall from '@ecl/ec-specs-button/demo/data--call';
+import demoContentGhost from '@ecl/ec-specs-button/demo/data--ghost';
+import demoContentSearch from '@ecl/ec-specs-button/demo/data--search';
+import template from '@ecl/ec-twig-component-button/ecl-button.html.twig';
+import { correctSvgPath } from '@ecl/website-utils';
```
emeryro, line 7: I don't think this is used here

emeryro (changes requested):
just spotted a few minor things, apart of that it's all good

src/website/src/pages/ec/components/button/demo/index.js
kalinchernev: It's used in this case
```js
export const buttonCall = template(correctSvgPath(demoContentCall));
```
Each time we have a spec demo data coming in with an .svg extension, this one will update the path accordingly

src/implementations/twig/storybook/package.json
kalinchernev: Good point. I did a step forward on this: set all packages to "private" and version 3.0.0. A few unnecessary deps got dropped during this update.

emeryro (approved):

PR #1704, opened 2020-10-08
chore(scss): remove import check - FRONT-1680

emeryro:
# PR description

remove scss import check

I tried to use scss' `@use` commend, which is the new way of importing files, but it seems that it is not supported yet by libsass (which we rely on).
So for now we should stick to `@import`


kalinchernev (approved):
Thanks for the changes.

PR #1705, opened 2020-10-08
fix(menu): accessibility - FRONT-1456

gingapa:
Menu operation via keyboard (tab to select items, esc to close the menu)


src/systems/ec/implementations/vanilla/packages/ec-component-menu/ec-component-menu.scss
```diff
@@ -356,6 +356,19 @@
     }
 
     .ecl-menu:not(.ecl-menu--forced-mobile):not(.ecl-menu--forced-close) {
+      .ecl-menu__item--focused {
```
emeryro, line 359: I think that these rules can be merged with the `.ecl-menu__item:hover`ones below (to avoid duplication)

emeryro (changes requested):
It seems to work well. I guess that this kind of keyboard navigation is enough in our case.

src/systems/ec/implementations/vanilla/packages/ec-component-menu/ec-component-menu.scss
gingapa: @emeryro I tried to merge with .ecl-menu__item:hover before but Lint always complains about "no-descending-specificity" no matter what order I add them! Do you have any recommendation to bypass this?

Thanks!

gingapa: I added the following items to this PR:

- adding aria-hidden="true" on Icons
- adding <span className="ecl-u-sr-only"> (has submenu)</span> on Menu items with children
- make sure the menu can be closed using the esc key, both on mobile and desktop.

src/systems/ec/implementations/react/components/icon/src/Icon.jsx
```diff
@@ -37,7 +37,7 @@ const Icon = ({
     >
       {title && <title {...(titleId && { id: titleId })}>{title}</title>}
       {desc && <desc {...(descId && { id: descId })}>{desc}</desc>}
-      <use xlinkHref={`${iconPath}#${shape}`} />
+      <use xlinkHref={`${iconPath}#${shape}`} aria-hidden="true" />
```
emeryro: there is already an aria-hidden attribute on the parent svg tag, so this one is not needed

src/systems/ec/implementations/react/components/menu/src/MenuItem.jsx
```diff
@@ -25,6 +25,7 @@ export const MenuItem = ({ label, href, isCurrent, subItems }) => {
         data-ecl-menu-link
       >
         {label}
+        {hasSubItems && <span className="ecl-u-sr-only"> (has submenu)</span>}
```
emeryro: the string "(has submenu)" should be translatable, so it has to be put in the data file

emeryro (changes requested):
thanks for the improvements.
there are a few small points to be fixed below:

src/systems/ec/implementations/react/components/menu/src/MenuItem.jsx
gingapa: Please check now

emeryro:
that's almost good, but I'd just name the parameter `hasSubmenuLabel` instead of `hasSubmenu` (it looks more like a boolean), and maybe check in the template if that parameter is ampty or not (no need to add an empty span)

gingapa:
@emeryro please check now, I'm not sure this is the best way to check in the template if **hasSubmenuLabel** is empty!
I'm also adding this to EU.

emeryro (approved):

PR #1707, opened 2020-10-12
feat(default-styling): add documentation and example - FRONT-1666

emeryro:
The goal is to find a clean enough way to handle styling of default HTML tags

Please have a look and comment/propose changes on the documentation.
https://github.com/ec-europa/europa-component-library/blob/FRONT-1666-default-style/docs/decisions/006-default-style.md

You can find below a small example of the different approach proposed in the document (just uncomment the corresponding code)
[default-style.zip](https://github.com/ec-europa/europa-component-library/files/5363435/default-style.zip)

Note: please do not merge this PR as long as the document hasn't reach its final state

### DIFF
--- a/docs/decisions/006-default-style.md
+++ b/docs/decisions/006-default-style.md
@@ -1,0 +1,101 @@
+# Apply ECL 3 styling to default HTML tags
+
+| Status        | proposed                                              |
+| ------------- | ----------------------------------------------------- |
+| **Proposed**  | 12/10/2020                                            |
+| **Accepted**  | --/--/----                                            |
+| **Driver**    | @emeryro                                              |
+| **Approver**  |                                                       |
+| **Consulted** | @kalinchernev, @planctus                              |
+| **Informed**  | [@team](https://github.com/orgs/ec-europa/teams/inno) |
+
+## Decision
+
+[To be added]
+
+## Context
+
+- Styling of default HTML tags (without CSS class) has been a redundant request from some ECL users
+- In ECL 2, package `ecl-editor` was provided as a workaround. It worked quite well, but ended up raising some CSS conflict issues like [FRONT-1697](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1697)
+- In ECL 3 we shoud come with a more robust way to handle this, without disturbing the recommanded way to use ECL (by adding CSS classes)
+- We have no control over the final markup of pages. So we can't rely on any markup based selectors (like `>`, `+`, ...)
+
+## Requirements
+
+- This default styling should be made available as an optional CSS, or in an optional preset
+- The provided solution should be able to handle every HTML tag without having to update the code for each new component/utility
+
+## Consequences
+
+[To be added]
+
+## Alternatives Considered
+
+### Option 1: style elements directly
+
+**example:**
+
+- h1 { ... }
+- button { ... }
+
+**main pros:**
+
+- easy to override for component (very low specificity)
+
+**main cons:**
+
+- no easy way to prevent css leak to other elements on the page (Drupal admin menu, webtools widget, ...)
+
+This option is not robust enough and would surely ends up with unwanted side effects on the page.
+
+### Option 2: global namespace
+
+**example:**
+
+- .ecl h1 { ... }
+- .ecl button { ... }
+
+**main pros:**
+
+- no css leak
+
+**main cons:**
+
+- all components will also have to use .ecl or the specificity would be too low
+
+Having to use yet another namespace for the components would break the BEM structure and would impact all ECL users.
+
+### Option 3: global namespace excluding ecl components
+
+**example:**
+
+- .ecl h1:not([class*="ecl-"]) { ... }
+- .ecl button:not([class*="ecl-"]) { ... }
+
+**main pros:**
+
+- no css leak
+- no override of components
+
+**main pros:**
+
+- will also target elements using utilities
+
+The use of utilites remains a requirement for ECL 3, so having a solution working with them too is mandatory.
+
+### Option 4: global namespace excluding ecl components, but including utilities
+
+**example:**
+
+- .ecl h1:not([class*="ecl-"]), .ecl h1:is([class*="ecl-u-"]) { ... }
+- .ecl button:not([class*="ecl-"]), .ecl button:is([class*="ecl-u-"]) { ... }
+
+**main pros:**
+
+- no css leak
+- no override of components
+- still working with utilties
+
+**main cons:**
+
+- ?
--- a/docs/decisions/006-default-style.md
+++ b/docs/decisions/006-default-style.md
@@ -23,7 +23,7 @@
 ## Requirements
 
 - This default styling should be made available as an optional CSS, or in an optional preset
-- The provided solution should be able to handle every HTML tag without having to update the code for each new component/utility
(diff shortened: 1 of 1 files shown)


docs/decisions/006-default-style.md
```diff
@@ -0,0 +1,101 @@
...
+## Context
+
+- Styling of default HTML tags (without CSS class) has been a redundant request from some ECL users
+- In ECL 2, package `ecl-editor` was provided as a workaround. It worked quite well, but ended up raising some CSS conflict issues like [FRONT-1697](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1697)
+- In ECL 3 we shoud come with a more robust way to handle this, without disturbing the recommanded way to use ECL (by adding CSS classes)
+- We have no control over the final markup of pages. So we can't rely on any markup based selectors (like `>`, `+`, ...)
+
+## Requirements
+
+- This default styling should be made available as an optional CSS, or in an optional preset
```
planctus, line 25: These requirements are not the ones we always considered, i'd say: accepting or even embracing the idea of styling basic and limited html tags never meant we would support "every html tag".
We already defined the list of these supported tags, which are only `headings, global font definition, lists, tables`.

I have the impression we confuse the ecl-editor with this, because in the past it could have been used also as a workaround for styling un-themed markup, but ideally in v3 the editor would only have to ensure the "correct" rendering of a component markup once it is wrapped in some other html (.ecl-editor in particular), while the definition of basic styles for headings would let them, for instance, define an h2 inside the editor without having to specify a class.

I don't even see the reason why we would again create a "separated" something for this, like for saying we don't want it and we only do it for who requested this and it will be not be part of the standard css defined by the library, we don't have "presets" in v3 in terms of differentiated final output, we would reintroduce them just for this.

Anyway, overall, my preference would be for "ckeditor plugins", theming the markup as requested in ECL, the info website has some of these widgets (like headings for instance) since years, also the table widget is customised in info to output the desired markup, it would super easy to add a class to every block of text (paragraphs), i wonder how can this still be an issue for the d8 version of the platform. 
Surely we are not enthusiast about doing something that we feel might not be needed, but again, instead of making a big fuss around this, we have to simply set some global style, not reinventing ECL without markup, imho.

docs/decisions/006-default-style.md
```diff
@@ -0,0 +1,101 @@
...
+
+**main cons:**
+
+- no easy way to prevent css leak to other elements on the page (Drupal admin menu, webtools widget, ...)
+
+This option is not robust enough and would surely ends up with unwanted side effects on the page.
+
+### Option 2: global namespace
+
+**example:**
```
planctus, line 53: I believe that if we wrap this style definition into an ECL namespaced whatever then we reintroduce the same issue from another door, let's say.
Probably i don't entirely get the concerns, i mentioned info before and what we would have to do is something that has always been in place, there.

`h2:not([class*="ecl-"]) , .ecl-h2-class`

with one that extends the other, basically.

Sure this means that every definition of a style for a themed/un-themed h2 in a component that needs a different style would mean "overriding" a default rule, but i would also say that this could help us in building a consistent library, it's clear that there will not be any reason for h2s looking all different one from the other, for instance.

emeryro: Without a global namespace, the CSS would leak to other elements on the page (Drupal related stuff, webtools widgets, ...)
That's why I discarded this option

docs/decisions/006-default-style.md
emeryro: There are indeed confusion here.
Maybe I missed it, but I don't know where your list of tags to be styled comes from (I don't remember seeing this before).
Also, just to be clear, there is no intention to have something like ecl-editor in ECL 3. That's why these global styling are considered a more robust way to replace it

docs/decisions/006-default-style.md
planctus: But how could this be a concern for us? Webtools is usually working the other way around, their styles are often global and override ours. 
About Drupal, that's the intended behaviour, to be able to enforce some ecl style without the usage of specific markup, so this would be a feature, not a bug.

So, about these concerns, in the first case it's a matter of agreeing with someone on how to do things, the current situation is problematic already, the next one would still be if we don't all do the right thing:
ECL will define global styles for certain elements, injected styles by webtools should always be targeting their unique markup, it would be good for today, it will be good for tomorrow.

For Drupal, if there is any unwanted impact of these generic rules (i can't see it) we'll see, but how could this be the reason why we decide to drop the main, if not the only, option we have? This is what we always said "styling by default a limited set of html classes", if it has to be a themed markup again, i would not really understand why we wouldn't request the right markup.

docs/decisions/006-default-style.md
planctus: We explicitely mentioned the tags we would style in many occasions, because we knew we had to limit this to a reasonable amount, if you need to find this back we probably have this list in one of the multiple places where we dislocated our conversations, documents, etc..but to me, it's not a matter of what we decided before, is that this is still the scope of this operation, not a different one, styling basic html tags, surely not making ECL work with whatever markup.

emeryro:
Things have to be clarified with Platform, as it seems that the needs are not totally clear to us (what has to be provided? what tags should be styled? ...)

kalinchernev:
Reading through comments here and briefly joining the meeting in the afternoon yesterday, I'm getting the impression about the following:

- Requirements are not fully clear. Specific use cases are necessary to be aligned with the platform 
- The direction we're going is more into creating a new [reset](https://github.com/ec-europa/europa-component-library/blob/v2-dev/src/systems/ec/implementations/vanilla/packages/ec-preset-website/src/ec-preset-website.scss#L5) rather than a preset

Could it be that removing ecl-editor namespace and delivering the result in a separate CSS file serve the purpose?

docs/decisions/006-default-style.md
papegaill: I think Davide's list is pretty much ok `headings`, `global font definition` (paragraphs, links, blockquotes), `lists`, `tables` maybe a default primary for `button`. however I don't think a ckeditor plugin is the way to go. There are other markup outputting from Drupal without going through ckeditor that must be styled, such as a simple link that will need to be preprocessed to add the class and get the right style. Adding classes and changing markup to fields saved in database is not a reliable solution in my opinion, ckeditor already leaves too much flexibility in this direction in my opinion.

docs/decisions/006-default-style.md
papegaill: > Reading through comments here and briefly joining the meeting in the afternoon yesterday, I'm getting the impression about the following:
> 
> * Requirements are not fully clear. Specific use cases are necessary to be aligned with the platform
> * The direction we're going is more into creating a new [reset](https://github.com/ec-europa/europa-component-library/blob/v2-dev/src/systems/ec/implementations/vanilla/packages/ec-preset-website/src/ec-preset-website.scss#L5) rather than a preset
> 
> Could it be that removing ecl-editor namespace and delivering the result in a separate CSS file serve the purpose?

@emeryro kalin's proposal is maybe a good solution, removing ecl-editor namespace and delivered it in a separate css file. we probably need to test the impact on the components, but maybe it will be minimal? what do you think?

emeryro: I'm still convinced that without namespace, we may run into troubles. But we'll see.
At least, what we agree on is that we should deliver this extra css in a separated file

PR #1711, opened 2020-10-14
feat(link): add link component from v2 to v3 

emeryro:
# PR description

port a component from ECL 2 to ECL 3 and document the procedure.
It will be useful also to generate documentation on how to create new components on ECL 3 from scratch.

I updated the file structure and package names to have the same pattern everywhere:
- twig: https://github.com/ec-europa/europa-component-library/tree/docs/FRONT-1691-component-v2-v3/src/implementations/twig/components/link
- vanilla: https://github.com/ec-europa/europa-component-library/tree/docs/FRONT-1691-component-v2-v3/src/implementations/vanilla/components/link
- specs: https://github.com/ec-europa/europa-component-library/tree/docs/FRONT-1691-component-v2-v3/src/specs/components/link

Note that I didn't update existing components yet, so there still be some include using old names. That would be fixed later


kalinchernev (approved):

kalinchernev:
I'll update the rest of the components as well as the twig loader environment settings (failing tests) in FRONT-1658

PR #1712, opened 2020-10-14
chore(storybook): Upgrade to version 6 - FRONT-1483

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/systems/ec/implementations/react/storybook/.storybook/manager.js
```diff
@@ -12,6 +12,7 @@ const theme = create({
 });
 
 addons.setConfig({
+  showRoots: false,
```
kalinchernev: We are currently using the default which is true. Why changing?

src/systems/ec/implementations/react/storybook/.storybook/main.js
```diff
@@ -10,26 +10,35 @@ const stories = [
...
     path.resolve(__dirname, '../../../../../..')
   );
 
-  // Exclude node_modules
-  baseConfig.module.rules[0].exclude = /node_modules/;
+  defaultConfig.module.rules[0].exclude = /node_modules/;
+
+  // Change media dist folder
+  defaultConfig.module.rules[5].query.name =
+    'dist/images/[name].[hash:8].[ext]';
```
kalinchernev, line 30: I am not sure what's the idea behind this change

kalinchernev (commented):
Thanks for the massive amount of updates here!
I've added 2 items which didn't quite fit during the review. In overall, works fine locally and in the preview

src/systems/ec/implementations/react/storybook/.storybook/manager.js
planctus: Ah, this is a mistake, yeah, i moved this according to the docs for storybook 6 but i changed it's value inadvertently, it was set to true.

src/systems/ec/implementations/react/storybook/.storybook/main.js
planctus: it was already in place, just that the rule position in the array changed.

PR #1535, opened 2020-02-04
feat(themes): PoC for multiple themes - FRONT-448

emeryro:
# PR description

DO NOT MERGE; this is just a PoC

See how we can handle multiple themes in ECL EU (and EC), to swap color palette, border radius, ...
An example has been done to EU buttons. Use the "CSS resources" tab in storybook to switch from default display to alternative theme ecl-eu-preset-website-poc.
https://ecl-preview-1535--europa-component-library.netlify.app/playground/eu/?path=/story/components-button--primary
Differences are visible in border radius and shadow on hover
button_theme

This PoC uses global parameters (border-radius and shadow), and apply them to the component (using SASS).
```
@if $ecl-global-border-radius == 1 {
    $border-radius: $ecl-border-radius-4;
  } @else {
    $border-radius: $ecl-border-radius-0;
  }

  @if $ecl-global-shadow == blue {
    $shadow: $ecl-shadow-blue;
  } @else {
    $shadow: $ecl-shadow-none;
  }
```
A preset has been added to handle the new theme

PR #1715, opened 2020-10-20
feat(theme): add theme mechanism - FRONT-1628

emeryro:
# PR description

Apply theme system introduced in https://github.com/ec-europa/europa-component-library/pull/1535 to curent dev branch.

demo: https://ecl-preview-1715--europa-component-library.netlify.app/playground/ec/?path=/story/components-button--primary
(use the "CSS resources" tab to switch theme)

Notes:
- I kept the old presets for now (editor, full and website). I don't know yet what will be needed here or not
- only border radius has been added as global (theme) parameter. It is visible on the button. More to come when we have the dedicated specs


src/themes/ec-core/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@ecl/theme-ec-core",
```
kalinchernev, line 2: This package uses `@ecl/theme-default`. I'd suggest we add it in the dependencies list.

src/themes/eu-core/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@ecl/theme-eu-core",
```
kalinchernev, line 2: Same suggestion for adding a dependency to the default theme for this package.

src/implementations/vanilla/packages/ec-preset-website/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@ecl/ec-preset-website",
+  "name": "@ecl/preset-website",
```
kalinchernev, line 2: Let's move the website/editor/etc. presets them to `presets` folder as well please.

src/themes/default/package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "@ecl/ec-theme-default",
+  "name": "@ecl/theme-default",
```
kalinchernev: Just a idea coming to mind: `theme-dev` instead, in order to match theme to preset in a way.

src/implementations/twig/storybook/.storybook/preview-head.html
```diff
@@ -30,15 +30,8 @@
   } else {
     // Manually inject styles
     var head = document.head || document.getElementsByTagName('head')[0];
-    head.appendChild(
```
kalinchernev: Not seeing any problems with this change so far, however it's a bit different target I think. Shall we maybe target the website preset here anyway? I think there could be some reason on this being loaded for later when there compositions, although not sure.

src/implementations/vanilla/presets/ec-core/ecl-builder.config.js
```diff
@@ -0,0 +1,69 @@
...
+        includePaths,
+        sourceMap: isProd ? 'file' : true,
+      },
+    },
+  ],
+  copy: [
+    {
+      from: path.resolve(
+        nodeModules,
+        '@ecl/ec-component-form-feedback-message/images'
```
kalinchernev: Please revise the builder config files. It seems they still contain some irrelevant instructions.

src/implementations/vanilla/presets/eu-core/ecl-builder.config.js
```diff
@@ -0,0 +1,69 @@
...
+        includePaths,
+        sourceMap: isProd ? 'file' : true,
+      },
+    },
+  ],
+  copy: [
+    {
+      from: path.resolve(
+        nodeModules,
+        '@ecl/ec-component-form-feedback-message/images'
```
kalinchernev: just a marker for a revision

kalinchernev (changes requested):
Thanks for making this important step towards themes!
Minor comments follow

kalinchernev (approved):
thanks :+1:

PR #1718, opened 2020-10-21
chore: use dart sass - FRONT-1718

kalinchernev:
Changes applied throughout the whole repository, also in the website.

As a first iteration, I've used default behavior meaning that module names are following the file names. This leads to more verbose module inclusion and actual implementation based on maps might not be seen as an improvement cosmetically.

The base package has gotten so thin and dependent on the default theme, that we can consider migrating its code to another package and removing it.

Please test carefully as changes touch upon almost all styling aspect of the project.


emeryro (approved):
That's working fine, thanks for this!
We just have to fix the conflicts with the theme update. I'll also update my documentation on migration from v2 to v3.

ps: just for fun, the PR number is exactly the same as the ticket number

emeryro (approved):

PR #1720, opened 2020-10-23
chore: refactor naming patterns - FRONT-1735

kalinchernev:
Highlights:

- renamed: vanilla+twig components, +resources
- layout has been left in the packages folder, not sure it's either worth moving to components neither to a new folder
- presets have been moved a few folders up in order to be next to the themes
- prefixes removed from tokens because of the @use modularity
- compliance checker moved all in and renamed in a way to simplify migration of other components without the compliance hassle 
- website + scripts: updated
- migration document in teams has been updated

I have left out ec-base. I'd suggest we discuss this package and find a way to remove in favor of other packages.


docs/decisions/005-themes.md
```diff
@@ -78,7 +78,7 @@ Resources are: logo, favicons, icons and other similar types of assets which are
 
 Current organization of v2.
 
-- `@ecl/ec-resources-logo`
+- `@ecl/resources-ec-logo`
 - `@ecl/eu-resources-logo`
```
emeryro, line 82: we will have to handle eu resources too, at least for the logo.
I guess we can have a dedicated task for this

emeryro (approved):
Thanks for the cleaning, that's much better now.

Here are a few topics that will have to be addressed, probably in dedicated tasks:
- what do we do with ec-base?
- what do we do with old presets (ec-preset-full, ec-preset-website)?
- what do we do with editor package and preset?
- how should we handle resources?

kalinchernev:
> what do we do with ec-base?

FRONT-1748 created

> what do we do with old presets (ec-preset-full, ec-preset-website)?

FRONT-1749 created

> what do we do with editor package and preset?

FRONT-1750 created

> how should we handle resources?

Nothing new in terms of approach I think. The "resources" folder contains these packages.
Noticed FRONT-1745 has already been created for the eu logo.

PR #1724, opened 2020-10-27
chore: remove ec-base - FRONT-1748

kalinchernev:
- `import-once` is no longer necessary since import statements with `@use` solve this issue (see https://sass-lang.com/documentation/at-rules/use)

> Any styles loaded this way will be included exactly once in the compiled CSS output, no matter how many times those styles are loaded.

- breakpoints have been migrated to the layout package since it's the only one using these definitions
- ecl- prefixing in breakpoints module has been removed as unnecessary
- js helpers have been removed, because not used at the moment; when js behaviors are back we can re-evaluate whether a query util is actually needed; the other for the file size is used only for file download
- ec-base has been removed
- themes refactored to be accessible without `'/index'` and relatives `'./'` removed
- components' mixins removed in order to make their private members [really private](https://sass-lang.com/documentation/at-rules/use#private-members)


planctus (approved):
Great that we moved to dart and we have now @use at our disposal.
It's not part of this pull request, but i believe that the repeated usage of map-get to retrieve each variable value can and should be improved by using functions, or modules as pointed out here: https://www.sitepoint.com/using-sass-maps/

PR #1727, opened 2020-10-29
feat: migrate radio - FRONT-1739

kalinchernev:
The following code has been removed during the migration

```scss
  // DEPRECATED
  .ecl-form-label--hidden {
    border: 0;
    clip: rect(0, 0, 0, 0);
    height: 1px;
    margin: -1px;
    overflow: hidden;
    padding: 0;
    position: absolute;
    visibility: visible;
    white-space: nowrap;
    width: 1px;
  }
```


src/implementations/vanilla/components/form-radio/form-radio.scss
```diff
@@ -0,0 +1,206 @@
...
+@use '@ecl/theme-dev/maps/spacing';
+
+$_label-color: map.get(color.$color, 'grey-100');
+$_label-color-disabled: map.get(color.$color, 'text');
+$_box-size: 1.25rem;
+$_box-border-width: 2px;
+$_box-border-width-checked: 7px;
+$_box-border-color: map.get(color.$color, 'text');
+$_box-border-color-disabled: map.get(color.$color, 'grey-50');
+$_box-border-color-checked: map.get(color.$color, 'grey-100');
```
emeryro: This colour seems to be wrong (same for the next line).
It is blue in current version

src/implementations/vanilla/components/form-radio/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@ecl/vanilla-component-form-radio",
```
emeryro: do we need to keep the 'form' in the package name? It isn't the case for twig and specs packages. For the checkboxes it had been removed

emeryro (changes requested):
In my PR for checkboxes, I merged several css related to forms in one package (I didn't see the point of keeping them separated): https://github.com/ec-europa/europa-component-library/pull/1723/files#diff-406a9c851ccdb2d94cee49332d74ab0273cb876962ddbdd7953ed67e34c6f84b
So I guess we should align here (sorry if that was not clear enough)

I also noticed a strange behavior on the radio: when on the [website demo](https://ecl-preview-1727--europa-component-library.netlify.app/ec/components/forms/radio/code/), if we select the second option (Belgium in the first example), we have to click twice to get it really checked. That's not the case on the current version

src/implementations/vanilla/components/form-radio/package.json
kalinchernev: I removed the "form" part.

src/implementations/vanilla/components/form-radio/form-radio.scss
kalinchernev: My bad. Corrected.

kalinchernev:
I've been debugging the issue with the radio button demo on the website and came to the conclusion that it's not related to changes in this pull request. I do not have an idea about the issue yet, here are the things I did:

- double-checked the CSS of the component and it seems ok in current state of the pull request here
- double-checked the order of files in the preset and I had made an error again with the icon which I fixed, but it's not the order for the radio on the demo
- the storybook demo works fine with the css, but the website demo is not
- double-checked the specs and didn't find issues with the updated files in demo

when i'm on the current v2, the checked attribute is fine, but on the v3 it's not.

```js
var input = document.getElementById('radio-default-1')
input.checked
```

With one and the same visible markup on the websites of v2 the result is `true`, whereas `false` on v3. It's something with how the component gets rendered. Reminds me of the select multiple case when the selected attribute got handled differently in 2 cases. Here in the radio we don't have JS so I don't know how to handle this exception.

src/implementations/twig/components/radio/README.md
```diff
@@ -0,0 +1,79 @@
+# ECL-Twig Radio component
```
emeryro:
```suggestion
# ECL Radio component
```

emeryro (approved):
that's good for me, just the small point below and we can merge.
I also created https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1761 for the check issue

PR #1729, opened 2020-10-30
feat: migrate text area - FRONT-1743

kalinchernev:


src/implementations/twig/components/text-area/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "license": "EUPL-1.1",
+  "version": "3.0.0",
+  "description": "ECL Text Area",
+  "private": true,
+  "devDependencies": {
+    "@ecl/vanilla-component-form": "^3.0.0",
+    "@ecl/vanilla-component-text-area": "^3.0.0",
+    "@ecl/specs-component-text-area": "^3.0.0"
+  },
+  "repository": {
```
emeryro, line 13: I guess we can remove all what is below, as it has been done on other components (or replace it with something non ecl-twig related).
I think it's the case for the text input too

emeryro (changes requested):

emeryro (approved):

PR #1730, opened 2020-11-03
feat(table): Migrate component - front-1741

planctus:
# PR description

In this pr the auto-init and the dom helpers package have been reintroduced to make the js functionality work.
It is not entirely clear to me the reason why we need to call the auto-init here to make the js work, in ecl-twig (like it is described in the docs) setting the data-ecl-auto-init attr triggers the js while here i could not see it working like that.
In order to be able to run a script on a specific story a package has been introduced, it works flawlessly but i'm not necessarily sure this is the only or the best way, other many attempts failed so i picked up this working solution.
In a late commit functions have been added to the scss stack to improve and simplify the usage of our maps, it has been only applied to the table component but it can be then used as the standard way if we like it. 

The preview-head.html has been reintroduced to load the js file as it was done in v2, since we only have one storybook instance for the moment (i don't think we want the switch between EC and EU to happen as a simple "theme" switch, i would still go for to parallel instances) it is loading ec-core.js only, though.

## QA Checklist

- [ ] I have provided tests


src/implementations/twig/storybook/.storybook/preview-head.html
```diff
@@ -0,0 +1,39 @@
+<style>
```
emeryro, line 1: If I'm not wrong, this file had been removed previously as it was not used.
Do we need it back?

src/implementations/vanilla/components/table/table-print.scss
```diff
@@ -0,0 +1,38 @@
+/**
+ * Table component
+ * @define table
+ */
+
+// Import base
+@use '@ecl/theme-dev/functions/colorfn';
+@use '@ecl/theme-dev/functions/spacingfn';
+@use '@ecl/theme-dev/functions/typographyfn';
```
emeryro: I don't think we need to have the 'fn' at the end of the name.
We could just keep 'spacing', 'typography' and 'color'

src/implementations/twig/components/table/table.story.js
```diff
@@ -0,0 +1,113 @@
...
+
+export const Multi = () => table(prepareTable(dataMulti));
+
+Multi.storyName = 'multi header';
+Multi.parameters = { notes: { markdown: notes, json: dataMulti } };
+
+export const WithRowExtraAttributes = () =>
+  table(prepareTable(dataDefault, true));
+
+WithRowExtraAttributes.storyName = 'With row extra attributes';
```
emeryro: what is this story about? I'm not sure it is needed

emeryro (changes requested):
Thanks for this.
I'm fine with the use of sass functions, but if we go this way it has to be used everywhere.
I just have a few questions/remarks below.

Also I noticed that the responsive behavior doesn't work on storybook (no way to display the mobile version). It shouldn't be linked to this PR, but we will have to fix it.

src/implementations/twig/storybook/.storybook/preview-head.html
planctus: it loads the js, i added a small text about it in the pr description

src/implementations/vanilla/components/table/table-print.scss
planctus: two modules with the same names generate an error, they were initially named as you suggested, this was later adapted, i wanted to name them like color-fn and the dash was triggering a false linter issue, then i moved them to colorFn and also this triggered a linter issue.. :)

src/implementations/twig/components/table/table.story.js
planctus: This was to show the additional extra_attributes per row that were added based on the request by the platform in ecl-twig. We can remove it from here, sure..

planctus:
About the functions i agree, that is also what i wrote in the description of this pr, if we like it we apply that everywhere, but let's do it in an "alignment" ticket ;)

src/implementations/twig/storybook/.storybook/preview-head.html
kalinchernev: This file was used mainly for this 

```js
  if (
    inIframe() &&
    window.frameElement &&
    window.frameElement.id === 'storybook-preview-iframe'
  ) {
    // Let CSS Resources inject the styles
  } else {
    // Manually inject styles
    var head = document.head || document.getElementsByTagName('head')[0];
    head.appendChild(createLink('./styles/ec-core.css', 'screen'));
    head.appendChild(createLink('./styles/eu-core.css', 'screen'));
    head.appendChild(createLink('./styles/ec-core-print.css', 'print'));
    head.appendChild(createLink('./styles/eu-core-print.css', 'print'));
  }
```
Which will add styles in the fullscreen iframe `playground/ec/iframe.html?id=components-table--sortable&viewMode=story`  version of the story. I personally doubt anyone is using these fullscreen iframes and thus the deletion of the template.

kalinchernev: With regards to loading ECL and it's autoinit: ok, let's have the global loading back. The select multiple is the only component with non-default init, I'll find a way to destroy and re-init there.

Please remove the new addon and take the example preview-head from ecl-twig. I tried this and it worked

```html
<script type="text/javascript" src="./scripts/ec-core.js"></script>
<script>
  // https://github.com/storybookjs/storybook/issues/6113#issuecomment-473965255
  function runOnPageChange() {
    ECL.autoInit();
  }

  document.addEventListener(
    'DOMContentLoaded',
    function () {
      const callback = function (mutationsList) {
        for (let i = 0, len = mutationsList.length; i < len; i++) {
          if (mutationsList[i].type == 'childList') {
            runOnPageChange();
            break;
          }
        }
      };
      const observer = new MutationObserver(callback);
      const config = { childList: true, subtree: false };
      observer.observe(document.getElementById('root'), config);
    },
    false
  );
</script>
```

If you want to take also the tippy for the compliance checker - ok for me.

The drawbacks of adding this new addon is loading scripts through a third party unnecessarily. When the init breaks, we'll be able to handle it on our side. I also don't see any advantage adding yet another new addon when we have a working solution in the ecl-twig

src/implementations/vanilla/packages/auto-init/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "3.0.0",
+  "description": "ECL Auto Init",
+  "main": "index.js",
+  "module": "index.js",
+  "dependencies": {
+    "@ecl/theme-dev": "^3.0.0"
+  },
+  "publishConfig": {
```
kalinchernev: Privatize please

src/implementations/vanilla/packages/auto-init/README.md
```diff
@@ -0,0 +1,87 @@
...
+</div>
+```
+
+Once `ECL.autoInit()` is called, you can access the component instance via an `ECLMessage` property on that element.
+
+```js
+document.querySelector('[data-ecl-message]').ECLMessage.handleClickOnClose();
+```
+
+## Migration
```
kalinchernev: Remove this please: will become obsolete note and irrelevant at the same time. (no react any more see)

src/implementations/vanilla/components/table/table-print.scss
kalinchernev: Regarding these functions: I don't see the benefit of having them to be honest. Neither the syntax is way better, nor we avoid the boilerplate of having color.$color for example. If we can remove this somehow, then it's a real saver in my view.

Can we maybe have the signature different like this: `color('grey-25')` vs `color('grey-25', 'bg')`? I.e. default is color/color, font/font, and sub-modules are second param? Typography and layout don't fit in in this pattern but if we could have it for the others it'll be great i think

If you guys prefer to have them anyway, I think the namespaces can help https://sass-lang.com/documentation/at-rules/use#choosing-a-namespace for avoiding naming collissions.

kalinchernev: It'll be helpful if we could extract this work on a separate track and pull request in order to not have inconsistencies or blocked migration pull requests.

src/implementations/twig/components/table/README.md
```diff
@@ -0,0 +1,153 @@
...
+- **"extra_attributes"** (optional) (array) (default: []) Extra attributes
+  - "name" (string) Attribute name, eg. 'data-test'
+  - "value" (string) Attribute value, eg: 'data-test-1'
+- **"\_compliance\_"** (boolean) (default: false) Activates debug
+
+### Example:
+
+<!-- prettier-ignore -->
+```twig
+{% include '@ecl/table/twig-component-table.html.twig' with {
```
kalinchernev: Please correct template name here

src/implementations/twig/components/table/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@ecl/twig-component-table",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "3.0.0",
+  "description": "ECL Table",
+  "devDependencies": {
+    "@ecl/specs-component-table": "^3.0.0"
```
kalinchernev: Please add the vanilla component as a dependency

src/implementations/vanilla/components/table/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@ecl/vanilla-component-table",
```
kalinchernev, line 2: Please add this package in the list of deps in `src/presets/dev/package.json`

src/tools/helpers/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@ecl/helpers",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "3.0.0",
+  "description": "ECL helpers",
+  "main": "dom.js",
+  "module": "dom.js",
+  "publishConfig": {
```
kalinchernev: Privatize please

kalinchernev (changes requested):
Suggestions added inline

src/implementations/twig/storybook/.storybook/preview-head.html
planctus: I applied the same approach as in ecl-twig. Mind the fact that with this, the first page loaded will not have the autoInit called and we would need to call it also outside of this function.
Did not take all the scripts and styles for the compliance tool because in any case its markup is based on utilities that do not exist yet in next-v3 so they would be pretty useless at the moment. We will reintroduce them once we will take care of this.

src/implementations/vanilla/components/table/table-print.scss
planctus: i reverted these changes for now, let's move them in a separate track.

src/implementations/twig/storybook/.storybook/preview-head.html
kalinchernev: What I meant in my previous comment was that the preview-head you had was ok, and you can only add the code after 
```
<script type="text/javascript" src="./scripts/ec-core.js"></script>
```
which is instantiating the ECL module. I did not mean that you should remove the css part, nor include the svg correction script. It's already here https://github.com/ec-europa/europa-component-library/blob/c546d79e2a4e3dbadfeb6a761b03ee73bb528870/src/implementations/twig/storybook/.storybook/ECL.jsx

Here's the contents of the file directly in order to avoid confusion

```html
<style>
  .sb-show-nopreview {
    display: none;
  }
</style>
<script>
  function inIframe() {
    try {
      return window.self !== window.top;
    } catch (e) {
      return false;
    }
  }

  function createLink(href, media) {
    var link = document.createElement('link');
    link.rel = 'stylesheet';
    link.type = 'text/css';
    link.href = href;
    link.media = media || 'all';
    return link;
  }

  if (
    inIframe() &&
    window.frameElement &&
    window.frameElement.id === 'storybook-preview-iframe'
  ) {
    // Let CSS Resources inject the styles
  } else {
    // Manually inject styles
    var head = document.head || document.getElementsByTagName('head')[0];
    head.appendChild(createLink('./styles/ec-core.css', 'screen'));
    head.appendChild(createLink('./styles/eu-core.css', 'screen'));
    head.appendChild(createLink('./styles/ec-core-print.css', 'print'));
    head.appendChild(createLink('./styles/eu-core-print.css', 'print'));
  }
</script>
<script type="text/javascript" src="./scripts/ec-core.js"></script>
<script>
  // https://github.com/storybookjs/storybook/issues/6113#issuecomment-473965255
  function runOnPageChange() {
    ECL.autoInit();
  }

  document.addEventListener(
    'DOMContentLoaded',
    function () {
      runOnPageChange();
      const callback = function (mutationsList) {
        for (let i = 0, len = mutationsList.length; i < len; i++) {
          if (mutationsList[i].type == 'childList') {
            runOnPageChange();
            break;
          }
        }
      };
      const observer = new MutationObserver(callback);
      const config = { childList: true, subtree: false };
      observer.observe(document.getElementById('root'), config);
    },
    false
  );
</script>
```

I've added `runOnPageChange();` invocation after dom is ready for running the ecl init on first run.

kalinchernev:
@planctus thanks for the updates here! I allowed myself to carry out the small left over tweaks in order to be able to use your JS solution in the select multiple case.

PR #1732, opened 2020-11-03
chore(storybook): ECL themes instead of cssresources

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/implementations/twig/storybook/.storybook/manager-head.html
+++ b/src/implementations/twig/storybook/.storybook/manager-head.html
@@ -51,14 +51,3 @@
   type="text/javascript"
   defer
 ></script>
-<style>
-  #tabbutton-css-resources {
-    visibility: hidden;
-  }
-  #tabbutton-css-resources:after {
-    content: 'ECL Themes';
-    visibility: visible;
-    display: block;
-    position: absolute;
-  }
-</style>
--- a/src/implementations/twig/storybook/.storybook/manager-head.html
+++ b/src/implementations/twig/storybook/.storybook/manager-head.html
@@ -51,3 +51,14 @@
   type="text/javascript"
   defer
 ></script>
+<style>
+  #tabbutton-css-resources {
+    visibility: hidden;
+  }
+  #tabbutton-css-resources:after {
+    content: 'ECL Themes';
+    visibility: visible;
+    display: block;
+    position: absolute;
+  }
+</style>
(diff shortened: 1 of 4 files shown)


kalinchernev:
To me, CSS resources is the right name for the tab. It's good to have it reordered.

emeryro (approved):
This is clearly a hack (using CSS to change the content), but as we don't have better solution for now, let's go with it

src/implementations/twig/storybook/.storybook/manager-head.html
```diff
@@ -51,3 +51,14 @@
   type="text/javascript"
   defer
 ></script>
+<style>
```
kalinchernev: I'd suggest we avoid making this change as it's not really reflecting themes, but css resources.
Have you checked https://github.com/tonai/storybook-addon-themes? see https://github.com/tonai/storybook-addon-themes/issues/21 which will probably be more relevant in our html case

planctus: i'm confused, before you said that cssresources is the right "title" for this tab, here it seems you are suggesting to use a different approach regarding the theme switcher. Personally speaking i only spent a bit of time in the past looking for alternatives to the current approach we are using and i haven't found much, i'm not surely enthusiast about the UI provided by cssresources nor the user experience that we will offer. 

As far as we are using cssresources, though, it looks to me that its label "css resources" doesn't really offer a clear indication of the relevance it would have in our system, therefore here we only perform its replacement into a more informative "ECL themes", since we are going to have a theme based system and this would be the interface the user will have to switch between them. 

If there is no agreement even about the opportunity to rename this tab, i don't mind, let's close this, If you believe there is a better overall approach for a  "theme switcher" in storybook i would be glad to entirely replace cssresources, indeed.

kalinchernev: Summarizing my previous points to clear up confusion:
- changes in the pull request are not fully ok for me
- my suggestion is to leave the name of the tab to css resources (revert css change)
- the other updates seem fine
- i suggested another addon for solving what you're actually trying to solve: expose "themes" feature

src/implementations/twig/storybook/.storybook/preview.js
```diff
@@ -10,16 +10,25 @@ addParameters({
...
       picked: false,
+      hideCode: true,
+    },
+    {
+      id: 'eu-core-print',
+      code: `<link rel="stylesheet" type="text/css" href="./styles/eu-core-print.css" />`,
+      picked: false,
+      hideCode: true,
     },
     {
```
kalinchernev: Let's remove this if there is no need of it. I checked for any implementations using this id, but didn't find any.

kalinchernev (changes requested):
https://github.com/ec-europa/europa-component-library/pull/1732/files#r517368610

kalinchernev (approved):

PR #1733, opened 2020-11-03
feat(date-block): Migrating component -  front-1757

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/implementations/twig/components/date-block/README.md
```diff
@@ -0,0 +1,45 @@
+# ECL-Twig Date block
```
emeryro:
```suggestion
# ECL Date block
```

src/implementations/twig/components/date-block/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@ecl/twig-component-date-block",
+  "private": true,
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "3.0.0",
+  "description": "ECL Date Block",
+  "dependencies": {
+    "@ecl/resources-ec-icons": "3.0.0"
```
emeryro: I don't think we need this dependency

src/implementations/vanilla/components/date-block/date-block.scss
```diff
@@ -0,0 +1,186 @@
...
+$separator-color-default: map.get(color.$color, 'grey-25');
+$year-background-default: map.get(color.$color, 'grey-10');
+$year-text-color-default: map.get(color.$color, 'grey');
+$background-ongoing: map.get(color.$color, 'yellow');
+$text-color-ongoing: map.get(color.$color, 'grey');
+$separator-color-ongoing: map.get(color.$color, 'grey-25');
+$year-background-ongoing: map.get(color.$color, 'grey-10');
+$year-text-color-ongoing: map.get(color.$color, 'grey');
+$background-cancelled: map.get(color.$color, 'grey-10');
+$text-color-cancelled: map.get(color.$color, 'white');
```
emeryro: this should be grey

src/implementations/vanilla/components/date-block/date-block.scss
```diff
@@ -0,0 +1,186 @@
...
+  .ecl-date-block__month {
+    padding-bottom: map.get(spacing.$spacing, 'xs');
+  }
+
+  .ecl-date-block__year {
+    padding-bottom: map.get(spacing.$spacing, 'xs');
+    padding-top: map.get(spacing.$spacing, 'xs');
+  }
+
+  .ecl-date-block__daytime {
```
emeryro, line 117: where does this css come from? I don't see it in the current implementation

emeryro (changes requested):

src/implementations/vanilla/components/date-block/date-block.scss
planctus: Ah, this is for the "sr-only"  span, in the current implementation is using an utility class, i copied the css properties of the "ecl-u-sr-only" into this selector that i added, not only because we don't have utilities yet, but also because i believe those should not be used in the component markup.

emeryro: I see... I agree that components should not use utilities directly. I checked quickly, there are only a few components in this case, and we will update them when migrating.
But we should not use `!important` here (we shouldn't have any override issue inside a component)

src/implementations/vanilla/components/date-block/date-block.scss
```diff
@@ -87,7 +87,7 @@ $circle-width: 0.6rem;
   padding-bottom: map.get(spacing.$spacing, '2xs');
   padding-left: map.get(spacing.$spacing, '2xs');
   padding-right: map.get(spacing.$spacing, '2xs');
-  text-decoration: none !important;
+  text-decoration: none;
```
emeryro: it seems that this `important` was needed; there is an underline on the component now.
So we should bring it back, or increase the specificity of the rule

emeryro (changes requested):

emeryro:
Also the test snapshot should be updated

src/implementations/vanilla/components/date-block/date-block.scss
```diff
@@ -0,0 +1,186 @@
...
+    top: -#{$circle-width / 2};
+    width: $circle-width;
+  }
+}
+
+.ecl-date-block__day {
+  color: $text-color-default;
+  font: map.get(typography.$font, '2xl');
+  font-weight: map.get(typography.$font-weight, 'bold');
+  padding-left: map.get(typography.$font, 'xs');
```
emeryro:
```suggestion
  padding-left: map.get(spacing.$spacing, 'xs');
```

src/implementations/vanilla/components/date-block/date-block.scss
```diff
@@ -0,0 +1,186 @@
...
+ */
+
+// Import base
+@use "sass:map";
+@use '@ecl/theme-dev/maps/color';
+@use '@ecl/theme-dev/maps/spacing';
+@use '@ecl/theme-dev/maps/typography';
+@use '@ecl/layout-grid/mixins/breakpoints';
+
+$background-default: map.get(color.$color, 'blue');
```
emeryro: I didn't notice it before, but we are using a new naming convention for local variables defined in the components: it should start with an underscore (`$_background-default`)
I think it is part of the migration document

emeryro (changes requested):

emeryro (approved):

PR #1734, opened 2020-11-04
feat(tags): migrate tags component - FRONT-1756

emeryro:
# PR description

Migrate tag component


src/implementations/twig/components/tag/.npmignore
```diff
@@ -0,0 +1,3 @@
+__snapshots__
+demo
```
planctus: i noticed that we are using the .npmignore from ecl-twig, we don't have the demo folder in ECL, though ;)
```suggestion
```

src/implementations/twig/components/tag/tag.story.js
```diff
@@ -0,0 +1,67 @@
...
+
+  return data;
+};
+
+export default {
+  title: 'Components/Tag',
+  decorators: [withKnobs, withNotes, withCode],
+};
+
+export const Display = () => tag(prepareTag(dataDisplay, false, true));
```
planctus:
```suggestion
export const Display = () => tag(prepareTag(dataDisplay));
```
We have this small bug in ecl-twig, we expose the aria_label prop for the display type in a knob, but the template only applies that to the removable.

src/implementations/twig/components/tag/tag.story.js
```diff
@@ -0,0 +1,67 @@
...
+    data.tag.path = text('tag.path', data.tag.path, tabLabels.required);
+  }
+  if (aria) {
+    data.tag.aria_label = text(
+      'tag.aria_label',
+      data.tag.aria_label,
+      tabLabels.required
+    );
+  }
+  if (aria) {
```
planctus: Maybe we can remove this duplicated if (aria) here

src/implementations/vanilla/components/tag/tag.scss
```diff
@@ -0,0 +1,96 @@
...
+@use '@ecl/theme-dev/maps/color';
+@use '@ecl/theme-dev/maps/icon';
+@use '@ecl/theme-dev/maps/spacing';
+@use '@ecl/theme-dev/maps/typography';
+
+.ecl-tag {
+  align-items: center;
+  background-color: map.get(color.$color, 'grey-15');
+  border-radius: (
+      2 * map.get(spacing.$spacing, 'xs') +
```
planctus, line 16: why this indentation here? was it formatted like this by the linter?

planctus (approved):
There are a couple of suggestions, but it looks good to me.

src/implementations/vanilla/components/tag/tag.scss
emeryro: indeed, I didn't change it, so that's the linter preference

PR #1737, opened 2020-11-05
feat: add select - FRONT-1740

kalinchernev:


src/implementations/twig/components/select/.npmignore
```diff
@@ -0,0 +1,3 @@
+__snapshots__
+demo
```
emeryro:
```suggestion
```
demo is not used anymore

src/implementations/twig/components/select/select.html.twig
```diff
@@ -0,0 +1,201 @@
...
+    </div>
+  {% endif %}
+
+  {% if _invalid and _invalid_text is not empty %}
+    <div class="ecl-feedback-message">
+      {%- block invalid_text _invalid_text|raw -%}
+    </div>
+  {% endif %}
+
+  <div class="ecl-select__container
```
emeryro, line 125: it seems that the twig template is still using the deprecated code for invalid select (it is visible as you removed deprecated CSS)
ECL code
image

Twig code
image

We should update the twig template to use latest markup. This could be done in this PR or in another one, as you prefer

kalinchernev:
I tried to correct the twig file at https://github.com/ec-europa/europa-component-library/pull/1737/commits/fbc6a8e7255c4c568ae42c5961de2ff75a7da45c in this pull request
@planctus can you please check if it's ok?

emeryro (approved):
that's good for me

PR #1738, opened 2020-11-05
chore: consolidate js utilities

kalinchernev:
Let's have all tiny utils in one place instead of 3. 
Overview:
- 1 package, 3 tiny sub-packages: autoinit, polyfills, generics (previous dom)
- import generics selectively
- polyfills and autoinit used once in dev preset in the same way
- sub-packages are not exported from main index in order to not overload unneeded


emeryro:
That's indeed cleaner that way, but I'm a little concerned about the naming.
We will have CSS utilities too at some point (usable by clients too, which are used to this naming), so maybe we could use another name for these js helpers.
It could be something like 'utilities-js', 'toolbox', 'helpers', or any other name

kalinchernev:
Renamed to `@ecl/dom-utils` which is placed in the toolbox (`tools` folder) next to `test-utils` and `story-utils`.
I don't mind any other suggestions

emeryro (approved):

PR #1740, opened 2020-11-05
chore: using sass functions - FRONT-1783

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/implementations/vanilla/components/accordion/accordion-print.scss
+++ b/src/implementations/vanilla/components/accordion/accordion-print.scss
@@ -5,12 +5,12 @@
 
 // Import base
 @use "sass:map";
-@use '@ecl/theme-dev/maps/color';
-@use '@ecl/theme-dev/maps/spacing';
-@use '@ecl/theme-dev/maps/typography';
+@use '@ecl/theme-dev/functions/colorfn' as *;
+@use '@ecl/theme-dev/functions/spacingfn' as *;
+@use '@ecl/theme-dev/functions/typographyfn' as *;
 
 .ecl-accordion__item {
-  margin-top: map.get(spacing.$spacing-print, 'l');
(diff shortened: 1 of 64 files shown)


emeryro (commented):
Bringing back `ecl-` namespace into the code feels like a step back...
I'd suggest using other name for the functions, like `get-spacing()` for instance, or maybe just `spacing()`, as we don't really need extra namespace here.

Also, that may be just me, but I feel that using `*` everywhere somehow goes against the idea of module system introduced in sass. Having something like `typography.line-height('m')` feels more robust to me (even if it's a little longer to type, I agree)

kalinchernev:
> Bringing back ecl- namespace into the code feels like a step back...

True. `ecl-color('grey');` is almost `ecl-color-grey` which means we go back on variables vs maps.

About naming I've shared my preference before https://github.com/ec-europa/europa-component-library/pull/1730#discussion_r517222130

It matches to 

> just spacing(), as we don't really need extra namespace here.

Regarding imports

> Also, that may be just me, but I feel that using * everywhere somehow goes against the idea of module system introduced in sass.

Yes. Similarly to naming as variables we go back to the previous import.

planctus:
In all honesty, i think this namespacing argument is not really the core of what's in here, not sure i understand any of the concerns regarding it, but at the same time, although i find that ecl should be in place to properly identify and distinguish our construct from any other we might use, i have no issues in removing it, if that is what would make the removal of thousand of map-get calls in our sass component definitions a valuable improvement. If i get it correctly you would both prefer having simply  `color('grey')`, as you can imagine this is how i started and then decided that it was too generic because these are "ecl" colors. If this is your preference i will update this pr reverting it to a previous stage but if you are simply fine with the current approach let's go with it, i don't necessarily have to agree with it or like it.

PR #1744, opened 2020-11-10
feat: migrate css utilities - FRONT-1744

kalinchernev:
Notes:

- moved into a new "utilities" folder, and removed ecl-utilities prefix
- editor moved to the new utilities folder because of [this](https://ec.europa.eu/component-library/playground/ec/?path=/story/utilities-editor--sample) in v2
- layout folder created within the vanilla implementation, matching previous react one
- added a new `styled` dom utility to reduce time spent for react to html refactoring

Not done on purpose:
- utils marked as "deprecated" are not migrated
- although print stylesheets rarely differ from their main scss files, I've left them as-is for consistency (i.e. dev-print not including main files, but {main}-print files)
- utilities stories in twig implementation do not use twig templates but string literals; these are not considered packages as well
- migrated react layout grid is plain HTML instead of twig templates; I personally don't see a gain having dynamic templates for a demo in the same way as react


src/implementations/vanilla/utilities/typography/typography.scss
```diff
@@ -0,0 +1,173 @@
+@use "sass:map";
+@use "@ecl/theme-dev/maps/typography";
+@use "@ecl/theme-dev/maps/color";
+@use "@ecl/theme-dev/maps/layout";
+@use '@ecl/vanilla-layout-grid/mixins/breakpoints';
+
+/*
+   * Headings
+   */
+.ecl-u-type-heading-1 {
```
emeryro, line 10: I think that the font values for h1 to h3 are wrong here.
Could you double check?

src/implementations/vanilla/utilities/aria/README.md
```diff
@@ -0,0 +1,3 @@
+# ECL Utility Aria
```
emeryro: I'm not sure that these utlites are needed anymore. I don't think that they are used, and there is also no demo of it.
So I'd suggest to remove them, and if they appear to be needed later we can bring them back, with a demo

src/implementations/twig/storybook/package.json
```diff
@@ -7,7 +7,7 @@
   "description": "Storybook EC Twig",
   "scripts": {
     "build": "build-storybook -c .storybook -o build",
-    "start": "start-storybook -p 6006 -s ../../../presets/ec-core/dist,../../../presets/eu-core/dist -c .storybook",
+    "start": "start-storybook -p 6006 -s ../../../presets/ec-core/dist,../../../presets/eu-core/dist,../../../presets/ec-preset-editor/dist -c .storybook",
```
emeryro: the status of this editor preset is still unclear to me, as it is redondent with the concept of default html styling (see https://github.com/ec-europa/europa-component-library/pull/1707)
We can have it as is for now, but we still need to see how to deal with it

src/implementations/vanilla/utilities/clearfix/README.md
```diff
@@ -0,0 +1,3 @@
+# ECL Utility Clearfix
```
emeryro, line 1: missing website documentation

src/implementations/vanilla/utilities/disablescroll/README.md
```diff
@@ -0,0 +1,3 @@
+# ECL Utility Disable scroll
```
emeryro, line 1: missing website documentation

src/implementations/vanilla/utilities/float/README.md
```diff
@@ -0,0 +1,11 @@
+# ECL Float utilities
```
emeryro, line 1: missing website documentation

src/implementations/vanilla/utilities/z-index/README.md
```diff
@@ -0,0 +1,11 @@
+# ECL Utility Z-index
```
emeryro, line 1: missing website documentation

src/implementations/vanilla/utilities/screen-reader/README.md
```diff
@@ -0,0 +1,6 @@
+# ECL Utility Screen readers
```
emeryro, line 1: missing website documentation

emeryro (changes requested):
Thanks for this, it looks nice!
I added a few points below, some of them may be done in separated tasks if needed (mainly documentation on the website)

One remark: on the website we could group the two layout utilities (grid and stack) into a sub category (similar to forms for components)

src/implementations/twig/storybook/package.json
kalinchernev: Removed in order to avoid future confusions until there is a plan for it

src/implementations/vanilla/utilities/aria/README.md
kalinchernev: Removed

src/implementations/vanilla/utilities/typography/typography.scss
kalinchernev: Thanks for noticing, there were indeed a few differences

src/implementations/vanilla/utilities/clearfix/README.md
kalinchernev: For this one and the others, I've taken these pages https://github.com/ec-europa/europa-component-library/tree/v2-dev/src/website/src/pages/ec/utilities

Do you mean there is other existing information I've missed?

emeryro: no, the documentation pages do not exist yet, but at some point that would be a good idea to create them.
That's why I suggested to have another ticket to add them (we are the one writing documentation for utilities, as it is not something managed by COMM)

kalinchernev: FRONT-1786 created as a follow-up
The rest of the feedback points have been covered

emeryro (approved):

PR #1757, opened 2020-11-13
feat: migrate date picker - FRONT-1773

kalinchernev:
Moment.js has been loaded from storybook's preview-head in order to not be included in the end ECL bundle.

The default script uglification in ecl-builder has been updated to preserve moment.js. If not configured, the local variable of the imported UMD pikaday loses scope with the globally loaded moment.js.

Unfortunately, moment.js is not optional dependency of pikaday and we are required to still carry it around https://github.com/Pikaday/Pikaday/issues/815

The `yarn start:dev` command has been updated to actually work.


kalinchernev:
WIP: turns out https://github.com/ec-europa/europa-component-library/pull/1757/files#diff-5ea60b663103cd1f563b3595b74f1c1414f52a8929715eac08cb9c54dcf4e3d9 solves it for dev mode and build folders, but the dist doesn't seem to play well.

If any ideas and suggestions: feel free to update

src/implementations/twig/components/datepicker/README.md
```diff
@@ -0,0 +1,45 @@
...
+- **"id"** (string) (default: '')
+- **"type"**: (string) (default: 'text')
+- **"name"** (string) (default: '')
+- **"icons_path"**: (string) (default: '')
+- **"placeholder"**: (string) (default: 'dd/mm/yyyy')
+- **"extra_classes"** (optional) (string) (default: '') Extra classes (space separated)
+- **"extra_attributes"** (optional) (array) (default: []) Extra attributes
+  - "name" (string) Attribute name, eg. 'data-test'
+  - "value" (string) Attribute value, eg: 'data-test-1'
+- **"_compliance_"** (boolean) (default: false) Activates debug
```
planctus:
```suggestion
- **"\_compliance\_"** (boolean) (default: false) Activates debug
```

src/presets/ec-core/package.json
```diff
@@ -14,6 +14,9 @@
     "build": "rimraf build && run-p build:* -ln",
     "clean": "rimraf build dist",
     "dist": "npm run clean && cross-env NODE_ENV=production run-p build:* -ln",
+    "watch:scripts": "chokidar \"../../implementations/vanilla/**/*.js\" \"../**/*.js\" -i \"/node_modules|build|dist|config\\.js/\" -c \"npm run build:scripts\"",
+    "watch:styles": "chokidar \"../../implementations/vanilla/**/*.scss\" \"../../../../themes/**/*.scss\" \"../**/*.scss\" -i \"/node_modules|build|dist/\" -c \"npm run build:styles\"",
```
planctus, line 18: nice to have these watchers, i wonder whether we could also have an "hot-reloading" in storybook, would be very handy

src/tools/builder/scripts/scripts.js
```diff
@@ -11,7 +11,12 @@ module.exports = (input, dest, options) => {
     options.uglify === true ||
     (options.uglify !== false && process.env.NODE_ENV === 'production');
 
-  const uglifyOptions = {};
+  // ECL uses Pikaday for the datepicker component. Pikaday dynamically requires moment.js which messes up JS bundling.
+  // ECL does not want to include moment.js in its release in order to reduce the final bundle size.
+  // Instruct minifier to preserve the UMD locally scoped 'moment' (default Pikaday module) variable in Pikaday in order to correctly reference the global 'moment' included separately from the ECL library bundle.
+  // When Pikaday really removes moment from its dependencies and does not load it dynamically, bundlers such as rollup will be able to handle this more gracefully.
+  // @see https://github.com/Pikaday/Pikaday/issues/815
```
planctus, line 18: So this means that for using the component the consumers need to add moment.js on their side, right? Shouldn't we add this information in the api.mdx file, then?

planctus (approved):

src/presets/ec-core/package.json
kalinchernev: Good remark. Actually it's there, but things are not as obvious and it's the subject I referred to today as "items to improve" in the dev task.

Two watch tasks are spawned in parallel [with lerna from project root](https://github.com/ec-europa/europa-component-library/blob/6e64dfcc13e79cd195779b8b34824cbe339912a1/package.json#L26-L28).

When any file of type .js or .css is updated in the vanilla implementation (presets watch task), ecl-builder rebuilds the corresponding styles and creates a new "build" version of the preset at `src/presets/(ec|eu)-core/build`. These build folders are indeed "sourced" through the watch task of the storybook instance https://github.com/ec-europa/europa-component-library/blob/6e64dfcc13e79cd195779b8b34824cbe339912a1/src/implementations/twig/storybook/package.json#L11 which is in its own turn also in watch mode (default to storybook actually)

The issues with watch tasks in presets in both ec/eu at the same time are several in my view:

1) log messages are lost because the way lerna spawns processes. i.e. we see `lerna [message]` types of logs, but it's not the same as running `yarn watch` in `src/presets/ec-core` for example (here in this comment) chokidar gives more details regarding its process. Error messages, warnings and useful feedback about rebuilding success on preset level is lost.

2) storybook watch task is not aware of the events emitted by chokidar which makes it so that ecl-builder successfully rebuilds the scss/javascript of the given preset, but it's not reflected in storybook until storybook application is reloaded fully and manually. If a file in storybook is touched, however, its active watcher in hot-reload will push changes to the storybook application successfully without any issues

For the first issue, a possible solution would be to update preset watch tasks to be like the storybook one: changing directory to the given preset folder and running the watch task without lerna.

For the second, I don't know yet. [chokidar](https://github.com/paulmillr/chokidar) has 'ready' event and we can potentially use it through a node api script. For the storybook watch task, I don't know: definitely welcome ideas how we can go lower than their utility script. Reading a bit, it goes to `node_modules/@storybook/html/dist/server/index.js` which is https://github.com/storybookjs/storybook/blob/next/lib/core/server.js and it seems on first look they use webpack. Not sure how we can listen to events emitted by their CLI (and even further emitting messages to the server )

src/tools/builder/scripts/scripts.js
kalinchernev: Documentation added

planctus (approved):

PR #1763, opened 2020-11-17
feat(menu): migrate component menu - FRONT-1780

emeryro:
# PR description

Migrate menu component to new structure


kalinchernev:
The migration is fine, no feedback from my side.
Only needs to be updated to the latest next-v3

PR #1781, opened 2020-11-24
feat: add site header harmonised - FRONT-1800

kalinchernev:


src/implementations/vanilla/components/site-header-harmonised/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "main": "site-header-harmonised.js",
+  "module": "site-header-harmonised.js",
+  "sass": "site-header-harmonised.scss",
+  "style": "site-header-harmonised.scss",
+  "dependencies": {
+    "@ecl/theme-dev": "^3.0.0",
+    "@ecl/vanilla-layout-grid": "^3.0.0",
+    "@ecl/dom-utils": "^3.0.0",
+    "@ecl/vanilla-component-icon": "^3.0.0",
+    "@ecl/vanilla-component-search-form": "^3.0.0",
```
emeryro: Do we need these dependencies? (vanilla-component-search-form and vanilla-component-icon)

kalinchernev: I'd suggest we keep them as dependencies here, yes, although real-world impact is questionable. The component is composed of these components and can't exist without them.

In v2, there is the [`check-imports`](https://github.com/ec-europa/europa-component-library/blob/v2-dev/src/systems/ec/implementations/vanilla/packages/ec-component-site-header-harmonised/ec-component-site-header-harmonised.scss#L10). I don't think we need to add these back, but if someone decides to install the npm package of the component, the CSS of the composed components should come in.

I can remove of course, I doubt anyone will actually fetch the npm package separately from ECL given the fact how many times we've been communicating publishing channels of the library.

What do you think? Remove or better leave in?

If we are to leave them in, we'll actually have to add the menu and a few more part of the composition I think.

emeryro: well, if we decide to keep this kind of dependencies, we should then add them in all components (most of our component uses other small components like icons or links).
that would be cleaner, but will require some work to update/maintain it

emeryro (approved):

PR #1783, opened 2020-11-25
feat(token): add eu tokens - FRONT-1814

emeryro:
# PR description

Add EU specific tokens. Currently it only concerns colours, as other token (spacing, typography, ...) are the same as EC

Changes:
- color palette defined in themes
- remove background specific colours (only used in background utility). I no longer see a reason to not use the full palette here... (https://github.com/ec-europa/europa-component-library/pull/1783/files#diff-ccaf8bd71390bd1bfab962017ab3700bd16be4d0805be66cfd42dd1f16f1b971R3)


src/implementations/vanilla/components/breadcrumb-harmonised/breadcrumb-harmonised.scss
```diff
@@ -9,10 +9,10 @@
 @use '@ecl/theme-dev/maps/typography';
 
 $_background-color: transparent;
-$_background-color-hover: map.get(color.$color, 'blue-5');
```
kalinchernev: I find the approach of having both ec/eu pairs in 1 theme sub-optimal. It's imposing an imperative mode which opens the door for possible blow in complexity and duplication

In the current setup in `next-v3` the specificity is organized in declarative way in 1 place in the (ec|eu)-core themes rather than the components.

Following the example we have with the globals, `src/themes/eu-core/_index.scss`:

```scss
@use '@ecl/theme-dev/globals' with (
  $border-radius: 'yes'
);
@use '@ecl/theme-dev/maps/color' with (
  $color: (
    'blue-5': #f3f6fc,
    'blue-10': #e7edfa,
    'blue-20': #cfdaf5,
    'blue-40': #9fb5ea,
    'blue-60': #6e91e0,
    'blue-80': #3e6cd5,
    'blue-100': #0e47cb,
    'blue-110': #0d40b7,
    'blue-120': #0b39a2,
    'blue-130': #0a328e,
    'blue-140': #082b7a,
    'yellow-10': #fffae6,
    'yellow-20': #fff5cc,
    'yellow-40': #ffeb99,
    'yellow-60': #ffe066,
    'yellow-80': #ffd633,
    'yellow-100': #fc0,
    'yellow-110': #e6b800,
    'yellow-120': #cca300,
    'yellow-130': #b38f00,
    'yellow-140': #b38f00,
    'grey-5': #f4f5f5,
    'grey-10': #e9eaeb,
    'grey-20': #d4d5d7,
    'grey-40': #a8aaaf,
    'grey-60': #7d8088,
    'grey-80': #515560,
    'grey-100': #262b38,
    'grey-110': #222732,
    'grey-120': #1e222d,
    'grey-130': #1b1e27,
    'grey-140': #171a22,
    'purple-100': #510dcd,
  )
);
```

I.e. the specificity comes by an override on a theme level.

This approach would mean that `map.get(color.$color, 'blue-5')` here will be replaced with `#f3f6fc` (instead of ` rgb(242, 245, 249)`) at `src/presets/eu-core/src/eu-core.scss`

```scss
@use '@ecl/theme-eu-core';
@use '@ecl/preset-dev/src/dev';
```

The theme layer being first will replace tokens with their ec/eu specific values and then compile the component-specific code based on the tokens' specific values.

emeryro: Interesting.. So basically, the default map file would contain the common values, and the specific ones (EC/EU) would be defined in the theme file, right?
If tomorrow we have other themes, we may have to duplicate these values, but it shouldn't be a big overload.

emeryro: But the tokens are not the same between EC and EU (for instance there is no yellow-20 in EC, and no yellow-25 in EU.
So in the end we would have to have 2 separate scss files in components in any case I think

kalinchernev: Missing tokens between themes can go into the dev becoming defaults. It's up to design rules for the tokens to not be used in the wrong places in the components

emeryro: The thing is, we don't know what would be the differences between EC and EU. That may be more than just a colour swap. For instance, there will probably be cases where there is a background colour on EU, and not on EC.
So I don't know how we could handle this kind of thing without having 2 scss files in the component.
But this does not invalidate your proposal.

kalinchernev: It's true we lack long-term vision. For now we know there is a difference in colors and we can handle them.

There's a card we can play before splitting files: having system-aware context variable passed to scss. Here's the ["can do" commit](https://github.com/ec-europa/europa-component-library/pull/1692/commits/c91aa7eab1213f0054332fd8462d9c959cc394b9#diff-37476fe9fa6e6d92477a0c7c7f0672ca07dab5685faee067d4c4c47889f29c0d).

If the difference is in a property or two in a given file, we can still use this approach and set specifics conditionally without having multiple files.

emeryro:
I applied the feedback from @kalinchernev, things are already cleaner this way.
We will see when dealing with real components if an environment variable could be enough to differentiate EC/EU, or if we have to split the scss into 2 files (see comments ahead).
But for the tokens, I guess we are good

src/themes/dev/maps/color.scss
```diff
@@ -1,85 +1,2 @@
-$color: (
-  'blue': rgb(0, 68, 148),
```
kalinchernev, line 2: I think these can stay here as defaults as the eu can be considered an override I think

src/themes/ec-core/_index.scss
```diff
@@ -1,7 +1,55 @@
 @use '@ecl/theme-dev/globals' with (
   $border-radius: 'no'
 );
-@use '@ecl/theme-dev/maps/color';
+@use '@ecl/theme-dev/maps/color' with (
+  $color: (
+    'blue': #004494,
```
kalinchernev: Not so important for the task at hand, but I will comment it as I notice it: the color formats are different, one in RGB and the other in hex.

src/themes/ec-core/_index.scss
```diff
@@ -1,7 +1,55 @@
...
+    'yellow': #ffd617,
```
kalinchernev: In the ec-core, there won't be overrides `with` if we agree to accept the `ec` as defaults.

src/website/src/components/SplashPage/SplashPage.scss
```diff
@@ -1,4 +1,5 @@
 @use "sass:map";
+@use '@ecl/theme-ec-core';
```
kalinchernev: I think these won't be necessary if the dev contains the ec as defaults

src/themes/dev/maps/color.scss
kalinchernev: additional comment: we'll have to merge the eu tokens in here

src/themes/ec-core/_index.scss
emeryro: yes, I changed it back to hexa, as it is on ECL 2 (not sure why they got converted to rgb at some point...)

src/themes/dev/maps/color.scss
emeryro: I don't think we need to megre the EU tokens. The `with` option handles this well. If the item do not exists, it is created.

kalinchernev: it'll be very handy if there is no requirement for these, i hadn't tested it manually but really great if we can keep them separate and treat the eu as override

src/website/package.json
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "@ecl/preset-full": "^3.0.0",
     "@ecl/resources-ec-icons": "^3.0.0",
-    "@ecl/theme-dev": "^3.0.0",
+    "@ecl/theme-ec-core": "^3.0.0",
```
kalinchernev: Pls revert this change

emeryro: yes, good catch

src/themes/dev/maps/color.scss
planctus: It looks like it's not working as expected, unless i'm missing something all the missing keys in the map we use in the components scss don't generate any output in the css, this is the blockquote in eu-core, for instance, that uses `yellow`:
image

kalinchernev: Yes, excellent catch. It's something we found and discuss indeed

emeryro: yes, just to be clear on one point: this PR just adds the EU tokens, but there are no relation between EC and EU colours (they don't even use the same token name).
So of course, they will be modification required at each component level (to use one of the colour available in EU), but that will be done when we have the corresponding specs. We can mitigate this by adding a few missing colours, but it won't solve everything

kalinchernev: As we discussed in the chat yesterday, it turns out that `@use` is not achieving what we aimed for in terms of final list of tokens. 

After spending a few hours trying to solve it, I can propose 2 options:

1) Use `map.merge` to actually merge the maps. The approach will have to change a bit and not be `with()` but go around modules and namespaces:

```scss
@use 'sass:map';

@use '@ecl/theme-dev/globals' with (
  $border-radius: 'yes'
);

@use '@ecl/theme-dev/maps/color' as *;

$color: map.merge(
  $color,
  (
    'blue-5': #f3f6fc,
    'blue-10': #e7edfa,
    'blue-20': #cfdaf5,
    'blue-40': #9fb5ea,
    'blue-60': #6e91e0,
    'blue-80': #3e6cd5,
    'blue-100': #0e47cb,
    'blue-110': #0d40b7,
    'blue-120': #0b39a2,
    'blue-130': #0a328e,
    'blue-140': #082b7a,
    'yellow-10': #fffae6,
    'yellow-20': #fff5cc,
    'yellow-40': #ffeb99,
    'yellow-60': #ffe066,
    'yellow-80': #ffd633,
    'yellow-100': #fc0,
    'yellow-110': #e6b800,
    'yellow-120': #cca300,
    'yellow-130': #b38f00,
    'yellow-140': #b38f00,
    'grey-5': #f4f5f5,
    'grey-10': #e9eaeb,
    'grey-20': #d4d5d7,
    'grey-40': #a8aaaf,
    'grey-60': #7d8088,
    'grey-80': #515560,
    'grey-100': #262b38,
    'grey-110': #222732,
    'grey-120': #1e222d,
    'grey-130': #1b1e27,
    'grey-140': #171a22,
    'purple-100': #510dcd,
  )
);
```

2) Place missing values in 

```scss
@use 'sass:map';

@use '@ecl/theme-dev/globals' with (
  $border-radius: 'yes'
);
@use '@ecl/theme-dev/maps/color' with (
  $color: (
    // defaults copy
    'blue': #004494,
    'yellow': #ffd617,
    'grey': #404040,
    'black': #000,
    'white': #fff,
    'blue-75': #4073af,
    'blue-50': #7fa1c9,
    'blue-25': #bfd0e4,
    'yellow-75': #ffde39,
    'yellow-25': #fff4bb,
    'grey-25': #cfcfcf,
    'grey-15': #e3e3e3,
    'blue-n': #006fb4,
    'orange': #f29527,
    'green': #467a39,
    'red-130': #981722,
    'red': #da2131,
    'primary': #004494,
    'secondary': #ffd617,
    'tertiary': #e3e3e3,
    'text': #404040,
    'white-100': #fff,
    'black-100': #000,
    'error': #da2131,
    'info': #006fb4,
    'success': #467a39,
    'warning': #f29527,
    'orange-100': #f29527,
    'green-100': #467a39,
    'red-100': #da2131,
    // new and overrides
    'blue-5': #f3f6fc,
    'blue-10': #e7edfa,
    'blue-20': #cfdaf5,
    'blue-40': #9fb5ea,
    'blue-60': #6e91e0,
    'blue-80': #3e6cd5,
    'blue-100': #0e47cb,
    'blue-110': #0d40b7,
    'blue-120': #0b39a2,
    'blue-130': #0a328e,
    'blue-140': #082b7a,
    'yellow-10': #fffae6,
    'yellow-20': #fff5cc,
    'yellow-40': #ffeb99,
    'yellow-60': #ffe066,
    'yellow-80': #ffd633,
    'yellow-100': #fc0,
    'yellow-110': #e6b800,
    'yellow-120': #cca300,
    'yellow-130': #b38f00,
    'yellow-140': #b38f00,
    'grey-5': #f4f5f5,
    'grey-10': #e9eaeb,
    'grey-20': #d4d5d7,
    'grey-40': #a8aaaf,
    'grey-60': #7d8088,
    'grey-80': #515560,
    'grey-100': #262b38,
    'grey-110': #222732,
    'grey-120': #1e222d,
    'grey-130': #1b1e27,
    'grey-140': #171a22,
    'purple-100': #510dcd,
  )
);
```

None of these are ideal in my opinion, though I tend to have preference for the second because it is ok to have all desired values in 1 place for easier access and reference and the first one is not a real solution but a work-around.

Let me know if you have other suggestions

PR #1789, opened 2020-11-27
chore: illustrate forward at rule

kalinchernev:
I didn't know that `@forward` at-rule exists until yesterday while working on https://github.com/ec-europa/europa-component-library/pull/1783

This is the smallest diff I could think of to try the functionality and see whether it's going to be liked by the team.

The benefit I see probably comes a bit too late in the migration process, but what we could have is `theme.{token}` for theme tokens instead of including and using them directly.

Speaking about accessing tokens indirectly, it's probably worth reflecting a bit on possible combination of `@forward` plus https://github.com/ec-europa/europa-component-library/pull/1740

The direction of though is like following:

```scss
// inside a component

@use '@ecl/theme-dev/theme' // new file containing facilities to access tokens indirectly

.ecl-button--primary {
  background-color: theme.get('color', 'blue-100');
  color: theme.get('color', 'white-100');
}
```

In `.get()` (or whatever other function/mixin) there is a (ec|eu) context which is hidden from the component and returns a value for the requested token.

This only a shared experiment for a discussion

### DIFF
--- a/src/implementations/vanilla/components/button/button.scss
+++ b/src/implementations/vanilla/components/button/button.scss
@@ -5,15 +5,12 @@
 
 @use "sass:map";
 @use '@ecl/theme-dev/globals';
-@use '@ecl/theme-dev/maps/color';
-@use '@ecl/theme-dev/maps/shape';
-@use '@ecl/theme-dev/maps/spacing';
-@use '@ecl/theme-dev/maps/typography';
+@use '@ecl/theme-dev/theme';
 
 $_border-width: 2px;
 $_outline-width: 3px;
-$_padding-horizontal: map.get(spacing.$spacing, 'm');
-$_padding-vertical: map.get(spacing.$spacing, 's');
+$_padding-horizontal: map.get(theme.$spacing, 'm');
+$_padding-vertical: map.get(theme.$spacing, 's');
 
 .ecl-button {
   appearance: none;
@@ -21,15 +18,15 @@ $_padding-vertical: map.get(spacing.$spacing, 's');
   border-radius: map.get(
     (
       no: 0,
-      yes: map.get(shape.$border-radius, 'm'),
+      yes: map.get(theme.$border-radius, 'm'),
     ),
     globals.$border-radius
   );
   border-width: 0;
   box-sizing: border-box;
   display: inline-block;
-  font: map.get(typography.$font, 'm');
-  font-weight: map.get(typography.$font-weight, 'bold');
+  font: map.get(theme.$font, 'm');
+  font-weight: map.get(theme.$font-weight, 'bold');
   margin: 0;
   padding: $_padding-vertical $_padding-horizontal;
   text-decoration: none;
@@ -41,7 +38,7 @@ $_padding-vertical: map.get(spacing.$spacing, 's');
   }
 
   &:focus {
-    outline: $_outline-width solid map.get(color.$color, 'secondary');
+    outline: $_outline-width solid map.get(theme.$color, 'secondary');
     outline-offset: -($_outline-width);
   }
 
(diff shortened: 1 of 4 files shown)


planctus:
I didn't know @forward either, @kalinchernev and i like how you used it to simplify the access to the modules by the components.
What i feel, but it's not anything regarding this POC, is that we shouldn't delegate to the themes the definitions of the "tokens", we don't have "systems" anymore as a main layer in the library but we should still have it to group together the tokens for each system, otherwise we would have to replicate for each theme all the overrides belonging to the "system" and not really the specific theme which should only set its own "global parameters", i believe.

planctus:
I've tried to imagine a different structure, then, where the definition of the tokens is in a different folder than the one holding the themes, i've used your work and extended the use of forward to a package per system in a "tokens" folder, i've created this pr to illustrate this idea, not sure it makes sense as it is right now, but you can give it a look:
https://github.com/ec-europa/europa-component-library/pull/1792

kalinchernev:
@planctus I agree the "theme" layer is not as mature as it needs to be. To be honest, I've been trying to delegate the ec/eu decision-making / token-specifying "logic" there, as when migrating to dart-sass I learnt about the `with()` feature which felt like a good way to organize, well "themig" :)

I don't mind having tokens layer at all. Though I don't feel the scss maps we have at the moment are really tokens. As far as my knowledge on the subject extends, tokens are a concept aiming to provide a reusable layer between different systems. Like what we had in the past yaml files generating json/scss, etc. So the maps at the moment are sort of an escape structure we poured previously known tokens from [theo](https://github.com/salesforce-ux/theo) to pure scss.

So in short, I don't see anything wrong in going to pure tokens direction. However, I foresee that the more we clear them up, we'll understand that we'll need to re-introduce theo :) The only [scss parser](https://github.com/salesforce-ux/scss-parser) on the market is also created by salesforce team. I mean, I guess they've walked the same path we're heading in terms of tokens management

emeryro:
I do love this `@forward` too.
What I don't full understand is why we are heading towards tokens managed with theo (or something else).
The current situation is:
- we have different set of "tokens" (in the sense of "a key and a value") shared by all systems/themes (spacing, typography, ...)
- some tokens are specific to a system (only colours for now, we don't know if there will be more later). Some of these specific tokens have the same key name in both systems (but different values), some keys only exists in one system (with no plan to have them on both)
- these tokens are only used in scss files

So, if we define tokens, that would be per system (we can't have a global "yellow-100" token for instance).
And as they are only used in scss, I don't see the need to bring back an aditional transcription step from a "token format" to a scss variable.
Also as things go on, I think that we are heading towards the "one system = one theme" structure that we had on ECL 2. But we should not forget that the inital goal was to have multiple themes per system (if that still makes any sense).

In short, having things like `background-color: theme.get('color', 'blue-100');` or `background-color:: map.get(theme.$color, 'blue-100');` would be really nice, but I don't really see how we could have the `@use '@ecl/theme-dev/theme'` instead of `@use '@ecl/theme-ec-core/theme'`

kalinchernev:
> What I don't full understand is why we are heading towards tokens managed with theo (or something else).

It's in the loop of the discussion in the context of having the suggestion in #1792. My feedback on this is that we don't need to formalize tokens as a new layer, but if we do, we need to make tokens to be real tokens.

> some tokens are specific to a system (only colours for now, we don't know if there will be more later). Some of these specific tokens have the same key name in both systems (but different values), some keys only exists in one system (with no plan to have them on both)

We are ready for multiple themes and we're not bound to theme to system. `@ecl/theme-dev` is a child of ec-base and ecl design tokens and it's holding ec-specifics simply because the eu has been a moving target and there is no real base. We can scale adequately having combinations of modules' overrides (variables holding maps) without complications on components level given the constraints of the maps are kept adequately. i.e. we can support exotics of having a theme which overrides $breakpoint, $container, etc. of the layout module.

> but I don't really see how we could have the @use '@ecl/theme-dev/theme' instead of @use '@ecl/theme-ec-core/theme'

Im not sure I understand this one. What is the use case?

In this pull request I added the [theme file](https://github.com/ec-europa/europa-component-library/pull/1789/files#diff-aa647292d6da1ad10a5dd3d1460c93a8442de57b439d19f2962f151007a56cfb) only out of laziness but the name could be anything else. The way use at-rule works is that it picks the name of the last file of the pattern. From `theme-dev/theme` it will be `theme` which will name the module as such. It's possible and sometimes necessary to rename like `@use @ecl/theme-dev` (which leads to index partial the way dart-sass works) rename to `@use @ecl/theme-dev as theme`. Thus, it should be possible that we combine themes if necessary 

```scss
@use '@ecl/theme-ec-core/theme' as ecCoreTheme;
@use '@ecl/theme-eu-core/theme' as euCoreTheme;
```

kalinchernev:
Created https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1896 as a follow-up based on the POC and the discussions around it

PR #1790, opened 2020-11-30
feat(eu-footer): add eu footer - FRONT-1813

emeryro:
# PR description

Specs on [Abstract:](https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/4102d9a8-50e3-47b8-b1f1-3458556f9cde/collections/c3226eb4-177a-4fcd-9b6b-1b743b3a0169) 

Breaking changes: 
- footer markup has been improved (it no longer jumps from section 1 to section 3)
- data have been updated to display only a link on mobile instead of the full list of EU institutions 

Other modifications:
- list items in each section can now contain optional plain text before and/or after the links


src/implementations/twig/components/footer-core/footer-core.html.twig
```diff
@@ -97,10 +100,6 @@
         extra_classes: 'ecl-footer-core__logo-link'
       } only %}
     {% endif %}
-    {# content before #}
```
kalinchernev, line 98: Removing these lines of code does not seem to be connected to links' before and after because they related to a section and not its links iterable property.

I don't know the reason of having this content before and after on a section level, though it could become a breaking change caused by seemingly unrelated change.

src/presets/eu-core/src/eu-core-print.scss
```diff
@@ -1,2 +1,5 @@
 @use '@ecl/theme-eu-core';
 @use '@ecl/preset-dev/src/dev-print';
+
+// EU specific (temporary solution)
```
kalinchernev: Checking on `footer-core-ec-print.scss` and `footer-core-eu-print.scss` there is no difference. Why having 2 files when there are no exceptions to handle?

src/presets/dev/src/dev-print.scss
```diff
@@ -29,7 +29,7 @@
 @use '@ecl/vanilla-component-expandable/expandable-print';
 @use '@ecl/vanilla-component-fact-figures/fact-figures-print';
 @use '@ecl/vanilla-component-file/file-print';
-@use '@ecl/vanilla-component-footer-core/footer-core-print';
+// @use '@ecl/vanilla-component-footer-core/footer-core-print';
```
kalinchernev: Having one and the same contents for both cases, I think it's save to keep this line.

src/presets/dev/src/dev.scss
```diff
@@ -32,7 +32,7 @@
 @use '@ecl/vanilla-component-expandable/expandable';
 @use '@ecl/vanilla-component-fact-figures/fact-figures';
 @use '@ecl/vanilla-component-file/file';
-@use '@ecl/vanilla-component-footer-core/footer-core';
+// @use '@ecl/vanilla-component-footer-core/footer-core';
```
kalinchernev: Could be removed

src/presets/ec-core/src/ec-core.scss
```diff
@@ -1,2 +1,5 @@
 @use '@ecl/theme-ec-core';
 @use '@ecl/preset-dev/src/dev';
+
+// EC specific (temporary solution)
```
kalinchernev: I find this solution to be good enough for a standard one. In this case, `footer-core-ec.scss` and `footer-core-eu.scss` are too different to try to accommodate them on the dev layer.

kalinchernev:
```suggestion
// Exceptions
```

src/presets/eu-core/src/eu-core.scss
```diff
@@ -1,2 +1,5 @@
 @use '@ecl/theme-eu-core';
 @use '@ecl/preset-dev/src/dev';
+
+// EU specific (temporary solution)
```
kalinchernev:
```suggestion
// Exceptions
```

kalinchernev (changes requested):
Questions and a few suggestions. Apart from these:
- https://ecl-preview-1790--europa-component-library.netlify.app/ec/core-template/footer/code/ is not showing the demo (storybook paths have been changed)
- latest target branch contains storybook split, will be good to have at this stage

src/implementations/twig/components/footer-core/footer-core.html.twig
emeryro: this part is not used on EU footer (it was used on EC one, and the code was the same).
I could bring it back, but as the content of the footer can't be modified on a site level (it's corporate), I don't think that it would cause trouble

src/presets/ec-core/src/ec-core.scss
emeryro: Yes, I think that we will have indeed 2 scss files here.
My concern is that we add these files after the others, and not in the order specified in dev preset.
For the footer that's not an issue, but we will have to be careful about it if we do the same for other components

emeryro:
Some updates have been made to apply feedback from COMM:
- remove extra spacing below the logo
- change breakpoints
- display only a link on mobile instead of the full list of EU institutions

src/presets/eu-core/src/eu-core-print.scss
kalinchernev: Double-checked, still relevant

src/presets/dev/src/dev-print.scss
kalinchernev: Double-checked, still relevant

src/presets/dev/src/dev.scss
kalinchernev: Double-checked, still relevant

src/presets/ec-core/src/ec-core.scss
kalinchernev: Yes, of course we'll be careful with order. I'm very skilled at messing this :)
Yet, let's remove the "(temporary solution)" part as it won't be temporary.
So far, we've had issues in order with icons and links which is not a great amount of code to keep in mind. If we did end up having issues with order, we can split problematic modules out of the dev preset and move them here. Still, I really hope this will be an exception.

kalinchernev (approved):

PR #1791, opened 2020-11-30
feat(gallery): Add embedded iframe video support - FRONT-1824

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/implementations/vanilla/components/gallery/gallery.js
+++ b/src/implementations/vanilla/components/gallery/gallery.js
@@ -468,13 +468,6 @@ export class Gallery {
         maxHeight: `${maxHeight}px`,
       });
     }
-
-    // DEPRECATED
-    else if (this.overlayImage) {
-      Object.assign(this.overlayImage.style, {
-        maxHeight: `${maxHeight}px`,
-      });
-    }
   }
 
   /**
--- a/src/implementations/vanilla/components/gallery/gallery.js
+++ b/src/implementations/vanilla/components/gallery/gallery.js
@@ -357,7 +357,9 @@ export class Gallery {
    */
   updateOverlay(selectedItem) {
     this.selectedItem = selectedItem;
-    const embeddedVideo = selectedItem.dataset.eclGalleryItemEmbedSrc;
+    const embeddedVideo = selectedItem.getAttribute(
+      'data-ecl-gallery-item-embed-src'
+    );
     const video = queryOne('video', selectedItem);
     let mediaElement = null;
 
@@ -367,8 +369,7 @@ export class Gallery {
       mediaElement = document.createElement('div');
       mediaElement.classList.add('ecl-gallery__slider-embed');
 
-      let mediaIframe = null;
-      mediaIframe = document.createElement('iframe');
+      const mediaIframe = document.createElement('iframe');
       mediaIframe.setAttribute('src', embeddedVideo);
       mediaIframe.setAttribute('frameBorder', '0');
 
@@ -422,14 +423,6 @@ export class Gallery {
       if (this.overlayMedia) {
         this.overlayMedia.innerHTML = '';
         this.overlayMedia.appendChild(mediaElement);
-      }
-
-      // DEPRECATED
-      else if (this.overlayImage) {
-        this.overlayImage.setAttribute(
-          'src',
-          selectedItem.getAttribute('href')
-        );
       }
     }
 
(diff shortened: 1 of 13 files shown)


src/implementations/vanilla/components/gallery/gallery.scss
```diff
@@ -15,6 +15,29 @@ $_image-height: 185px;
 $_description-background: map.get(color.$color, 'grey-100');
 $_description-color: map.get(color.$color, 'white-100');
 
+@mixin ecl-aspect-ratio($width, $height) {
```
papegaill: Here I needed to define a `mixin` which is already declared in the media-container component:

https://github.com/ec-europa/europa-component-library/blob/4092b3aadcf87e148b679aee92bede0a0ad794d0/src/implementations/vanilla/components/media-container/media-container.scss#L11

Where can I put this `mixin` to make it global, create a new package?

FI, there is also an existing `mixin` in the utilities, but which doesn't take the iframe into account:
https://github.com/ec-europa/europa-component-library/blob/4af96b17a5d9ab4a2a5e6e2dbbdd04d27889de04/src/implementations/vanilla/utilities/media/media.scss#L6

src/implementations/vanilla/components/gallery/gallery.js
```diff
@@ -357,11 +357,27 @@ export class Gallery {
    */
   updateOverlay(selectedItem) {
     this.selectedItem = selectedItem;
+    const embeddedVideo = selectedItem.dataset.eclGalleryItemEmbedSrc;
```
kalinchernev: You can use `getAttribute()` in a similar way as the `setAttribute()`. The `dataset`:

```js
const embeddedVideo = selectedItem.getAttribute('data-ecl-gallery-item-embed-src')
```

It'll be also more consistent to the following used in the same file

```javascript
const shareHref = this.selectedItem.getAttribute('data-ecl-gallery-item-share');
```

src/implementations/vanilla/components/gallery/gallery.js
```diff
@@ -357,11 +357,27 @@ export class Gallery {
...
     let mediaElement = null;
 
     // Update media
-    if (video != null) {
+    if (embeddedVideo != null) {
+      // Media is a embedded video
+      mediaElement = document.createElement('div');
+      mediaElement.classList.add('ecl-gallery__slider-embed');
+
+      let mediaIframe = null;
```
kalinchernev: `mediaIframe` is not changing its value, so it can be `const`.

src/implementations/vanilla/components/gallery/gallery.js
```diff
@@ -407,29 +423,39 @@ export class Gallery {
...
-    // Limit image height (fix for FF and IE)
-    const maxHeight =
-      this.overlay.clientHeight -
-      this.overlayHeader.clientHeight -
-      this.overlayFooter.clientHeight;
-    if (this.overlayMedia) {
-      Object.assign(mediaElement.style, {
-        maxHeight: `${maxHeight}px`,
-      });
+      // DEPRECATED
```
kalinchernev: Is the deprecated part of the code carried for a specific reason? I had the impression we use v3 for removing deprecated code.

src/implementations/vanilla/components/gallery/gallery.scss
kalinchernev: Good finding that there is an existing mixin in utilities. This is the right global place to put items to be shared between components and website.

Looking briefly at the source code, I think the best approach would be to add the iframe specifics into the utility and then use this utility in the media container and this component.

kalinchernev: On a separate note, there is no need to have `ecl-` namespaces at variables and mixins in v3 branch anymore, as we use the `@use` feature to import files which are treated as modules and there are no naming clashes.

I.e. v2 uses libsass and there are no modules, whereas v3 uses dart-sass and uses modules.

kalinchernev (commented):
Added a few comments which are not blocking the merge although I think they are opportunities to improve and are good to have

kalinchernev:
Added "Approved" tag as apart from the comments I consider the PR mergeable

src/implementations/vanilla/components/gallery/gallery.scss
papegaill: Super @kalinchernev many thanks for the clarifications !

src/implementations/vanilla/components/gallery/gallery.js
papegaill: Thank you I didn't pay attention to this deprecated code. Tested and removed.

papegaill:
> Added "Approved" tag as apart from the comments I consider the PR mergeable

Thanks for the review and the suggestions I applied 👍

kalinchernev (approved):

PR #1792, opened 2020-12-01
chore: A different file system structure, separating the definition of the tokens and the themes

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


kalinchernev:
I added a comment in [the other pull request](https://github.com/ec-europa/europa-component-library/pull/1789#issuecomment-736384195) though wanted to add a reference to what we have in v2 to illustrate my point about tokens: 
https://github.com/ec-europa/europa-component-library/tree/v2-dev/src/systems/ec/design-tokens

To sum up:

- https://github.com/ec-europa/europa-component-library/pull/1789 tries to facilitate using scss maps of theme-dev by using the forward at-rule which makes it so that we can access map values without directly using the maps' modules
- maps of themes-dev are what we previously called tokens, but they are not really tokens
- i suggest that we really but back theo or other tool to have real tokens if we have `@ecl/tokens`

kalinchernev:
Closing this one for now after the discussions we had and the agreement to leave out tokens abstraction out for the moment.
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1896 has been created for follow-up on the track

PR #1794, opened 2020-12-02
feat(storybook): Adding EU instance - FRONT-1684

planctus:
# PR description

the variable needs to start with STORYBOOK, so i haven't renamed it as we wanted to.
https://storybook.js.org/docs/react/configure/environment-variables

## QA Checklist

- [ ] I have provided tests


package.json
```diff
@@ -21,11 +21,13 @@
     "pretty-check": "prettier --check \"src/**/*.{md,mdx,html,json,yml,js,jsx,scss}\"",
     "publish": "scripts/publish.sh",
     "serve": "cd src/website && npm run serve",
-    "start": "cd src/implementations/twig/storybook && npm run start",
+    "start:ec": "cd src/implementations/twig/storybook/ec && npm run start",
+    "start:eu": "cd src/implementations/twig/storybook/eu && npm run start",
     "start:dev": " lerna --scope \"@ecl/preset-*-core\" run clean && lerna --scope \"@ecl/preset-*-core\" run build && npm run start:watch",
     "start:watch": "run-p start:watch:*",
     "start:watch:presets": "lerna --scope \"@ecl/preset-*-core\" run watch",
-    "start:watch:storybook": "cd src/implementations/twig/storybook && npm run start:watch",
```
kalinchernev, line 28: We'll need an additional modification for handling the 2 storybook watch tasks in parallel.
We can either have ports 6006 and 6007 manually in start tasks https://github.com/ec-europa/europa-component-library/pull/1794/files#diff-76163f35cbcc457d90982d6732585f7392a6bf2ae2f44792194991e753b4c9e6R10
Or we can remove the parameter altogether and allow storybook to pick free ports randomly.

Interesting fact, the strings have close hexadecimal code match: 'ec' translates to 6563 and 'eu' to 6575.

src/website/package.json
```diff
@@ -67,6 +67,7 @@
     "gray-matter": "4.0.2",
     "html-loader": "1.3.2",
     "html-webpack-plugin": "4.5.0",
+    "levenary": "1.1.1",
```
kalinchernev: I double-checked this package you mentioned and I believe we don't need to add it. It happens sometimes that items in the node_modules folder get messy in time and probably it's the case.

Please reset the node_modules in your project by `rm -rf node_modules **/node_modules` and `yarn cache clean`.
Thenplease remove the this line as well as the yarn.lock addition. Best to do it manually, as if you try to `yarn remove` when not having the node modules folder things can get messy. Re-try yarn + yarn dist and let me know if you have issues and we can debug together

kalinchernev (changes requested):
All is ok except the yarn dep thingy and ports. The rest is quite ok :+1:

package.json
emeryro: I'd love to use the hexadecimal code of "ec" and "eu" (and then explain to the first one who ask the question "why these random numbers?") :)

src/website/package.json
emeryro: I think that I had a similar issue when generating jsdoc at some time, and indeed removing the root node_modules folder did solve it (i didn't even had to remove anything else).
Also I can be lazy, but we could add a custom command to do the cleanup correctly in one move

package.json
```diff
@@ -21,11 +21,13 @@
     "pretty-check": "prettier --check \"src/**/*.{md,mdx,html,json,yml,js,jsx,scss}\"",
     "publish": "scripts/publish.sh",
     "serve": "cd src/website && npm run serve",
-    "start": "cd src/implementations/twig/storybook && npm run start",
+    "start:ec": "cd src/implementations/twig/storybook/ec && npm run start",
+    "start:eu": "cd src/implementations/twig/storybook/eu && npm run start",
     "start:dev": " lerna --scope \"@ecl/preset-*-core\" run clean && lerna --scope \"@ecl/preset-*-core\" run build && npm run start:watch",
```
emeryro, line 26: shouldn't we have something like
```
"start:ec:dev": ...
"start:eu:dev": ...
```
?

src/implementations/twig/components/site-header-harmonised/site-header-harmonised.story.js
```diff
@@ -33,6 +34,9 @@ const dataG1 = { ...dataGroup1 };
 const dataG2 = { ...dataGroup2 };
 const dataG3 = { ...dataGroup3 };
 
+const system = process.env.STORYBOOK_SYSTEM === 'EU';
```
emeryro: that's a detail, but `system` is a boolean here, so maybe the name could be changed to something more explicit (`isSystemEU` ?)

emeryro (commented):
some small comments too

package.json
planctus: Wow, it took a me while to come out with an answer to this question, but i think it's better leaving it as such.
`start:dev` is what we would use to build the presets in the build folder and then start the watchers tasks for them, It would be basically inconvenient to separate them, because you would have to launch them in two different windows of your terminal. This is also the reason why `start:dev `is not also taking care of launching storybook, which happens instead in `start:watch:ec:storybook`. So to start a storybook instance using the build folder of the presets and the watcher in action, you run `start:dev` and then in another window `start:watch:ec:storybook` or `start:watch:eu:storybook` (this doesn't mean that storybook will hot reload the resources, but these tasks are using the build folder of the presets where the watch scripts are going to update them and you can reload the storybook page and see the changes).
`start:ec` and `start:eu` are using the dist folder of the presets, instead, which is where you get them by running `yarn`.
@kalinchernev correct me if i'm wrong.. :)
In overall i'm a bit confused about the need for these different folders where to build the presets (dist and build) and some complexity and proliferation of the scripts is related to this, couldn't we live out of the dist folder only, i wonder?
`start:ec` could simply do what `start:watch:ec:storybook` is doing, and i guess that we would always want to work with the watchers in action and we could reduce the scripts and simplify their usage.

Anyway, for the sake of this pr i've just fixed the `start:watch:ec:storybook` and `start:watch:eu:storybook` scripts, they were having a wrong path, and the `start:watch` ones in the storybook instances.

kalinchernev: The way I understand the initial question: shall we have 2 separate watch tasks instead of 1.

I think it's useful yes. It should be possible to work on ec in watch mode and in eu only.  I didn't comment on it specifically before because for me a single command `start:dev` should be enough to have all processes spawned in watch mode and just develop in both. Previously both meant presets+storybook, now both would mean ec presets + ec storybook + eu presets + eu storybook.

Unfortunately, the change of default ports 6006 and 6007 did not solve this and only presets watch tasks are started with `yarn start:watch`. So at the moment the `start:dev` is not matching the reason of its existence.

@planctus you have a good point on dist/build folders. It could be easier using build folder everywhere consistently in development mode and dist in production mode.

I'll merge this pull request which introduces `start:eu` well enough for https://github.com/ec-europa/europa-component-library/pull/1790 and i'll check the watch tasks and the other [known issue](https://github.com/ec-europa/europa-component-library/pull/1757#discussion_r526164319) at https://github.com/ec-europa/europa-component-library/pull/1795 where I touch storybook as well.

kalinchernev (approved):

PR #1795, opened 2020-12-02
feat: codepen storybook addon - FRONT-1785

kalinchernev:
Storybook packages have been moved upper in the file structure, putting them next to the website. After all, they deserve this and we won't really have more than 1 implementation layer to keep them next to such.

When testing the newly set panel position, clear previous settings stored in local storage:
Screenshot from 2020-12-08 12-32-18


src/tools/storybook-addon-code/components/CodeToggle.jsx
```diff
@@ -0,0 +1,51 @@
...
+    const { expanded } = this.state;
+
+    return (
+      <IconButton
+        key="code"
+        active={!!expanded}
+        title={!expanded ? 'Show source code' : 'Hide source code'}
+        onClick={this.handleClick}
+      >
+        Prorotype with the source code
```
emeryro:
```suggestion
        Prototype with the source code
```

emeryro (changes requested):
There seems to be an issue with the resources loaded in codepen (css and js).
The path is incorrect (extra `/ec`)
image

kalinchernev:
I haven't refactored watch tasks, which can be done in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1735 discussed today during the day

src/tools/storybook-addon-code/utils/codepen.js
```diff
@@ -0,0 +1,30 @@
+const system = (process.env.STORYBOOK_SYSTEM || 'ec').toLowerCase();
+const baseUrl =
+  ['ec.europa.eu', 'localhost'].indexOf(window.location.hostname) !== -1
+    ? `https://v3--europa-component-library.netlify.app/playground/${system}`
+    : window.location.origin + window.location.pathname;
+
+const replaceLogo = (code) =>
+  code.replace(
+    /dist\/images\/logo([\d-az-]*)\.([\da-z]*)\.svg/gi,
+    `${baseUrl}images/logo/logo$1.svg`
```
papegaill: baseUrl is not added to the logo path on codepen, img is broken: `src="static/media/logo--en.5055ef4f.svg"`

src/tools/storybook-addon-code/utils/codepen.js
```diff
@@ -0,0 +1,30 @@
...
+  );
+
+const replaceIcons = (code) =>
+  code.replace(
+    /dist\/images\/icons\.([\da-z]*)\.svg/gi,
+    `${baseUrl}images/icons/sprites/icons.svg`
```
papegaill: Same for svg icons baseUrl is not added to the logo path on codepen.

src/tools/storybook-addon-code/components/CodeToggle.jsx
```diff
@@ -0,0 +1,51 @@
...
+    const { expanded } = this.state;
+
+    return (
+      <IconButton
+        key="code"
+        active={!!expanded}
+        title={!expanded ? 'Show source code' : 'Hide source code'}
+        onClick={this.handleClick}
+      >
+        Prototype with the source code
```
papegaill: Why replacing with such a long label? Isn’t it better to keep previous title `HTML`

papegaill (changes requested):

src/tools/storybook-addon-code/components/CodeToggle.jsx
papegaill: As discussed, might it make more sense to have the `HTML` tabs in the Addons section next to TWIG, CSS, ... tabs?

src/tools/storybook-addon-code/utils/codepen.js
kalinchernev: Yes, I tried to find a way to solve this within a few hours, but couldn't. V2 has the same issue.
I tried a few ideas, but I haven't figured how to work with the webpack-generated hashes consistently

kalinchernev: i.e. I am aware of this issue but haven't found a solution yet

src/tools/storybook-addon-code/components/CodeToggle.jsx
kalinchernev: I'm ok with any solution desired by the team. The one in this pull request is the existing one from v2. If there is an alignment that it's to be put in the other panel there will be a need of a rewrite. Which I don't mind

kalinchernev: This one I managed to do together with the panel repositioning.

src/tools/storybook-addon-code/utils/codepen.js
kalinchernev: Sorry, I couldn't manage to solve the issue with display of SVG images on codepen.io domain.
I did manage to solve the issues with paths, but not with the cross-domain resource sharing limitations.
I spent the day trying to solve the issue with proxy server, but couldn't, so created https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1908 as a follow-up in order to not keep this pull request blocked for too long.

emeryro (approved):

PR #1804, opened 2020-12-03
fix(date-block): Fix sr-only date label display - FRONT-1841

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/implementations/vanilla/components/date-block/date-block.scss
+++ b/src/implementations/vanilla/components/date-block/date-block.scss
@@ -112,21 +112,21 @@ $_circle-width: 0.6rem;
     padding-bottom: map.get(spacing.$spacing, 'xs');
     padding-top: map.get(spacing.$spacing, 'xs');
   }
+}
 
-  .ecl-date-block__daytime {
-    border: 0;
-    clip: rect(0, 0, 0, 0);
-    /* stylelint-disable-next-line property-no-vendor-prefix */
-    -webkit-clip-path: inset(50%);
-    clip-path: inset(50%);
-    height: 1px;
-    margin: -1px;
-    overflow: hidden;
-    padding: 0;
-    position: absolute;
-    white-space: nowrap;
-    width: 1px;
-  }
+.ecl-date-block__daytime {
+  border: 0;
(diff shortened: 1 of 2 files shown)


kalinchernev (commented):
Thanks @papegaill for debugging this issue and finding a way to solve it. You're right it's confusing.

There are a few interwoven topics:

- A new `ecl-date-block__daytime` has been added instead of `ecl-u-sr-only` => breaking change => either find a way to revert it or move on and https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1835
- The contents of `ecl-date-block__daytime` === the contents of `ecl-u-sr-only`. It's been inlined because utilities hadn't been present yet in the time of the component migration => we should use the sr-only utility directly as https://github.com/ec-europa/europa-component-library/blob/next-v3/src/implementations/vanilla/components/media-container/media-container.scss Possible future changes should be reflected correctly. i.e. in this case, making a mixin with the rule and using it in the date block component.
- It's also possible to use the previous structure before the migration and keep things simple of course

papegaill:
> Thanks @papegaill for debugging this issue and finding a way to solve it. You're right it's confusing.
> 
> There are a few interwoven topics:
> 
> * A new `ecl-date-block__daytime` has been added instead of `ecl-u-sr-only` => breaking change => either find a way to revert it or move on and https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1835
> * The contents of `ecl-date-block__daytime` === the contents of `ecl-u-sr-only`. It's been inlined because utilities hadn't been present yet in the time of the component migration => we should use the sr-only utility directly as https://github.com/ec-europa/europa-component-library/blob/next-v3/src/implementations/vanilla/components/media-container/media-container.scss Possible future changes should be reflected correctly. i.e. in this case, making a mixin with the rule and using it in the date block component.
> * It's also possible to use the previous structure before the migration and keep things simple of course

To me a global mixin seems the best approach 👍 to be discussed also with @planctus as he is the initiator of this change ;)

kalinchernev (approved):

PR #1809, opened 2020-12-07
feat(storybook-system-switcher): Adding widget in the toolbar - FRONT-1836

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


emeryro (changes requested):
Thanks for this. I like the overall display. I wonder if that would be possible to put it in the left navigation area?

Also, there seems to be an issue currently: if I select EU, the EU instance is correctly loaded, but the EC enivronment button is selected by default, and I have no way to go back to EC
image

planctus:
I know @emeryro, i saw that yesterday after the deployment, that's why the pr is still a draft, the thing is that locally this works, so i wonder what's the problem here...

planctus:
It seems it's working now.. :)

.eslintignore
```diff
@@ -8,3 +8,4 @@
 **/static
 src/website/public/playground
 src/tools/**/register.jsx
+src/tools/storybook-addon-system-switcher/src/index.js
```
kalinchernev: I had a similar issue. I think it's better to have linting in these files and help the linter understand the syntax rather than ignoring.

I'd suggest we gradually add eslint instruction files to help the parser know that these files contains jsx syntax in the following way https://github.com/ec-europa/europa-component-library/pull/1795/files#diff-410219d9ab9a7203a40f4816cc7d65a30447a726af7c755b284cd9dead56770a

It's not so bad to have these eslint files in specific places as we already have several for multiple purposes.

src/implementations/twig/storybook/ec/.storybook/preview.js
```diff
@@ -56,4 +55,4 @@ addParameters({
   },
 });
 
-addDecorator(withCssResources);
+export const decorators = [withCssResources];
```
kalinchernev, line 58: Nice change, more consistent syntax with the stories.

.eslintignore
kalinchernev: On a similar note, now having multiple custom addons and 2 instances I think it's worth giving storybook the place it's drawing attention to and in the other pull request I put things in `src/playground`. We can put the addons in this folder too and have 1 single eslint file for all jsx syntax of storybook instead of per-addon. Which will let us remove `src/tools/**/register.jsx` line here too.

src/tools/storybook-addon-system-switcher/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "name": "@ecl/storybook-addon-system-switcher",
+  "version": "3.0.0",
+  "description": "A custom Storybook addon",
+  "main": "src/index.js",
+  "private": true,
+  "keywords": [
+    "storybook-addons"
+  ],
+  "devDependencies": {
+    "@babel/preset-react": "7.12.7",
```
kalinchernev: Why adding this dependency?

kalinchernev: Checking the files locally, it's strange we have it in the other places too. There is only 1 babel configuration which is using this preset and it's `src/website/config/babel.config.js`, so it should be listed and installed only at `src/website/package.json`

src/tools/storybook-addon-system-switcher/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "@storybook/addon-essentials": "6.1.7",
```
kalinchernev: Probably a forgotten attempt as I don't see any usage of the package.

src/tools/storybook-addon-system-switcher/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "react": "^17.0.1"
```
kalinchernev: All items of `devDependencies` can go into `dependencies`. I know it after adding the eslint rules mentioned above

src/tools/storybook-addon-system-switcher/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+import euLogo from '@ecl/resources-eu-logo/logo--en.svg';
+import ecLogo from '@ecl/resources-ec-logo/logo--en.svg';
+
+export const Switcher = () => {
+  const euActive = process.env.STORYBOOK_SYSTEM;
+  const queryString = window.location.search;
+
+  return (
+    <React.Fragment>
+      <IconButton
```
kalinchernev: Definitely not a blocker for this pull request, though as I also had issues in another pull request, I suggest we keep the react component part in separate files with `.jsx` extension.

```js
import { Switcher } from '@ecl/storybook-addon-system-switcher/src/index.js';
```

Will be

```js
import { Switcher } from './Switcher';
```

For example

src/tools/storybook-addon-system-switcher/src/index.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { useGlobals } from '@storybook/api';
+import { IconButton } from '@storybook/components';
+import euLogo from '@ecl/resources-eu-logo/logo--en.svg';
+import ecLogo from '@ecl/resources-ec-logo/logo--en.svg';
+
+export const Switcher = () => {
+  const euActive = process.env.STORYBOOK_SYSTEM;
```
kalinchernev: The variable here is actually `system`.

```js
const system = process.env.STORYBOOK_SYSTEM;
```

src/tools/storybook-addon-system-switcher/src/register.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { addons, types } from '@storybook/addons';
+import { Switcher } from '@ecl/storybook-addon-system-switcher/src/index.js';
```
kalinchernev: I think this line can be improved, please see comment above.

src/tools/storybook-addon-system-switcher/src/index.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
kalinchernev: I am not sure whether there are official conventions regarding the index file of an addon, however in most of the addons I've seen, the `index.js` is a place for making use of `makeDecorator`, i.e. exporting a decorator interface.

If it's only a react component, it could be as single as `Switcher.jsx` and no issues with the linter.

kalinchernev: It's sometimes useful to have an empty `index.js` with `// noop` in oder to avoid breaking module systems but instructing a no-operation approach.

src/tools/storybook-addon-system-switcher/src/register.jsx
```diff
@@ -0,0 +1,14 @@
...
+
+const ADDON_ID = 'ecl/storybook/system-switcher';
```
kalinchernev: Nice approach. Can be split into `constants.js` when necessary.

src/tools/storybook-addon-system-switcher/src/index.js
kalinchernev: A bit tricky part here is that this will be an initial value only because this environment variable will be available only at build time. So loading the first time the contents of build/dist folders it will serve as a sort of a default.

So it might worth treating it as a default case of a last resort. 

```js
const system = window.location.pathname.split('/').filter(p => p)[1] || process.env.STORYBOOK_SYSTEM;
```

I use `window.location.pathname` as a more reliable information source as it's based on runtime.

src/tools/storybook-addon-system-switcher/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const Switcher = () => {
+  const euActive = process.env.STORYBOOK_SYSTEM;
+  const queryString = window.location.search;
+
+  return (
+    <React.Fragment>
+      <IconButton
+        key="ec"
+        active={!euActive}
```
kalinchernev:
```js
const isEc = system === 'ec`
const isEu = system === 'eu`
```
Somewhere after calculating the `system`.

kalinchernev (changes requested):
Pretty good solution, thanks
Adding a few suggestions below

src/tools/storybook-addon-system-switcher/src/index.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { useGlobals } from '@storybook/api';
```
kalinchernev: This line can be removed as it's not used

kalinchernev:
Something which can be surely done later: let's try to make use of [prefetch](https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ) for the ec/eu switcher links and try to have at least part of the resources prepared before the switch in order to make it a bit smoother. Totally ok to be done on a separate task if you don't find a quick win. I didn't succeed with the default way described in the docs.

papegaill:
> Thanks for this. I like the overall display. I wonder if that would be possible to put it in the left navigation area?
> 
> Also, there seems to be an issue currently: if I select EU, the EU instance is correctly loaded, but the EC enivronment button is selected by default, and I have no way to go back to EC
> image

Nice 👍 But I don't particularly like using logos as a switcher. This is not part of graphics used by default in Storybook. 
I personally find that it's visually confusing and doesn't fit correctly in the UI… Why not just use the `EU` | `EC` or a dropdown with a specific title?

Something like this (with a separator): 
switcher-UI-test-1

Or:
switcher-UI-test

What do you think?

src/tools/storybook-addon-system-switcher/src/index.js
kalinchernev: Noticed storybook is using reach/router under the hood and provides some utils which could be useful in working with the paths https://github.com/storybookjs/storybook/blob/5987db49f4264ad88b04b97fe406c3c496e475b5/lib/router/src/utils.ts

kalinchernev (approved):

PR #1810, opened 2020-12-08
feat(controls): POC to test Controls addon on button component - FRONT-1845

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/implementations/twig/components/button/button.story.js
+++ b/src/implementations/twig/components/button/button.story.js
@@ -1,7 +1,5 @@
-import { withKnobs, text, select, boolean } from '@storybook/addon-knobs';
 import { withNotes } from '@ecl/storybook-addon-notes';
 import withCode from '@ecl/storybook-addon-code';
-import { getExtraKnobs, tabLabels, getIconKnobs } from '@ecl/story-utils';
 
 // Import data for demos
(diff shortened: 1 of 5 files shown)


src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,166 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
+    delete data.icon;
+  }
 
   return data;
 };
 
 export default {
   title: 'Components/Button',
-  decorators: [withKnobs, withCode, withNotes],
+  argTypes: {
```
kalinchernev, line 41: `argTypes` seem to be the future, if the guys behind storybook place so much documentation and tools around it, then it will be the standard. We definitely need to make use of those.

It's pity that we don't have the react components anymore, there could have been inferred from the propTypes.

kalinchernev: I really like this structure way better than the knobs. When you think about it, it might not be impossible to extract schema from these getting the object keys + name + type + default value.

kalinchernev (commented):
Good job and a very nice demo, I am all in for going in the direction of having a first step adding types through the `argTypes` and thus having controls vs addons.

PR #1814, opened 2020-12-09
feat(footer): add EU footer standardised - FRONT-1905

emeryro:
[Abstract](https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/6d6e5df9-7474-44db-9f30-a3af1af86d97)

Modifications (similar to EU footer core):
- section ids have been updated to be more consistent (no more jumping from section 3 to section 6 for instance)
- twig parameter "type" has been removed (not used)
- twig parameters added for each link, to allow content before and after it


kalinchernev (approved):
Looks good to me. Not merging myself as I'm not sure whether there is a need for business approval in addition to the peer review.

PR #1818, opened 2020-12-10
feat(controls): Button, migrate knobs to controls - FRONT-1857

papegaill:


package.json
```diff
@@ -37,6 +37,7 @@
   "devDependencies": {
     "@babel/plugin-transform-runtime": "7.12.1",
     "@size-limit/preset-big-lib": "4.9.0",
+    "@storybook/addon-controls": "^6.1.10",
```
kalinchernev: Please update to latest next-v3 and use the updated structure for storybook project. Keep related dependencies within `src/playground` folder. In the case of this addon, it will be `@ecl/twig-ec-storybook` and `@ecl/twig-eu-storybook` using the addon in `main.js`, thus keeping the dep as close as possible to its usage.

kalinchernev: Please keep versions of storybook packages in version sync. Other addons in the repo make use of `6.1.7`. As storybook team releases all addons from a mono-repo like ecl, it happens that assumptions of version sync can introduce bugs.

src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,162 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
-  if (name !== '') {
-    getIconKnobs(data, name, 'ui', 'xs');
-  } else if (name === '' && data.icon) {
-    delete data.icon.name;
-  }
+export const Primary = (args) =>
+  button(prepareControls(Primary, dataPrimary, args));
 
-  return button(data);
+Primary.args = {
```
kalinchernev, line 128: If I understand correctly, `{Component}.args = ...` is a way to set defaults and let storybook infer type. So in the basic case here and in the documentation it will be string which is fine.

However, all the other properties, values and types are in `prepareControls()` and `argTypes`, so I suggest we try to move the setting of specification-based values (defaults) to one of these places too, for consistency.

src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,162 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
-    'variant',
-    [data.variant],
-    data.variant,
-    tabLabels.required
-  );
-
-  data.type = select('type', [data.type], data.type, typeLabel);
-
-  getExtraKnobs(data);
+const prepareControls = (data, args) => {
```
kalinchernev: Just an idea, shall we rename this function throughout and call it `applySpecs()`? Like having a consistent place to merge specs

src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,162 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
+  data.label = args.label;
+  data.disabled = args.disabled;
```
kalinchernev, line 104: We can try having Object.assign() merge before the latter specifics and if/else statements.

src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,162 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
+        category: 'Content',
+      },
+      control: {
+        type: 'text',
+      },
+    },
+    icon_name: {
+      name: 'icon name',
+      type: { name: 'select', required: false },
+      defaultValue: '',
```
kalinchernev: Do we need a default value when it's empty?

src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,162 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
+        defaultValue: { summary: '' },
+        category: 'Content',
+      },
+      control: {
+        type: 'text',
+      },
+    },
+    icon_name: {
+      name: 'icon name',
+      type: { name: 'select', required: false },
```
kalinchernev: Similar question: do we need required to be false if it's not?

src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,162 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
-  decorators: [withKnobs, withCode, withNotes],
+  argTypes: {
+    label: {
+      name: 'label',
+      type: { name: 'string', required: true },
+      defaultValue: '',
+      description: 'The main label of the button',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '' },
```
kalinchernev: Do we need empty summary items if nothing to put in them?

src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,162 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
-export const Primary = () => {
-  const data = prepareButton(dataPrimary);
-  const name = select('icon.name', iconsList, '', tabLabels.optional);
-  if (name !== '') {
-    getIconKnobs(data, name, 'ui', 'xs');
-  } else if (name === '' && data.icon) {
-    delete data.icon.name;
-  }
+export const Primary = (args) =>
+  button(prepareControls(Primary, dataPrimary, args));
```
kalinchernev: There's a mismatch of the signature `prepareControls(data, args)` and usage.
Having the component itself as a first argument looks tricky, what's the goal here?
I think we can skip passing the `args` if we agree to set defaults in the function itself.

src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,162 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
 import dataSearch from '@ecl/specs-component-button/demo/data--search';
 
+import defaultSprite from '@ecl/resources-ec-icons/dist/sprites/icons.svg';
 import button from './button.html.twig';
 import notes from './README.md';
 
 const iconsList = {};
-iconsList.none = '';
-
 uiIcons.forEach((icon) => {
```
kalinchernev, line 17: A bit prematurely extracted and could be put inline as it's used only once. 
In case of considering a one-liner, check this example

```
const arr = ['a','b','c'];
const res = arr.reduce((a,b)=> (a[b]=b,a),{});
```

kalinchernev (commented):
Adding a few more comments completing the review

src/implementations/twig/components/button/button.story.js
papegaill: it was a mistake.. testing code removed

src/implementations/twig/components/button/button.story.js
kalinchernev: By the way, there are only 2 icons at the component's story on v2, not sure whether there are any limitations from the business regarding which icons are to be used here.

src/implementations/twig/components/button/button.story.js
papegaill: yes it showcase an empty string in the table
Screenshot 2020-12-11 at 11 04 47

src/implementations/twig/components/button/button.story.js
papegaill: Updated

src/implementations/twig/components/button/button.story.js
papegaill: Updated

src/implementations/twig/components/button/button.story.js
papegaill: @planctus name it `prepareDataWithArgs()` what do you think ?

src/implementations/twig/components/button/button.story.js
papegaill: I don't think we need it, control can be reset to default state. this is why I removed the none option.

src/implementations/twig/components/button/button.story.js
```diff
@@ -1,7 +1,5 @@
-import { withKnobs, text, select, boolean } from '@storybook/addon-knobs';
```
kalinchernev, line 1: Putting the request here as related to the whole file: let's please disable the knobs as a first step of the migration as discussed in https://github.com/ec-europa/europa-component-library/pull/1820#discussion_r540850867

src/implementations/twig/components/button/button.story.js
papegaill: Not sure to fully understand here, if you ever have a more precise example, or update the PR... this will help me.

kalinchernev: Just as an example, although not a problem not having it. In this case, the `data` is the target: `data = Object.assign(data, args)`

```javascript
var data = { label: 'specsLabel', prop: 'specProp'}
var args = { label: 'argsLabel', prop: 'argProp'}

Object.assign(data, args)

console.log(data) // {label: "argsLabel", prop: "argProp"}
```

src/implementations/twig/components/button/button.story.js
kalinchernev: Could you please share the resolution to this one?
I mean, why isn't this part of `getArgTypes()` instead?

kalinchernev: I'm still wondering whether https://github.com/ec-europa/europa-component-library/blob/f196eae5e77af0150f54b19cc666f67093515dd1/src/implementations/twig/components/accordion/accordion.story.js#L15 would be possible as an approach

src/implementations/twig/components/button/button.story.js
```diff
@@ -11,117 +9,178 @@ import dataCall from '@ecl/specs-component-button/demo/data--call';
...
+    control: {
+      type: 'boolean',
+    },
+  };
+
+  return argTypes;
+};
 
-  getExtraKnobs(data);
+const applySpecs = (data, args) => {
```
kalinchernev: With apologies for the change of mind, but this function is `prepareData()` rather than applying specs. I didn't pay attention in an earlier comment what's happening in the function.

kalinchernev (changes requested):
From me the last request would be to rename `applySpecs` to `prepareData`.

src/implementations/twig/components/button/button.story.js
kalinchernev: Anyways, reading the docs for the controls they have the same approach, so it should be fine after all having per-variant args overriding the default's argTypes

kalinchernev (approved):

PR #1819, opened 2020-12-10
feat(controls): Blockquote, migrate knobs to controls - FRONT-1848

papegaill:


src/implementations/twig/components/blockquote/blockquote.story.js
```diff
@@ -1,28 +1,61 @@
...
-const prepareQuote = (data) => {
-  data.citation = text('citation', data.citation, tabLabels.required);
-
-  data.author = text('author', data.author, tabLabels.required);
+const getArgTypes = () => {
+  const argTypes = {};
+  argTypes.author = {
+    name: 'Author',
+    type: { name: 'string', required: true },
+    description: 'Author of the citation',
```
emeryro: for some reasons, I don't see the description on storybook
image
Is it the same for you?

src/implementations/twig/components/blockquote/blockquote.story.js
```diff
@@ -1,28 +1,61 @@
...
+    table: {
+      type: { summary: 'string' },
+      defaultValue: { summary: '' },
+      category: 'Content',
+    },
+    control: {
+      type: 'text',
+    },
+  };
+  argTypes.citation = {
```
emeryro: a small detail, but can we put the citation before the author in the control list?
That would reflect the normal reading order of the component

emeryro (changes requested):
please fix the conflict reported (I'm not sure what has to be kept here)

src/implementations/twig/components/blockquote/blockquote.story.js
papegaill: Thks forgot to return `argTypes` in getArgTypes function 👍

emeryro (approved):

PR #1820, opened 2020-12-10
feat(controls): Accordion, migrate knobs to controls - FRONT-1847

planctus:


papegaill (approved):

src/implementations/twig/storybook/ec/.storybook/main.js
```diff
@@ -3,12 +3,14 @@ const path = require('path');
 const stories = ['../../../**/!(eu*).story.js'];
 
 const addons = [
+  '@storybook/addon-docs',
```
papegaill: removing this will remove the tabs docs tab

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -43,8 +61,16 @@ export default {
...
+Default.argTypes = argTypes;
 Default.storyName = 'default';
-Default.parameters = { notes: { markdown: notes, json: demoData } };
-Default.decorators = [withKnobs, withCode, withNotes];
+Default.parameters = {
+  notes: {
+    markdown: notes,
+    json: demoData,
+  },
+  knobs: {
```
papegaill, line 67: do we need to specify this? I guess the addon will be removed globaly

planctus: it will, but then once the migration will be completed in the same ticket we will have to remove the knobs addon we will also remove this from each component that has been migrated to controls. I find it weird to have both panels in place with the knobs one empty, just that...

src/implementations/twig/storybook/ec/.storybook/main.js
planctus: not sure when you left this comment, if it's before the conversation we had on teams i guess we can close this.. :)

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -1,40 +1,58 @@
...
+      category: 'Hidden content',
+    },
+    control: {
+      type: 'text',
+    },
+  };
+});
 
-  getExtraKnobs(data);
+const prepareDataWithArgs = (data, args) => {
```
kalinchernev: I am not sure whether the change in name of this function matches what it actually does.
From the pull request for controls in the button component I put a comment about it. If we rename the function, which i think we should do, it should be related to specs, not args.
We might need to align whether this function will be the place where defaults are applied. 
At any case, it will be useful to have consistent naming and operation in this function throughout all stories.

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -1,40 +1,58 @@
...
-      tabLabels.optional
-    );
-  });
+const argTypes = {};
+demoData.items.forEach((item, i) => {
+  argTypes[`toggle${i + 1}`] = {
+    name: `toggle ${i + 1}`,
+    type: { name: 'string', required: true },
+    defaultValue: item.toggle.label,
+    description: `Text of the accordion toggler`,
```
kalinchernev: The plural of toggle is either toggles or toggle switches, but not togglers :)

src/implementations/twig/components/accordion/accordion.story.js
kalinchernev: I like the approach of disable-first and remove-when-done approach here

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -1,40 +1,58 @@
...
-      tabLabels.required
-    );
-    item.level = select(
-      `items[${index}].level`,
-      levels,
-      level,
-      tabLabels.optional
-    );
-  });
+const argTypes = {};
```
kalinchernev: Let's put argTypes generation logic in a function in the same way we keep logic for applying defaults based on specs data and have something like `Default.argTypes = getArgTypes();` Others later if having variants will be `Variant.argTypes = getArgTypesVariant();`

kalinchernev (changes requested):

src/implementations/twig/components/accordion/accordion.story.js
planctus: true!! :)

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -1,40 +1,62 @@
...
+      },
+    };
   });
 
-  getExtraKnobs(data);
+  return argTypes;
+};
+
+const applySpecs = (data, args) => {
+  demoData.items.forEach((item, i) => {
```
papegaill: `data` instead of `demoData` ?

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -1,40 +1,62 @@
...
+    item.toggle.icon.path = defaultSprite;
+    item.toggle.icon.type = 'general';
+    item.toggle.icon.size = 's';
+    if (i === 0) {
+      item.toggle.icon.name = 'digital';
```
papegaill: I don't understand this condition for the toggle `icon.name`.. as it is a `plus` icon for every toggle?

src/implementations/twig/components/accordion/accordion.story.js
planctus: it's a parameter name, we pass demoData to it, which is how we call what we import from the specs when we call the function..

src/implementations/twig/components/accordion/accordion.story.js
planctus: ops, funny, i guess that i was looking at the old-accordion specs when i did this..thanks for noticing it :)

planctus: i fixed it :)

src/implementations/twig/components/accordion/accordion.story.js
kalinchernev: I agree with Alexis, `demoData` should be `data`.
The parameter is `data`, the return value is `data`, whereas `demoData` is a variable from the module's scope, not the `applySpecs()` scope. It's a code smell to mutate variable from outer scope, as well as passing around parameter as a return without any modifications to it.

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -1,40 +1,53 @@
...
+        type: 'text',
+      },
+    };
   });
 
-  getExtraKnobs(data);
+  return argTypes;
+};
+
+const applySpecs = (data, args) => {
```
kalinchernev: With apologies for the change of mind, but this function is `prepareData()` rather than applying specs. I didn't pay attention in an earlier comment what's happening in the function.

kalinchernev (changes requested):
Summary of change requests:
- rename `applySpecs` to `prepareData`
- change `demoData` to `data` for the scope of `prepareData` to avoid scope creep

```js
  data.items.forEach((item, i) => {
    item.toggle.icon.path = defaultSprite;
    item.toggle.label = args[`toggle${i + 1}`];
    item.content = args[`content${i + 1}`];
  });

  return data;
};
```

.eslintrc.js
```diff
@@ -22,5 +22,11 @@ module.exports = {
         'no-undef': 'off',
       },
     },
+    {
```
kalinchernev: Please remove this ignore rule. This is not a solution to the problem at hand, but an illustration of the scope creep commented on earlier.

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -1,40 +1,53 @@
-import { withKnobs, text, select, optionsKnob } from '@storybook/addon-knobs';
 import { withNotes } from '@ecl/storybook-addon-notes';
-import { getExtraKnobs, tabLabels } from '@ecl/story-utils';
 import withCode from '@ecl/storybook-addon-code';
 import defaultSprite from '@ecl/resources-ec-icons/dist/sprites/icons.svg';
-import demoData from '@ecl/specs-component-accordion/demo/data';
```
kalinchernev, line 6: First, leave this be `demoData`, there is no issue with it.

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -1,40 +1,53 @@
...
-      { display: 'inline-radio' },
-      tabLabels.required
-    );
-    item.level = select(
-      `items[${index}].level`,
-      levels,
-      level,
-      tabLabels.optional
-    );
+const getArgTypes = () => {
```
kalinchernev: Two, change this function to accept a parameter: `data`. This will bind the `data` in `getArgTypes` function scope, the contents of the curly brackets after the function name.

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -43,8 +56,16 @@ export default {
   title: 'Components/Accordion',
 };
 
-export const Default = () => accordion(prepareAccordion(demoData));
-
+export const Default = (args) => accordion(prepareData(data, args));
+Default.argTypes = getArgTypes();
```
kalinchernev: Three, invoke `getArgTypes()` with a value for the input parameter. In the scope of the module `accordion.story.js` it's going to be `demoData`, therefore

```javascript
Default.argTypes = getArgTypes(demoData);
```

Inside the `getArgTypes`, `demoData` will be replaced with `data` and there will be no shadowing or scope creep.

The main benefit: you can extract the function out of the module at any time without any issues. But you also make the logic flow clear and do not risk accidental changes of the data.

src/implementations/twig/components/accordion/accordion.story.js
```diff
@@ -43,8 +56,16 @@ export default {
...
 Default.storyName = 'default';
-Default.parameters = { notes: { markdown: notes, json: demoData } };
```
kalinchernev, line 49: Lastly, this is also `demoData` from module scope.

src/playground/addons/storybook-addon-system-switcher/systemSwitcher.jsx
```diff
@@ -3,10 +3,9 @@ import { IconButton, Separator } from '@storybook/components';
 
 const Switcher = () => {
   const queryString = window.location.search;
-  let system = window.location.pathname.split('/').filter((p) => p)[1];
-  if (!system) {
-    system = process.env.STORYBOOK_SYSTEM ? 'eu' : 'ec';
-  }
+  const system =
```
kalinchernev: Noticing a change here in the switcher. We need to have https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1838 soon in order to avoid difference in system extraction logic.

Here's the way I did it in the code addon, taking env variable first and first system-specific match from path as a second.

```javascript
const system = (
  window.location.pathname
    .split('/')
    .find((part) => part === 'ec' || part === 'eu') ||
  process.env.STORYBOOK_SYSTEM
).toLowerCase();
```

kalinchernev (changes requested):

papegaill (approved):

PR #1822, opened 2020-12-11
feat(controls): Link, migrate knobs to controls - FRONT-1853

planctus:


src/implementations/twig/components/link/link.story.js
```diff
@@ -26,86 +25,125 @@ const withParagraph = (story) => {
...
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '' },
+        category: 'Link',
+      },
+      control: {
+        type: 'text',
+      },
+    },
+    aria_label: {
```
emeryro: I don't think we need to keep this control for aria-label; that's not something directly visible

src/implementations/twig/components/link/link.story.js
```diff
@@ -26,86 +25,125 @@ const withParagraph = (story) => {
...
+        type: 'text',
+      },
+    },
+    icon_position: {
+      name: 'icon position',
+      type: { name: 'select', required: false },
+      description: 'Position of the icon (required to show the icon)',
+      table: {
+        type: { summary: 'string' },
+        category: 'Link',
```
emeryro: we should put the icon position in the icon category I guess

src/implementations/twig/components/link/link.story.js
```diff
@@ -26,86 +25,125 @@ const withParagraph = (story) => {
...
+  return {
+    label: {
+      name: 'label',
+      type: { name: 'string', required: true },
+      defaultValue: data.link.label,
+      description: 'The main label of the button',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '' },
+        category: 'Link',
```
emeryro: This category is called "Content" in other PR. 
I think it's better to keep the same terminology everywhere

emeryro (changes requested):

src/implementations/twig/components/link/link.story.js
planctus: i've reproduced the knobs we have in ecl-v2, where the aria label knob is present. Anyway, ok.

src/implementations/twig/components/link/link.story.js
planctus: Questionable, the icon_position is a property of the link, not one belonging to the icon, indeed.

src/implementations/twig/components/link/link.story.js
planctus: Not sure it makes sense as "content", i believe these labels should reflect somehow the component, or "object" they belong to.

src/implementations/twig/components/link/link.story.js
papegaill: for the button, I put everything in the icon category

src/implementations/twig/components/link/link.story.js
```diff
@@ -26,86 +25,125 @@ const withParagraph = (story) => {
...
+    },
+  };
+};
 
-// Preserve the adapted specs.
-const prepareLink = (data) => {
-  let typeLabel = tabLabels.required;
-  if (data.link.type === 'default' || data.link.label === '') {
-    typeLabel = tabLabels.optional;
+const applySpecs = (data, args) => {
```
papegaill: prepareData ?

planctus: just pushed it like that :)

src/implementations/twig/components/link/link.story.js
```diff
@@ -26,86 +25,110 @@ const withParagraph = (story) => {
...
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '' },
+        category: 'Content',
+      },
+      control: {
+        type: 'text',
+      },
+    },
+    icon_position: {
```
papegaill: I think it would be good to align with the icon controls, see button controls. I think the control for the position should be last.  I'm also using inline-select type for this field control. Also in the buttons the position of the icon by default is "after", which does not seem to be the case with links, we can maybe update the default value for links?

I'm using "icon name" and here you use "icon (sample)" for the icon control label. 

I think we have to align either the links or the buttons and use the same approach for the icons controls?

papegaill (changes requested):

src/implementations/twig/components/link/link.story.js
```diff
@@ -26,86 +25,110 @@ const withParagraph = (story) => {
...
-uiIcons.forEach((icon) => {
-  iconsList[icon] = icon;
-});
+const iconsList = uiIcons.reduce((a, b) => {
+  a[b] = b;
+  return a;
+}, {});
+
+const getArgTypes = (data) => {
+  return {
```
papegaill, line 34: I like this return statement, I think we should align existing migrated component

src/implementations/twig/components/link/link.story.js
planctus: I reproduced what we have in ECL v2, which it was what we agreed to do...
There is no default for the icon_position in the link component, i see in the "docs" as such but the documentation is not aligned with the template.
I'll align this pr with what has been merged already, then, but i'm not sure whether to fix the template by setting a default value for the icon_position or fixing the documentation, instead.
I'll add the default value to the template, i think it makes sense to have it.

PR #1827, opened 2020-12-15
chore: global system context - FRONT-1838

kalinchernev:
- New `getSystem()` utility in ecl-builder. Because it's not a dom-related utility and is currently relevant only in styles script in the package
- There are tests for the new utility. It's written in a compatible way both node native and ES6 through transpilation, as seen in the tests.
- Updated storybook addons to use the new utility for consistency
- Removed unused storywrapper addon
- Previously demoed `postcss-env-function` implementation does not work with current setup because it requires postcss 7 and we are at version 8.
- The [`functions`](https://sass-lang.com/documentation/js-api#functions) feature has been used to pass the new `getSystem()` utility to SCSS layer.
- Tests for the `functions` part are not 100% following conventions for unit testing because they run code already covered by `sass`, yet, it's a useful illustration of the feature and allows faster iterations on it rather than ecl-builder config only. I did try testing the signature of entry, dest and options, but paths never get resolved correctly, even when I took ec-core config as a stub. Decided to not spend too much time on signature testing, as it would either create files and slow down or need of fs module mocking.


src/tools/builder/builder.test.js
```diff
@@ -0,0 +1,189 @@
...
+
+    it('should be able to expose getSystem utility in scss source code', async () => {
+      let system = getSystem();
+      const options = {
+        data: `
+            $system: getsystem();
+            .system-specific { content: $system; };
+        `,
+        functions: {
+          'getSystem()': () => new sass.types.String(system || ''),
```
emeryro:
```suggestion
          'getsystem()': () => new sass.types.String(system || ''),
```
I guess...

emeryro (approved):

PR #1828, opened 2020-12-15
feat(inpage-navigation): Migrate knobs to controls - FRONT-1852

papegaill:


src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
```diff
@@ -1,112 +1,127 @@
...
-  // Prevent the story from being reloaded.
-  return false;
+const getArgTypes = (data) => {
+  const argTypes = {};
+  data.links.forEach((item, i) => {
+    argTypes[`heading${i + 1}`] = {
+      name: `heading ${i + 1}`,
+      type: { name: 'string', required: true },
+      defaultValue: item.label,
+      description: `The label for heading  ${i + 1}`,
```
planctus: There are two spaces in this string, but maybe like this they are confusing, i mean that the "title" of the control is actually matching its value, and if you edit that it does not make much sense anymore, in v2 they were: `First element label` `Second element label`, etc.. i know that in your loop it makes it a little bit trickier but i think it's worth, you use then the "derived" string in two places, the `name` and the `description`

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
```diff
@@ -1,112 +1,127 @@
...
+    defaultValue: false,
+    description: 'Inject a test content block in the main column',
+    table: {
+      category: 'Test content',
+    },
+    control: {
+      type: 'boolean',
+    },
+  };
+
```
planctus, line 58: Not sure if you tried and then decided not to reproduce the demo as we had it with knobs, but we introduced in v2 a functionality that here we don't demonstrate, the fact that if a new h2 with an id is injected in the main content of the page the in-page navigation is updated adding a corresponding element, and viceversa, when a block of h2 and content "connected" to an in-page navigation element is removed also the linking element in the in-page gets removed.

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
```diff
@@ -1,112 +1,127 @@
...
+    description: 'Inject a test content block in the left sidebar',
+    table: {
+      category: 'Test content',
+    },
+    control: {
+      type: 'boolean',
+    },
+  };
+  argTypes.blockMain = {
+    name: 'main content block',
```
planctus: Maybe here we have the chance to better explain the reason of these button knobs we had. Basically we want to demonstrate that the in-page recalculates the position of the linked content when something gets added in the dom.
We could maybe say in the title what we are currently saying the description and use the description to say that the in-page navigation won't loose the correct reference to the linked elements.

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
```diff
@@ -1,112 +1,127 @@
...
+  parameters: {
+    knobs: {
+      disable: true,
+    },
+  },
+};
 
+export const Default = (args) => {
+  const navHtml = inpageNavigation(prepareData(demoData, args));
+  const contentHtml = prepareHtmlContent(demoData, args);
```
planctus: I was initially confused about the reason why we were passing args to this function, but i see that also previously the change of the label in the knob was also producing a change in the heading in the content, i find it confusing, there is no reason why it should happen, the link between the two is the anchor value, the label of the in-page link is independent.

But then i also noticed that when migrating the component the specs were enriched by some content not in place in v2, like the lorem ipsum text, which we are not using here, maybe we could.

planctus (commented):
I put some comments, i'll let you evaluate them, the demo is working fine but we miss some of the pre-existing features showcase, maybe not anything we could reproduce with controls, though.

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
papegaill: As it's not possible to have button type controls, I didn't go further on this feature. 
I also don't find it particularly relevant. what do you think?

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
papegaill: I also wondered about this point. However, this is currently the case on v2, by changing the label of an element, it's updated in the nav and the content, I personally find that it makes more sense than just updating the navigation labels.

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
papegaill: Updated as suggested, not sure the description is super clear, what do you think?

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
kalinchernev: I didn't know the idea behind the buttons knobs injecting markup to demonstrate such a sophisticated behavior.
Feels a bit unnecessary to me, but I don't have a strong argument against it. I spent some time in the docs and the issue queue of storybook, there is no button control and there is no way to create a custom control at the moment. (at least not in a stable way).

I'm personally ok to keep the knobs with the buttons if it's a high importance demo. If we decide to keep the knobs and the buttons, let's please include texts explaning the feature rather than lorem ipsum which doesn't bring value to the demo

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
```diff
@@ -1,112 +1,130 @@
...
+
+// Add generic container in the left or main column.
+const blockHandler = (region, state) => {
+  if (state) {
+    return `<div class="ecl-u-bg-grey-25 ecl-u-mb-m ecl-u-pa-l">
+              <h4 class="ecl-u-type-heading-4">This is an injected block in the ${region} column</h4>
+            </div>`;
+  }
+
   return false;
```
kalinchernev: What is the idea behind returning the boolean false here? To me it seems the function is used for returning strings/markup. 

If my assumption is correct, then `${leftBlock || ''}` can be `${leftBlock}` and default here is an empty string. Not a huge difference but the function would have a consistent return type.

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
```diff
@@ -1,112 +1,130 @@
...
 };
 
-export default {
-  title: 'Components/Navigation/Inpage navigation',
+// Prepare data for the navigation.
+const prepareData = (data, args) => {
+  data.icon_path = defaultSprite;
+  data.links.forEach((item, i) => {
```
kalinchernev, line 76: This part of the code is not needed I think as specs already have data for these so `navHtml` using demoData, they will be there without defaults of argtypes

```js
  data.links.forEach((item, i) => {
    item.label = args[`heading${i + 1}`];
  });
```

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
```diff
@@ -1,112 +1,130 @@
...
+  }
+
   return false;
 };
 
-export default {
-  title: 'Components/Navigation/Inpage navigation',
+// Prepare data for the navigation.
+const prepareData = (data, args) => {
+  data.icon_path = defaultSprite;
```
kalinchernev: The website has `correctSvgPath()` for a similar future repetitive issue

Screenshot from 2020-12-16 15-49-46

`src/playground/addons/story-utils/src/index.js` could provide a similar utility but for `icon_path`

kalinchernev: If done, then stories won't need to import the default sprite and 

```js
const navHtml = inpageNavigation(prepareData(demoData, args));
```

Would become 

```js
const navHtml = inpageNavigation(correctSvgPath(demoData));
```

I.e. no need for more transformations i think

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
```diff
@@ -1,112 +1,130 @@
...
-  button(btnIdRemoveLabel, btnIdRemoveHandler, tabLabels.cases);
+// Prepare Html for the main content.
+const prepareHtmlContent = (data, args) => {
+  let html = '';
+  const lorem = loremIpsum({ count: 25 });
 
-  let pageFillerHtml = '';
-  demoData.links.forEach((content) => {
-    pageFillerHtml += content.item;
+  data.links.forEach((item, i) => {
```
kalinchernev: There's a difference in style compared to the file and even the concatenation items themselves. i.e no need for concatenation, the string literal will do just fine here too

```js
const prepareHtmlContent = (data) => {
  const lorem = loremIpsum({ count: 25 });
  return data.links.map(({ label }, index) => {
    return `
    <h2 class="ecl-u-type-heading-2" id="inline-nav-${index}">${label}</h2>
    <p class="ecl-u-type-paragraph-m">${lorem}</p>
    <p class="ecl-u-type-paragraph-m">${lorem}</p>
    `;
  });
};
```

Btw, I'm using the demo data directly

kalinchernev (changes requested):
Few comments below

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
kalinchernev: Texts for the demo are ok for me, thanks

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
kalinchernev: I agree it's a bit confusing to have binding of args to data in this way. I tried locally and it seems to me demo data is good enough as Davide suggests

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
planctus: Well, it's a feature of the component that is otherwise hidden and difficult to discover, but since this is maybe the only use case where we would find some irreplaceable knob i wouldn't really keep the knobs addon in place just for this. Fine to me to give up about them.

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
papegaill: This is necessary to link the controls to the template value. Otherwise the controls would not work.

src/implementations/twig/components/inpage-navigation/inpage-navigation.story.js
papegaill: Thks, correctSvgPath utils added, and I also updated existing migrated components https://github.com/ec-europa/europa-component-library/pull/1828/commits/f0ff58e91c92d2a5ff76beca9aa83cba9f41165b
prepareData is needed to sync `data` with their `args`.

kalinchernev (approved):

PR #1829, opened 2020-12-16
feat(menu): Remove knobs - FRONT-1854

papegaill:


emeryro (changes requested):
What could be nice would be to have a control to update the site name, and/or remove it. It is optional in the data, and we have been asked in the past to have an example of menu without site name

papegaill:
> What could be nice would be to have a control to update the site name, and/or remove it. It is optional in the data, and we have been asked in the past to have an example of menu without site name

I added site_name control, by putting an empty string the site name disappears, 
so I didn't add an additional toggle to display it or not. what do you think?

emeryro (approved):

PR #1830, opened 2020-12-16
feat(pagination): Remove knobs - FRONT-1855

papegaill:


src/implementations/twig/components/pagination/pagination.story.js
```diff
@@ -1,101 +1,27 @@
...
 
   return data;
 };
 
 export default {
   title: 'Components/Navigation/Pagination',
+  decorators: [withNotes, withCode],
+  parameters: {
+    knobs: { disable: true },
+    controls: { hideNoControlsWarning: true },
```
emeryro: isn't it cleaner to use `controls: { disable: true }` when there are controls?

papegaill: yes indeed much better 👍

papegaill:
@planctus, @emeryro I needed to update twig template file to fix display of the icons. See: https://github.com/ec-europa/europa-component-library/pull/1830/commits/8714f032e730bb0f710b426bdde59c48a9494301
I spent a some time investigating this and I wonder how this was working until now to properly display the icons.

I'm also updating the already migrated accordion component to use `correctSvgPath` like other components.

src/implementations/twig/components/pagination/pagination.html.twig
```diff
@@ -50,12 +50,14 @@
           {% if _item.aria_label is defined and _item.aria_label is not empty %}
             {% set _link = _link|merge({ aria_label: _item.aria_label }) %}
           {% endif %}
+          {% set _icon = {} %}
           {% if _item.link.icon is defined and _item.link.icon is not empty %}
             {% set _icon = _icon_path is not empty ? _item.link.icon|merge({ path: _icon_path }) : _item.link.icon %}
             {% set _link = _link|merge({ 'icon': _icon }) %}
           {% endif %}
           {% include '@ecl/link/link.html.twig' with _item.link|merge({
```
planctus: didn't we say to just pass the local variables where we already merged the item.link objects? just like this:
```
{% include '@ecl/link/link.html.twig' with {
  link: _link|default({}),
  icon: _icon|default({}),
  extra_classes: "ecl-pagination__link",
} only %}
```

and yeah, doing like this we don't need to initialise the `_icon` variable as an empty object

src/implementations/twig/components/pagination/pagination.html.twig
```diff
@@ -50,12 +50,14 @@
           {% if _item.aria_label is defined and _item.aria_label is not empty %}
             {% set _link = _link|merge({ aria_label: _item.aria_label }) %}
           {% endif %}
+          {% set _icon = {} %}
           {% if _item.link.icon is defined and _item.link.icon is not empty %}
             {% set _icon = _icon_path is not empty ? _item.link.icon|merge({ path: _icon_path }) : _item.link.icon %}
             {% set _link = _link|merge({ 'icon': _icon }) %}
```
planctus: This line can be removed, we have two separated variables for the two objects link and icon, this merge does not make sense.

planctus (commented):
Thanks for the fix on the template, there are just these two things that we can further do..

planctus (approved):

PR #1832, opened 2020-12-16
feat(preset): remove preset full - FRONT-1916

emeryro:
- remove unused preset ec-preset-full
- update migration documentation
- rename folders in dist, to be more consistent with their content


docs/Migrating-v3.md
```diff
@@ -116,6 +116,20 @@ Example: `@ecl-twig/ec-component-accordion/ecl-accordion.html.twig` => `@ecl/acc
 
 [ECL Compliance component](https://github.com/ec-europa/ecl-twig/tree/master/src/ec/packages/ec-component-ecl-compliance) has been removed.
 
+## Presets / themes
+
+A new concept of "theme" has been introduced, which complete the presets used in ECL 2. Presets are still used under the hook (one for each theme), but what is exposed to end users is now the following:
```
kalinchernev:
```suggestion
A theme layer has been introduced in ECL v3. It sits on top of the components layer and transforms variables into values targeting specific use cases called presets. Presets are ready-made ECL releases serving for different purposes and target audiences. The following presets are available:
```

docs/Migrating-v3.md
```diff
@@ -116,6 +116,20 @@ Example: `@ecl-twig/ec-component-accordion/ecl-accordion.html.twig` => `@ecl/acc
...
+
+- ec-core: contains everything related to EC sites
+- eu-core: contains everything related to EU sites
+
+Some of the old v2 presets have been removed or modified:
+| Modified presets | Status |
```
kalinchernev:
```suggestion
| Preset | Modifications |
```

As it's already mentioned that they have been modified, the important info is how

docs/Migrating-v3.md
```diff
@@ -116,6 +116,20 @@ Example: `@ecl-twig/ec-component-accordion/ecl-accordion.html.twig` => `@ecl/acc
...
+| - | - |
+| ec-preset-full | preset removed (not used anymore) |
```
kalinchernev, line 129: There are eu presets too https://www.npmjs.com/package/@ecl/eu-preset-full/v/2.10.1?activeTab=dependencies

docs/Migrating-v3.md
kalinchernev: Reading the overall text makes me think whether it's worth putting information regarding themes here. There's nothing essential and the focus is vice versa: it's on the presets and the themes is under the hood.

kalinchernev (approved):

PR #1833, opened 2020-12-16
feat(controls): Card, replacing knobs with controls - FRONT-1858

planctus:
I've spent some time in checking what was really needed of the verbose syntax we are using to define our argTypes, i could remove some of the redundant properties still getting the same result.
I have also noticed that we can enrich the summary describing the prop type with basically what we want, i've then described a bit more deeply the structure of these arrays that we offer as a simple comma separated string.
To properly use this simplified control a bit of logic had to be added to handle them regarding the data we pass to the template..


src/implementations/twig/components/card/card.story.js
```diff
@@ -18,123 +9,187 @@ import dataCardTile from '@ecl/specs-component-card/demo/data--tile';
...
+    },
+  };
+  if (data.card.meta) {
+    argTypes.meta = {
+      name: 'meta (comma separated)',
+      type: { name: 'array' },
+      defaultValue: data.card.meta,
+      description: 'The card meta',
+      table: {
+        type: { summary: 'array' },
```
emeryro, line 45: it may be better to put the "comma separated" information here instead of in the name (same for other controls)

src/implementations/twig/components/card/card.story.js
```diff
@@ -18,123 +9,187 @@ import dataCardTile from '@ecl/specs-component-card/demo/data--tile';
...
+  if (data.card.infos) {
+    const infos = data.card.infos.map(({ label }) => label);
+    argTypes.infos = {
+      name: 'infos (comma separated)',
+      type: 'array',
+      defaultValue: infos,
+      description: 'Additional elements like a location or a date',
+      table: {
+        type: {
+          summary: 'array of objects [{ label: "", path: "", icon: {} }]',
```
emeryro: I don't think we need such a complex structure here. The path is not visible for instance, and having a default icon could be enough (that's just a showcase of the component, having one icon or another here won't make a big difference)

src/implementations/twig/components/card/card.story.js
```diff
@@ -18,123 +9,187 @@ import dataCardTile from '@ecl/specs-component-card/demo/data--tile';
...
+  }
+  if (data.card.tags) {
+    const tags = data.card.tags.map(({ label }) => label);
+    argTypes.tags = {
+      name: 'tags (comma separated)',
+      type: 'array',
+      defaultValue: tags,
+      description: 'Tags to be placed at the bottom of the card.',
+      table: {
+        type: { summary: 'array of objects [{ label: "", path: "" }]' },
```
emeryro: same remark for the structure here

src/implementations/twig/components/card/card.story.js
```diff
@@ -18,123 +9,187 @@ import dataCardTile from '@ecl/specs-component-card/demo/data--tile';
...
+  if (data.card.links) {
+    const links = data.card.links.map(({ label }) => label);
+    argTypes.links = {
+      name: 'links (comma separated)',
+      type: 'array',
+      defaultValue: links,
+      description: 'Links to be placed at the bottom of the card.',
+      table: {
+        type: {
+          summary: 'array of objects [{ label: "", path: "", type: "" }]',
```
emeryro: same here

emeryro (changes requested):
Thanks for this.
As written below, the work you did to bring some more complex structure in controls may be useful in some cases, but here I feel like it's a little overkill

emeryro (approved):

PR #1834, opened 2020-12-16
feat(checkbox): Migrate knobs to controls- FRONT- 1859

papegaill:


src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
-      defaultSprite,
-      tabLabels.required
-    );
-  });
+  return argTypes;
+};
 
-  getExtraKnobs(data, true);
+const prepareData = (data, args) => {
+  data.icon_path = defaultSprite;
```
planctus: We miss the icons in the demo, because the icon_path is a property of the single checkbox item, questionable i'd say, but unless we want to add a parameter in the group and then spread it ourselves into the items, we have to set the right icon_path in the items.

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
+  data.label = args.label;
+  data.helper_text = args.helper_text;
+  data.invalid_text = args.invalid_text;
+  data.required_text = args.required_text;
+  data.optional_text = args.optional_text;
+  data.invalid_text = args.invalid_text;
+  data.required = args.required;
+  data.invalid = args.invalid;
```
planctus, line 124: But, for how you built the args you could avoid all these assignments and simply merge the two objects:
` return {...data, ...args};`

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
+      type: { summary: 'string' },
+      defaultValue: { summary: '' },
+      category: 'Content',
+    },
+    control: {
+      type: 'text',
+    },
+  };
+  argTypes.optional_text = {
+    name: 'optional text',
```
planctus: In cases like this, where the argType property is equal to its name you can omit the last one.

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
+    name: 'legend',
+    type: { name: 'string', required: true },
+    defaultValue: data.label,
+    description: 'The heading of the checkbox group',
+    table: {
+      type: { summary: 'string' },
+      defaultValue: { summary: '' },
+      category: 'Content',
+    },
+    control: {
```
planctus: I investigated a bit the possibilities of reducing the syntax we use for defining the argTypes and we basically don't need to specify the control, it is assumed from the type key of the argType you define, and string is fine enough for storybook, i haven't tried but i bet the same is also true for the booleans you have.

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
+      type: 'text',
+    },
+  };
+  argTypes.helper_text = {
+    name: 'group helper text',
+    type: { name: 'string' },
```
planctus: This can be simply `type: 'string',` (in all the cases where `name` it's the only key of the type object)

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
+    },
+    control: {
+      type: 'text',
+    },
+  };
+  argTypes.optional_text = {
+    name: 'optional text',
+    type: { name: 'string' },
+    defaultValue: data.optional_text,
+    description: 'Indicator text stating if input fields are optional',
```
planctus:
```suggestion
    description: 'Text to be shown when the input field is optional',
```

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
+    },
+    control: {
+      type: 'text',
+    },
+  };
+  argTypes.required_text = {
+    name: 'required text',
+    type: { name: 'string' },
+    defaultValue: data.required_text,
+    description: 'Indicator text stating if input fields are mandatory',
```
planctus:
```suggestion
    description: 'Text to be shown when the input field is mandatory',
```

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
+    },
+    control: {
+      type: 'text',
+    },
+  };
+  argTypes.invalid = {
+    name: 'invalid',
+    type: { name: 'boolean' },
+    defaultValue: data.invalid,
+    description: 'Make the checkbox group invalid',
```
planctus: I see that the others are "generic" enough to be reused in all the forms elements (this is the typical case then for a "utility" to be put in place)
```suggestion
    description: 'Marks the input field as invalid.',
```

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
+    },
+    control: {
+      type: 'boolean',
+    },
+  };
+  argTypes.required = {
+    name: 'required',
+    type: { name: 'boolean' },
+    defaultValue: data.required,
+    description: 'Make the checkbox group required',
```
planctus: Same as before, i would make them generic and provide them through a utility, these will be the same in all the form elements.

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -14,35 +7,137 @@ import dataDefault from '@ecl/specs-component-checkbox/demo/data--default';
...
+    table: {
+      type: { summary: 'boolean' },
+      defaultValue: { summary: false },
+      category: 'State',
```
planctus: Maybe plural, `States`..?

planctus (changes requested):

src/implementations/twig/components/checkbox/checkbox.story.js
planctus: I did this work of returning the form elements controls through a utility in https://github.com/ec-europa/europa-component-library/pull/1835 ;) 
I'm seeing that for knobs we had two different ones, one for the form element and one for the group, you don't have the `disabled` for instance in the form groups, maybe we can now simply disable the control in the stories of the "form groups"...

src/implementations/twig/components/checkbox/checkbox.story.js
planctus: For the way you defined your argTypes, instead of copying the values one by one in `data` for then returning it we can do:
`return Object.assign(data, args);` 

I did the same in: https://github.com/ec-europa/europa-component-library/pull/1835 using the utility i've added.

src/implementations/twig/components/checkbox/checkbox.story.js
papegaill: Interesting but the underscore remain, I'll therefore leave it like this.

src/implementations/twig/components/checkbox/checkbox.story.js
papegaill: Allélluia 👍 I have also updated blockquote and date-block to use` Object.assign`

planctus (approved):

PR #1835, opened 2020-12-17
feat(controls): Datepicker, replacing knobs with controls - FRONT-1860

planctus:


src/implementations/twig/components/datepicker/datepicker.story.js
```diff
@@ -1,45 +1,44 @@
...
-    tabLabels.required
-  );
-
-  data.placeholder = text('placeholder', data.placeholder, tabLabels.required);
-
-  getExtraKnobs(data);
+const getArgTypes = (data) => {
+  return {
+    ...getFormControls(data),
+    autoinit: {
```
papegaill, line 12: Do we really need to have autoinit control ?

src/playground/addons/story-utils/index.js
```diff
@@ -525,14 +525,97 @@ export const getLinkKnobs = (data) => {
...
-  data['_compliance_'] = boolean(
-    '_compliance_',
-    data['_compliance_'],
-    tabLabels.checks
-  );
-
-  return data;
+export const getFormControls = (data) => {
+  return {
+    label: {
```
papegaill: Thanks for the utils !! 

I updated checkbox accordingly, see: https://github.com/ec-europa/europa-component-library/pull/1834
You can now maybe use  `getFormControls(data, 'element')` in this case.

On checkbox PR, I removed the ending dots on the description since we have never use them on existing migrated components.

src/implementations/twig/components/datepicker/datepicker.story.js
```diff
@@ -1,45 +1,44 @@
...
+        defaultValue: { summary: false },
+        category: 'States',
+      },
+    },
+  };
+};
 
-  return data;
+const prepareData = (data, args) => {
+  return Object.assign(correctSvgPath(data), args);
```
papegaill, line 27: nice 👍 `correctSvgPath`

papegaill (changes requested):

src/implementations/twig/components/datepicker/datepicker.story.js
planctus: i think so because this is the only component that has this opt-in for the autoinit, i think it's important to demo it

papegaill (approved):

PR #1837, opened 2020-12-18
feat: ecl-builder watch task - FRONT-1735

kalinchernev:
## Change highlights

Project-level:

- `start:ec` and `start:eu` are in watch mode by default, the rest has been removed
- storybook instances are treated as apps and have their ports configured in a central place
- lerna scope execution is not used in order to keep information stream in terminal visible

ECL builder:
- new script added: a watch task which is a thin wrapper on top of browser-sync

Storybook instances:
- a single npm script
- start scripts have been migrated to node child process spawning in order to have their ports set dynamically

Presets:
- chokidar has been replaced with browser-sync: an abstraction of chokidar + ws
- `build` folders are created on postinstall scripts just like the `dist`

## Other notes

Changes done in this pull request change only the ways CSS/JS assets are visualized in a proxy instance of storybook. There are no changes in storybook scripts or webpack configurations. This is an important note to set expectations correctly:

- changes in `*.story.js` files will trigger hot module replacement coming from default storybook webpack configurations
- changes in SCSS files configured through ecl-builder watch task will trigger CSS injection if configured properly from `.reload` property. These should result in `(ec|eu)-core.css` files being replaced inside the proxy without page reload which feels like webpack+react hot module replacement in storybook, but it's not.
- changes in JS files (which are not sourced by storybook) will always result in a full page reload. Newly built assets sourced by storybook's start scripts from `./build` are the updated compiled version of ECL js file.

In short: files sourced by storybook configurations will be handled by storybook's webpack setup, whereas assets sourced by ecl-builder's watch task will be handled by ecl-builder's rollup/babel setup.

## Ambitious?

You might wonder how this setup can be used with a Drupal website instead of a storybook because you see it in the docs?

- clone https://github.com/openeuropa/oe_theme
- follow installation instructions (I used docker compose option)
- copy/paste `builder` folder somewhere in this folder. Either symlink or make a workspace
- change `@ecl/builder` version to 3.0.0
- add a new task `"watch": "ecl-builder watch"`
- add the new config to `ecl-builder.config.js`
- either bump node to v12 in the theme, or refactor back recursive folder creation to be a promise with `mkdirp`

```js
  watch: {
    init: {
      proxy: "127.0.0.1:8080",
    },
    handlers: [
      {
        pattern: `${path.resolve(__dirname, ".")}/sass/**/*.scss`,
        events: [
          {
            on: "change",
            name: "styles changes",
            command: "npm run build:styles",
            message: "New styles ready",
            reload: "*.css",
          },
          ,
        ],
      },
    ],
  },
```

Start the watch task and make a change in a file and see it in `css/style-ec.css`. I didn't go for the full port exposing dance with docker-compose but started the node process on the host machine directly and it was ok.

I didn't spent too much time checking why changes did not appear on the site, seeing changes reflect in the file system and browsersync notifying the proxy was good enough for me.


planctus:
A biiiiig wow and thanks @kalinchernev , this is so cool!!!! :D 
i've been just enjoying it for now, will look better at the code, mostly to understand it, but it's a super mega improvement of the development experience in ECL, really, so good!!! Thaaaaanks :)

emeryro (approved):
that's indeed much cleaner this way, thanks!

PR #1839, opened 2020-12-18
feat(controls): Search form, replacing knobs with controls - front-1863

planctus:
By chance i've tried to add also disabled and invalid use cases for the text input. It seems that we have, at least for the invalid one, a style issue. The red border is only visible when hovering the input, the issue seems inherited from v2.


papegaill:
> By chance i've tried to add also disabled and invalid use cases for the text input. It seems that we have, at least for the invalid one, a style issue. The red border is only visible when hovering the input, the issue seems inherited from v2.

Nice catch, this issue will be followed by separate ticket for v2 and v3? 
For the rest it's approved 👍

planctus:
Thanks for reviewing, @papegaill, a ticket for ECL has been created to follow up the issue discovered here: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1923

PR #1842, opened 2020-12-21
feat(select): Add controls - FRONT-1864

papegaill:


src/implementations/twig/components/select/select.story.js
```diff
@@ -1,90 +1,39 @@
...
-import dataMultiple from '@ecl/specs-component-select/demo/data--multiple';
+import dataDefaultSingle from '@ecl/specs-component-select/demo/data--single';
+import dataDefaultMultiple from '@ecl/specs-component-select/demo/data--multiple';
 
 import selectBox from './select.html.twig';
 import notes from './README.md';
 
-const prepareSelect = (data, multi) => {
-  getFormKnobs(data);
+const dataSingle = { ...dataDefaultSingle, width: 'm' };
```
planctus: Why defining two new variables? Couldn't you simply add the property to the existing ones?
But anyway, wouldn't it be better adding this in the specs instead of "adapting" the data in the story?

src/playground/addons/story-utils/index.js
```diff
@@ -539,95 +539,113 @@ export const correctSvgPath = (data) => {
...
-    label: {
-      type: { name: 'string', required: true },
-      defaultValue: data.label,
-      description: `Label of the form ${type}`,
-      table: {
-        type: { summary: 'string' },
-        defaultValue: { summary: '' },
-        category: 'Content',
-      },
+  const argTypes = {};
```
planctus, line 542: You refactored the whole function to add the condition for the width.. It's ok, eh, but you could have created the argTypes object directly instead of initialising as an empty object for the common ones:
```
const argTypes = {
 label: {},
helper_text: {}
...
};
```

and then:
```
if (data.width) {
  argTypes.width = {};
}
```
Anyway, as i said, nothing you have to necessarily change this way.

src/playground/addons/story-utils/index.js
```diff
@@ -539,95 +539,113 @@ export const correctSvgPath = (data) => {
...
+  if (data.width) {
+    argTypes.width = {
+      name: 'width',
+      type: { name: 'select' },
+      defaultValue: data.width,
+      description: `The width of the form ${type} {s: small, m: medium, l: large}`,
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: data.width },
+        category: 'Styles',
```
planctus: Maybe creating a group for one single control is not that useful, but i agree that no other of the existing groups would fit for this one, could we name it maybe "size"? And then in the text-area we would add also "rows" in the same category?

src/playground/addons/story-utils/index.js
```diff
@@ -539,95 +539,113 @@ export const correctSvgPath = (data) => {
...
+      },
+      control: {
+        type: 'select',
+        options: ['s', 'm', 'l'],
```
planctus: i see that you added a legend for this, still i feel that it would be better to define the labels in the select in an human readable way, so keeping the legend in place but having "small" "medium" and "large" as labels for the options.

planctus (commented):
It's just comments, nothing broken but a few small suggestions.

src/implementations/twig/components/select/select.story.js
papegaill: Initially I put it in the specs, but as the value is optional and has a default value (m) ... I thought it shouldn't be in the specs. But yes, it's the cleanest option indeed.

papegaill: Updated in this way 👍

src/playground/addons/story-utils/index.js
papegaill: True but I like to have all the controls defined in the same way ;)

src/playground/addons/story-utils/index.js
papegaill: I was wondering the same, but I don't see how to define select control options with a specific label and it's value.
This is why I added them in the legend.

PR #1844, opened 2020-12-21
feat(text-field): Add controls - FRONT-1866

papegaill:


src/playground/addons/story-utils/index.js
```diff
@@ -539,95 +539,113 @@ export const correctSvgPath = (data) => {
...
+      defaultValue: data.width,
+      description: `The width of the form ${type} {s: small, m: medium, l: large}`,
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: data.width },
+        category: 'Size',
+      },
+      control: {
+        type: 'select',
+        options: ['s', 'm', 'l'],
```
emeryro: options should be consistent with https://github.com/ec-europa/europa-component-library/pull/1842/files#diff-a7602161f00bf48ed6e08f933df9b586729725e3ddc921d3f84a95832f7a01c3R645

emeryro (changes requested):
small fix

emeryro (approved):

PR #1846, opened 2020-12-21
feat(description-list): Add controls - FRONT-1868

papegaill:


src/implementations/twig/components/description-list/description-list.story.js
```diff
@@ -9,66 +7,59 @@ import dataDescriptionListHorizontal from '@ecl/specs-component-description-list
...
-        tabLabels.required
-      );
-    }
-  });
-
-  getExtraKnobs(data);
+const getArgTypes = (data) => {
+  return {
+    term: {
+      name: 'first item term',
```
planctus: Think it would be better to label the control as the property, in general, and if possible. Here it could be:
`term (first item)`
`definition (first item)`

src/implementations/twig/components/description-list/description-list.story.js
```diff
@@ -9,66 +7,59 @@ import dataDescriptionListHorizontal from '@ecl/specs-component-description-list
...
+      type: { name: 'string', required: true },
+      defaultValue: data.items[0].term,
+      description: 'Test the content for the first item term',
```
planctus: We don't normally use this expressions in the description of the controls, no? I mean, not that it sounds too bad, although i guess it would be "test the content **of** the first item term" but since we have already specified what it is about i think it could simply be "The heading of the description list item" and the definition "The content of the description list item"

src/implementations/twig/components/description-list/description-list.story.js
```diff
@@ -9,66 +7,59 @@ import dataDescriptionListHorizontal from '@ecl/specs-component-description-list
...
+  },
 };
 
-export const Vertical = () =>
-  descriptionList(prepareList(dataDescriptionListDefault));
+export const Vertical = (args) =>
+  descriptionList(dataDescriptionListDefault, args);
 
 Vertical.storyName = 'vertical';
+Vertical.argTypes = getArgTypes(prepareData(dataDescriptionListDefault));
```
planctus: This is confusing me,  normally we get the` argTypes` from `getArgTypes()` and we apply the data we prepared in `prepareData()` where we pass the` args` "collected" in the rendering function. This way we are not passing `args` to `prepareData()` at all and then it's simply an `undefined` what we are merging with data.

src/implementations/twig/components/description-list/description-list.story.js
```diff
@@ -9,66 +7,59 @@ import dataDescriptionListHorizontal from '@ecl/specs-component-description-list
...
+        defaultValue: { summary: '' },
+        category: 'Content',
+      },
+    },
+  };
+};
 
-  return data;
+const prepareData = (data, args) => {
+  return Object.assign(data, args);
```
planctus: In this case we cannot use `Object.assign()` like this, what it does is to "merge" the two objects, but the properties of  `args` are not the ones we want to apply to data, here we have to assign them to `data.items[0]`.
So, like:
```
  data.items[0].term = args.term;
  data.items[0].definition = args.definition;
  return data;
};
```

planctus (changes requested):

src/implementations/twig/components/description-list/description-list.story.js
papegaill: Oops, thank you, yes this was a mistake on my part 👍

emeryro (changes requested):
there is an issue with the description list: the definition of the first item is not displayed
image
I guess that it is linked to the data passed to it

papegaill:
> there is an issue with the description list: the definition of the first item is not displayed

Thank you @emeryro, fixed 👍

emeryro (approved):

PR #1849, opened 2020-12-21
fix(forms): update deprecated css - FRONT-1899

emeryro:
adapt form css according to recommandations made in the migration documentation

### DIFF
--- a/docs/Migrating-v3.md
+++ b/docs/Migrating-v3.md
@@ -50,6 +50,7 @@ The following packages have been re-organized:
 - `@ecl/polyfills` has been moved to `@ecl/dom-utils/polyfills`
 - `@ecl/(ec|eu)-auto-init` have been moved to `@ecl/dom-utils/autoinit`
 - `@ecl/(ec|eu)-base/helpers` have been moved to `@ecl/dom-utils`
+- all generic form packages (feedback-message, form-group, help-block, form-label) have been merged into `@ecl/vanilla-component-form`
 
 Example from v2:
 
--- a/docs/Migrating-v3.md
+++ b/docs/Migrating-v3.md
@@ -47,10 +47,10 @@ The following packages have been re-organized:
 - `@ecl/ec-utility-font-size` has been removed
 - `@ecl/ec-utility-ratio` has been removed
 - `@ecl/ec-utility-text` has been removed
+- `@ecl/(ec|eu)-component-form-form-group`, `@ecl/(ec|eu)-component-form-feedback-message`, `@ecl/(ec|eu)-component-form-help-block` and `@ecl/(ec|eu)-component-form-form-label` have been consolidated in `@ecl/vanilla-component-form`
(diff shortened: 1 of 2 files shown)


src/implementations/vanilla/components/form/form.scss
```diff
@@ -22,37 +22,22 @@
   margin: 0;
   padding: 0;
 
-  > *:not(.ecl-form-label--hidden) + * {
+  > *:not(.ecl-u-sr-only) + * {
     margin-top: map.get(theme.$spacing, 'xs');
   }
 
-  > *:not(.ecl-form-label--hidden) + .ecl-checkbox {
+  > *:not(.ecl-u-sr-only) + .ecl-checkbox,
```
planctus, line 29: I think i don't understand the use case we are dealing with here...where do we have an `ecl-u-sr-only` element in the definition of our forms? I'd say that if we don't use utilities in components then we shouldn't even use them in their style, but i'm sure i'm missing something here... :)

docs/Migrating-v3.md
```diff
@@ -50,6 +50,7 @@ The following packages have been re-organized:
 - `@ecl/polyfills` has been moved to `@ecl/dom-utils/polyfills`
 - `@ecl/(ec|eu)-auto-init` have been moved to `@ecl/dom-utils/autoinit`
 - `@ecl/(ec|eu)-base/helpers` have been moved to `@ecl/dom-utils`
+- all generic form packages (feedback-message, form-group, help-block, form-label) have been merged into `@ecl/vanilla-component-form`
```
kalinchernev: Thanks for adding this note in the document, it's useful to communicate

Please move the note under `@ecl/ec-utility-text has been removed` or some other line as otherwise adding it here breaks the logical flow of the examples for the `@ecl/dom-utils` which follow under it.

Not a blocker, but good to have, replace "all generic form packages" with their corresponding items to follow the pattern in the list

`@ecl/(ec|eu)-component-form-form-group`, `@ecl/(ec|eu)-component-form-feedback-message`, `@ecl/(ec|eu)-component-form-help-block` and `@ecl/(ec|eu)-component-form-form-label` have been consolidated in `@ecl/vanilla-component-form`

kalinchernev (changes requested):
Only one small docs update request

src/implementations/vanilla/components/form/form.scss
kalinchernev: I tracked the original rules setting to https://github.com/ec-europa/europa-component-library/pull/1035
It seems to me there are specific rules to follow which we better keep in order to avoid breaking some spacing in edge cases.

kalinchernev: Regarding the usage of utility classes within components' styling: seems that the linter is ok and there are no "weak" rules although the mixed specificity. I don't know why but i think it's fine for this case

kalinchernev (approved):

PR #1854, opened 2020-12-22
feat(controls): Fact and figures, replacing knobs with controls, adding getOptions() utility - FRONT-1872

planctus:


src/playground/addons/story-utils/index.js
```diff
@@ -649,3 +649,10 @@ export const getFormControls = (data, type) => {
 
   return argTypes;
 };
+
+export const getOptions = (data) => {
```
kalinchernev: The name of the utility is not describing it's purpose quite exactly. The function converts an array of strings to an object of key value pairs containing the the same strings. I'm not sure whether other cases will interest of the same conversion on the level of making an options list of just any contol.

As suggested in https://github.com/ec-europa/europa-component-library/pull/1855/files#r547903798 we can import  uiIcons directly in this story-utils file, prepare the list of icons and use them in a contol-related utility. If relevant the list of icons can also be exported, but i doubt it will be useful

This file 
```javascript
import uiIcons from '@ecl/resources-ec-icons/dist/lists/ui.json';

export const iconList = uiIcons.reduce((a, b) => {
    a[b] = b;
    return a;
  }, {});
  
  export const getIconControls = (data) => {
//  return the contols structure
  }
```

Stories will then be able to pick whichever way `import { iconList } from '@ecl/story-utils';` for the ready list or `import { getIconControls } from '@ecl/story-utils';` for the ready control.

src/implementations/twig/components/fact-figures/fact-figures.story.js
```diff
@@ -1,135 +1,117 @@
...
+        type: { summary: 'object' },
+        defaultValue: { summary: '{}' },
+        category: 'Others',
+      },
+    },
+    displayIcons: {
+      name: 'icons visibility',
+      type: { name: 'boolean' },
+      defaultValue: true,
+      description: 'Toggle visibility of the icons',
```
kalinchernev, line 74: Checking https://v3--europa-component-library.netlify.app/playground/ec/?path=/story/components-fact-figures--columns-3 the toggle is related to the descriptions, not the icons. Not an issue obviously, just a difference without notes in the pull request

kalinchernev: Another difference: no control for view_all.link.label

kalinchernev (commented):
A few comments below, though in it's good to be merged

src/implementations/twig/components/fact-figures/fact-figures.story.js
planctus: Yeah, but we are following the ECL v2 knobs, not the ecl-twig ones although i actually added some here, to fully experience an item customisation (only one). 
The description can be reset by simply removing the text inside the control, so that's why i haven't been adding a "toggle" for it, the ones i placed are the ones from V2, basically.
Since there is the toggle i believe is not worth to place a control for the label of the link, otherwise i would be better to remove the toggle and delete the object when the control is reset, but i would keep it as it is.

image

src/playground/addons/story-utils/index.js
planctus: Well, the idea was to provide a generic utility for creating the right object to pass to a select control, but the reality is that this is not needed, this was a legacy from the knobs, in controls i see we can pass directly the object as we import it, so i'm simply removing the utility and just using the imported list as it is.
Good if we have a utility for quickly importing icons knob but it would not be used i think in this case, here we just want to show that the icon can be chosen, not worth i think to have all the related controls.

PR #1855, opened 2020-12-23
feat(icon): add controls - FRONT-1880

emeryro:


src/playground/addons/story-utils/index.js
```diff
@@ -538,6 +538,88 @@ export const correctSvgPath = (data) => {
   return data;
 };
 
+export const getIconControls = (data, icons) => {
```
kalinchernev, line 541: I was reading https://github.com/ec-europa/europa-component-library/pull/1854/files#diff-a7602161f00bf48ed6e08f933df9b586729725e3ddc921d3f84a95832f7a01c3R653 and briefly checked this pull request to have an idea how to have alignment in the two pull requests.

Do you think it's worth removing the icons parameter and use this property internally in the utility?

I mean, in https://github.com/ec-europa/europa-component-library/pull/1854/files# we have a utility which prepares a list of icons and this pull request returns controls for the icon list: can we make it here in 1 place and avoid stories caring about the list of icons at all?

kalinchernev: Placing a reference to the other pull request https://github.com/ec-europa/europa-component-library/pull/1854/files#r547909364

papegaill (approved):

src/playground/addons/story-utils/index.js
emeryro: As the utility has been removed from the other PR, I guess that we can merge this one

PR #1862, opened 2021-01-04
feat(language-list): Remove knobs - FRONT-1882

papegaill:


planctus (commented):
I understand is not in the scope of this ticket, but why don't we take the chance to add the EU version of the component? Up to you, otherwise we can create a follow up ticket..

src/implementations/vanilla/components/language-list/_language-list.scss
```diff
@@ -8,16 +8,16 @@
...
 $_background-link-active: map.get(theme.$color, 'yellow-100');
 $_color-link-active: map.get(theme.$color, 'black-100');
-$_color-title: map.get(theme.$color, 'white-100');
+$_color-title: null !default;
 $_logo-height-desktop: 5.5rem;
-$_logo-width-desktop: 11.25rem;
+$_logo-width-desktop: null !default;
 $_logo-height-mobile: 3.75rem;
-$_logo-width-mobile: 7.5rem;
+$_logo-width-mobile: null !default;
```
emeryro, line 21: the color of the close button is not correct on EU
image
I guess it will require an additional variable

emeryro (changes requested):

src/implementations/vanilla/components/language-list/_language-list.scss
papegaill: Thanks for spotting this, Updated using existing `$_color-title` variable.

planctus:
What happened in this pr @papegaill..? I see we are reverting all the changes in the breadcrumbs for the EC/EU different specs, don't think it is on purpose, also the tests are failing because i guess there is some misalignment still with the breadcrumb specs as well as the preview generation, guess some merge mixed up things.

PR #1863, opened 2021-01-04
feat(media-container): add controls - FRONT-1883

emeryro:


src/implementations/twig/components/media-container/media-container.story.js
```diff
@@ -1,71 +1,115 @@
...
+import dataVideo from '@ecl/specs-component-media-container/demo/data--video';
+import dataEmbed from '@ecl/specs-component-media-container/demo/data--embed-video';
 import mediaContainer from './media-container.html.twig';
 import notes from './README.md';
 
-const prepareMediaContainer = (data, media) => {
+const getArgTypes = (data, media) => {
+  const argTypes = {};
+
   if (media === 'video') {
```
papegaill: I don't think you need to create a new `media` param here. 
Is this condition not enough ?

```
if (data.embedded_media) {
...
} else if (data.sources) {
...
}
```

papegaill (changes requested):

src/implementations/twig/components/media-container/media-container.story.js
emeryro: I kept the same logic as in the original file, but I'll check if it can be simplified

PR #1865, opened 2021-01-04
feat(page-header-core) Add controls - FRONT-1885

papegaill:


src/implementations/twig/components/page-header-core/page-header-core.story.js
```diff
@@ -13,98 +11,106 @@ import dataBreadcrumbLong from '@ecl/specs-component-breadcrumb/demo/data';
...
 
-const preparePageHeaderCore = (data, desc, meta, img) => {
-  data.breadcrumb = dataBreadcrumbLong;
-  data.title = text('title', data.title, tabLabels.required);
-  if (meta) {
-    data.meta = text('meta', data.meta, tabLabels.optional);
+const getArgTypes = (data) => {
+  const argTypes = {};
+  argTypes.title = {
+    name: 'title',
```
planctus: This is not needed, if your argType key is "title" the control will show "title" even without specifying a "name" for it

src/implementations/twig/components/page-header-core/page-header-core.story.js
```diff
@@ -13,98 +11,106 @@ import dataBreadcrumbLong from '@ecl/specs-component-breadcrumb/demo/data';
...
+    description: 'The page title',
+    table: {
+      type: { summary: 'string' },
+      defaultValue: { summary: '' },
+      category: 'Content',
+    },
+  };
+  if (data.description) {
+    argTypes.description = {
+      name: 'description',
```
planctus: Same as before, up to you to remove them, no harm if they stay..

src/implementations/twig/components/page-header-core/page-header-core.story.js
```diff
@@ -13,98 +11,106 @@ import dataBreadcrumbLong from '@ecl/specs-component-breadcrumb/demo/data';
...
   }
-  if (desc) {
-    data.description = text(
-      'description',
-      data.description,
-      tabLabels.optional
-    );
+  if (data.meta) {
+    argTypes.meta = {
+      name: 'meta',
```
planctus: Same...

src/implementations/twig/components/page-header-core/page-header-core.story.js
```diff
@@ -13,98 +11,106 @@ import dataBreadcrumbLong from '@ecl/specs-component-breadcrumb/demo/data';
...
-    );
-  });
-  getExtraKnobs(data);
 
-  return data;
+  return argTypes;
+};
+
+const prepareData = (data, args) => {
+  data.breadcrumb = dataBreadcrumbLong;
```
planctus: I see that we did not do when we worked on the breadcrumbs, but we have now the EU storybook and in theory we have, or better had, different specs for EU (European union instead of European commission, basically), not a big deal but it looks like we have the time and the chance to do this now. 
Up to you to do this in the context of this pr, otherwise we can create a separate ticket for it, or we can simply ignore this.. :)

planctus (commented):
To me we can merge it as it is, but if you feel to follow the suggestions in the comments, i'll wait for your feedback.. ;)

src/implementations/twig/components/page-header-core/page-header-core.story.js
papegaill: Updated all argTypes for the 3 page headers ;)

planctus (approved):

PR #1875, opened 2021-01-05
chore: visual regression testing - FRONT-1686

kalinchernev:
Inspiration taken from saucelabs testrunner toolkit as mentioned already in FRONT-1686. Docker becomes a new required dependency for generating reference screenshots.

ECL [v1 uses saucelabs's cloud compute to generate reference screenshots](https://github.com/ec-europa/europa-component-library/blob/v1/docs/testing/visual.md) based on machines which are not dev's host machines. In the testrunner and the suggested approach here, it's docker the layer independent from host machine's architecture. 

Scripts are thin wrappers on top of docker deamon tasks. They work locally and in github actions without rework for CI environment specifics.

Main advantages compared to https://github.com/ec-europa/europa-component-library/pull/814
- Simpler setup, no cloud and no tunnels
- Stories with parameters receive regression testing automatically
- Only google chrome, faster and more focused testing

Implementation notes:
- Tests have been split into sub-categories because of the way storyshots work. It's a single test file with single setup rather than multiple files with multiple assertions like in the snapshot testing. The `jest --verbose` command was running this single file in a wrong context
- There's a new `gen:refs-ec` task for (re)generating reference screenshots and `test:visual-regressions-ec` for making the comparison of current state vs references. These need to be done in a docker environment only on statically built version of storybook. Generating references on served (storybook-start) storybook will not generate consistent results.
- `@storybook/html` needs to be added as a dependency in the root, unfortunately. Under the hood, storyshots puppeteer uses jest utilities to load the right storybook flavor (html / react / etc.) and it's not node's path resolution algorithm. So when the html flavor is a symlink in a workspace (if not in root but from child deps) the actual physical file is not resolved correctly.
- `STORYBOOK_DIR=$(pwd)` is made so because docker configurations won't accept `./build` relative pattern with a special character directly. That's why the script converts the current working directory to an absolute before running the deamon task.
- Having puppeteer in the stack allows for adding [automated accessibility checks with axe](https://www.npmjs.com/package/@storybook/addon-storyshots-puppeteer#user-content-axetest) in addition to the image screenshot comparision. It's out of scope but good to know


kalinchernev:
Dropping an alternative I found browsing the web today https://github.com/wKich/creevey

PR #1878, opened 2021-01-06
feat(button): add EU button - FRONT-1931

emeryro:
specs: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/b30247d2-5e18-48dc-b3db-c6531941245f/collections/ca4f53e5-29b6-404e-931f-0928c67d3fa8

Notes:
- variant "search" has been removed (will be part of the search form). This has been agreed with designers.
- scss for button has been split into different files (common, ec values, eu values). The same has been done for search form (we don't have the final specs for search form, so it may be modified later)

Design has been validated by COMM (only the icons will have to be updated, but that will be handled in another PR)


planctus:
Looks perfect, although the icon size seems to big in my personal view, but that follows the specs. I've just noticed something not related to this implementation, meaning that also the EC version is affected in the same way: since we use flex in the content of the button we should prevent the icon to reduce its size as the text can wrap and use multiple lines:
image
It would be enough to add `flex-shrink: 0;` to the` .ecl-button__icon` element.
image

planctus (approved):

emeryro:
good catch @planctus. I updated the CSS accordingly

PR #1879, opened 2021-01-06
feat(site-header-harmonised): Adding Eu version, replacing knobs with controls - FRONT-1889

planctus:


emeryro (approved):
This harmonised site header is quite messy currently, and will require some deep rewriting when we have the validated specs (avoid duplicating code in stories and css, align story names, cleanup data, ...)
For now I'd say it is ok-ish

PR #1882, opened 2021-01-07
feat(timeline): Add Controls - FRONT-1895

papegaill:


src/implementations/twig/components/timeline/timeline.story.js
```diff
@@ -1,64 +1,75 @@
...
-    item.title = text(`items[${i}].title`, title, tabLabels.optional);
-    item.content = he.decode(
-      text(`items[${i}].content`, content, tabLabels.required)
-    );
-  });
+  data.toggle_collapsed = `Show ${hiddenCount} more items`;
+  data.toggle_expanded = `Hide ${hiddenCount} items`;
 
-  getExtraKnobs(data);
+  data.icon_path = '/icon.svg';
```
planctus: Wouldn't it be better to put this in the specs? Also because otherwise you could just set the right value in the story instead of the "placeholder" for then replacing it via the helper function.

src/implementations/twig/components/timeline/timeline.story.js
```diff
@@ -1,64 +1,75 @@
...
+  parameters: {
+    knobs: {
+      disable: true,
+    },
+  },
 };
 
-export const Default = () => timeline(prepareTimeline(demoData));
+export const Default = (arg) => {
+  const timelineHtml = timeline(prepareData(demoData));
```
planctus: It's fine eh, take this only as a suggestion, i personally don't like this string concatenation that we return by this function that forces you to return an empty string from the prepareHtmlContent when the boolean is false.
You could prepare the story inside the aforementioned function, instead, like this:
```
const prepareHtmlContent = (state) => {
  let story = timeline(prepareData(demoData));
  if (state) {
    story += `<p class="ecl-u-type-paragraph-m">${loremIpsum({ count: 25 })}</p>
            <p class="ecl-u-type-paragraph-m">${loremIpsum({ count: 25 })}</p>`;
  }

  return story;
};
```
and then you would export Default the normal way:

`export const Default = (arg) => prepareHtmlContent(arg.showDummyContent);`

Just a suggestion.
Another suggestion is: ES6 introduced a method for repeating strings, which is, predictably, `string.repeat(times)`
you could then add the "dummy content" like this:
```
    const dummy = `<p class="ecl-u-type-paragraph-m">${loremIpsum({ count: 25 })}</p>`;
    story += dummy.repeat(2);
```
but to be honest, do we really need two paragraphs..? :)

planctus (commented):
It works fine, just a couple of suggestions.. :)

src/implementations/twig/components/timeline/timeline.story.js
papegaill: Nice, Thks for suggestion 👍

planctus:
Thanks for the changes, we need to update the snapshot, though, @papegaill ;)

PR #1883, opened 2021-01-07
feat(netlify-cms): Adding configuration - FRONT-1627

planctus:
The admin interface is reachable at https://ecl-preview-1883--europa-component-library.netlify.app/admin, but it's not accessible by any user that is not "invited" as the authentication uses netlify "identity". 
This means that if you want to take a look i can add you to the list of authorized users (maximum 5), you would receive an invitation by email and then you would be able to login into the cms by setting a password.
All the relevant pages has been added in the configuration file of the cms and the files extension has been changed to simply `.md` instead of `.mdx`.
The plan is to have this merged in next-v3, but this is not really going to make much of a difference, just that the url for accessing the admin interface would be more relevant, but still our main branch is v2 so we would still have problems with the links in the email sent by netlify, they refer indeed to the main url of the website linked to the ecl configuration in netlify and these links need to be manually edited to point to the right preview url.

image


emeryro (changes requested):
If I understood well, you prepared the files to be editable in the CMS, and did an example for the accordion, which is good.
I have a few remarks/questions:

1. I think you mentioned it before, but do you think we could have the same order for the tabs as before (and have the page named "Usage" instead of "usage")?
image

2. Can we use a larger font for the first paragraph, to match what we have currently? I know that this is just an example, but let's make it as close to the expected result as possible
image

3. The anatomy image is now set into a paragraph, which reduces its width. Can we use a simple div here instead? If not, we may add some css to display it full width
image

4. I think that it would make sense to have a shared account for FRONT team to access admin.

planctus:
1) ah, now i see what that comment block on top of the files is for...at the moment this is an issue, because that block is not seen by the editor, when it saves the files it adds an empty line, if i remember correctly, and since we run prettier only on the code that is coming from the cms editor, the empty line is not removed and makes the prettier check fail...but i see that we need to find a solution, removing the comment on top of the usage page files has consequences..ok

2) Not as a paragraph (unless we use like a pseudo selector like first-of-type in css), but if they want to keep the bigger font size with the rich text editor it could be set as an h something...

3) I see, i think indeed css might be the way to go, i presume the paragraph is the default element wrapper for all the widgets provided by the editor and maybe we don't want to change that, overriding this behaviour only for the image i'm not sure it's going to be possible, but we can give it a try..

4) I know that we have an email for the team, right..? If we can access it i can add that in the list of authorized accounts in netlify ;)

planctus:
Ok..well, this is what i did:

1. since the usage pages will be handled via the cms and prettier will run every time one of these files will be saved, i excluded them from the `pretty-check`, there was no way of leaving that comment on top of the files to preserve the order of the tabs without breaking prettier...

2. I noticed that the "legend" of the anatomy images is not part of the images only after "fixing" via css the max-width for the containing paragraph, so i've also "reproduced" the style of that element of the anatomy react component in the `markdown.scss`.

3. I've also fixed the style of the first paragraph, formerly part of the react component, in the same scss file, these fixes are a little bit hacky, considering that their selectors are relatively generic, especially the one to find the "anatomy" image, let's check we don't break anything this way...

4. I added an npm package to make the cms work also in the local environments, instructions have been added in the `README.md` file in the `admin` folder in `src/website`, so now we can all access it locally.

5. I've also changed the branch it points to, including the relative preview, so as long as we don't merge this the cms will not be accessible "online" from the netlify preview url.

planctus:
Not removing the previous comment i left, but i could notice myself that, indeed, the css rules were breaking everything...
Now, there are not really ways to get a nicely formatted html page based on an md file, the way i could do it is maybe a bit more than hacky...basically if you check the accordion usage page there is now an empty blockquote before the "anatomy" image, it's harmless and it give us the chance to reproduce the existing styles but there is an additional issue: this empty blockquote gets removed by the cms when you reopen the file for editing...
I know, it's nothing good, but that's it, an alternative could be to embed the legend and the border in the anatomy images, and export them "full width", maybe it's better that way and considering that most of these anatomy images will not be part of the website anymore (only the ones in the families templates), also the effort by the "designers" should be minimal...
For the first paragraph, now it's bigger in all the pages, even when it was not, but i left it because it might make sense like this.. :)

emeryro (approved):
For a first version, it works well (especially the local editing)
I guess that we will have soon to find a better solution for the hacky parts (first paragraph and full width images), but this could probably be achieved with some custom plugin like the one you did for table

PR #1889, opened 2021-01-12
feat(presets): add reset preset - FRONT-1911

emeryro:
- remove preset "ec-preset-website"
- add new "reset" preset
- cleanup content of release packages
- update doc

This is the structure of ECL packages provided to end users:
image


planctus:
I see that the result is exactly what we talked about, but at the same time i realise that what we call here "reset" is basically just an external library which doesn't even really reset in its basic approach and i wonder: why do we ship this with ECL, even only as an optional css, wouldn't it be better to "recommend" it in case the ECL user needs it, as an external resource that the can be added?
I also see that we currently put some global css in this reset and maybe that should be instead part of the global.css we have in mind. At that point i see how we could use the approach in this pull request for defining this "optional" global.css and just mention normalize somewhere as a recommended approach to harmonise different browsers defaults when implementing ECL in a website, without bundling it in the library presets, maybe.

emeryro:
@planctus thanks for your feedback.
Concerning the reset, we need to have it somewhere, to be able to use it ourselves in storybook and on the website. So why not shipping it with ECL? At least we can ensure that the display is correct with this reset, and make it easier for ECL users.
Now concerning the globals, I agree that it is kind of tricky. I won't recommand putting it in the same place as html tag styling, as it doesn't serve the same purpose and doesn't target the same audience. We could put these (2) rules in a dedicated css, but that would be overkill to me. I suggest that we move forward as is, and we will handle next step when dealing with html tag styling.

planctus:
Ok, i'm not in favour of masquerading an external library into a resource shipped as a resource in the library, i would prefer naming it "normalize", maybe, but then it would be clearly non sense working it internally as a preset, but fine, if you think it's ok, no reason to block this.
About the two properties applied to the body, i see them more appropriately placed in a global.css than in a "reset" one, probably, not sure about the different targets you mentioned, but if that means that they are needed for ECL to work properly, for instance, is not an optional css the right place for them.
Let's go like this, we will maybe come back to this at some point.

PR #1899, opened 2021-01-19
feat(icons): Add eu icons - FRONT-1934

emeryro:
- integrate EU icons. Icons name are the same between EC and EU. Also social media icons haven't been modified yet, as I don't know if they will be different for EU
- remove icon categories. Now all icons are using their name directly
- update components and utilities to use EC or EU icons depending on the needs
- cleanup package dependencies

Note:
- the way to handle icons in client-facing js (select and table) should be improved; currently it is only relying on URL. Ideally the icon should be assigned when creating the compressed js file. This will be handled in a follow-up task


planctus (commented):
It's a big pr with nothing changing at the moment, can we maybe wait for these icons to come? I don't think there will be many other prs merged and potential conflicts to be solved if we do it.

planctus (approved):

PR #1901, opened 2021-01-20
feat(footer-standardised): Refactor markup - FRONT-494

papegaill:


planctus (commented):
I saw the label "review needed" and so i started looking at this POC, at the moment you have an error preventing the EC demo to work, but that was easy to spot.
The result in terms of markup is good, i find it much clearer and meaningful, we have now rows, columns and sections, reflecting the structure of this complex component.
I see maybe some minor css difference about the spacing with the current version in mobile, but in general i was expecting a different approach, in terms of the css approach (i've only checked EC for now).
What i mean is that we still have a grid css for each row, while in the codepen you created you were using a flex approach, if i'm not wrong, and i think it would also simplify the css together with the markup.
What really surprises me, maybe, is the implementation in the twig template, using macros.
I don't see why going this way, honestly, it is very difficult to read and to my understanding is not really providing any particular advantage, macros are meant to avoid code repetition, here we are using it to basically create a certain wrapper which is still based on the level in a loop, so why not doing it while we are looping.
Basically we have now rows, columns and sections, i would also recommend renaming the "old" sections in rows, starting from the specs, so basically this 3 levels loop would do this:

```
<footer class="ecl-footer-standardised">
  <div class="ecl-container ecl-footer-standardised__container">
  {% for _row in rows %}
   <div class="ecl-footer-standardised__row">
   {% for _columns in row %}
      <div class="ecl-footer-standardised__column">
      {% for _sections in columns %}
        <div class="ecl-footer-standardised__section"></div>
        {% if this %}
        {% endif %}
        {% if that %}
        {% endif %}
        </div>
      {% endfor %}
      </div>
    {% endfor %}
    </div>
  {% endfor %}
  </div>
</footer>
```

Unless i'm missing something this looks enough to render the structure of these "specs" and it seems more readable than the macro in this pr, but yeah, tell me if i'm wrong or if you find the logic you used clearer or smarter.

Mind the fact that both the `demo_id` and the `type` can be removed from the specs, those are used in ecl-twig to handle the complex operation in the demo to hide sections and switch columns when needed, something we don't do in our demo in ECL, so they can just be removed.
I also think that the "special cases" like the inline list of the social media should not be handled via an `extra_class` passed to the component, that was a way to adapt the ECL specs in ecl-twig, but i feel that it could just be a key like "social" or "inline" ( if there are other use cases) and then we construct the css class in the template.

Thanks for this work, i'd love to see this markup replacing the current one and i think we are on the good track, the things we can improve, in my opinion, just to recap, are these:

1.  changing sections in rows in the specs and removing outdated keys
2.  Use a simple loop structure in the template
3.  Using flex instead of a css grid for the rows
4.  We have to find a reasonable way to fix those css issues about the spacing in mobile

image

papegaill:
@planctus Many thanks for this great review and suggestions 👍 I just updated the PR with the 2 first point.
Regarding point 3: "Using flex instead of a css grid for the rows", I first tried using flex like my CodePen POC. I preferred to keep the existing grid to have exactly the same margins(gutter) between the columns. I think it's easier, because otherwise you have to apply different margins for each column. First column must not have a margin-left the last one must not have a margin-right..etc ... For point 4, I haven't gone any further yet, in order to have the simplest version possible at markup level ... but yes there are differences in spacing on some sections ... it's maybe best to discuss it directly together..

src/specs/components/footer-standardised/demo/data--ec.js
```diff
@@ -1,239 +1,238 @@
...
-      {
-        title: 'About us',
-        links: [
-          {
-            link: {
-              label: 'Information about the DG',
-              path: '/example',
-              aria_label: 'Link to Information about the DG',
+          ],
+          list_class_name: 'ecl-footer-standardised__list--inline',
```
planctus: same as for the "separator" use case, maybe it could be `with_inline_list`, but to be honest, this is only for the social media block, so why not referring to that, instead of a "style" of the list, it's easier for the user to identify this section as the one for the social media, it's harder if he has to realise that the list containing it has to be shown inline, so why not: `links_to_social_networks`?

src/specs/components/footer-standardised/demo/data--ec.js
```diff
@@ -1,239 +1,238 @@
...
+                aria_label: 'Link to Contact information of the DG',
+              },
             },
-            icon: {
-              path: '/icons.svg',
-              type: 'branded',
-              name: 'twitter',
-              size: 'xs',
+          ],
+          title_class_name: 'ecl-footer-standardised__title--separator',
```
planctus: In my opinion this can be a boolean, key: `with_separator`
We can build the class ourselves in the template without needing the users to pass a long class name

src/specs/components/footer-standardised/demo/data--ec.js
```diff
@@ -1,239 +1,238 @@
...
+            },
+            {
+              link: {
+                label: 'Class name 2',
+                path: '/example',
+                aria_label: 'Link to Class name 2',
+              },
+            },
+          ],
+          list_class_name: 'ecl-footer-standardised__list--condensed',
```
planctus: Also in this case, i would not rely on the understanding of the style of the element, maybe on its position, like `middle_row`, considering that this is the only section in the middle row, when it is set. In any case the class name is a construct that we can do ourselves, no need to ask the users for the whole string.

src/specs/components/footer-standardised/demo/data--ec.js
```diff
@@ -1,239 +1,238 @@
...
+          content_before: 'More information on:',
+          section_class_name: 'ecl-footer-standardised__section--small-spacing',
```
planctus: if we know that this is the `middle_row`, we can also set this specific class. We could even deduce this, by counting the rows, and if we have three both these classes could be applied without any "hint" from the data...

src/specs/components/footer-standardised/demo/data--ec.js
```diff
@@ -1,239 +1,238 @@
...
+              },
+            },
+            {
+              link: {
+                label: 'Resources for partners',
+                path: '/example',
+              },
+            },
+          ],
+          section_class_name: 'ecl-footer-standardised__section--split-list',
```
planctus, line 197: Same as the other comments related to these class names, not sure about the class name itself, it is kind of doing the opposite, like "joining" more than splitting, maybe, but here we might simply be able to select this section without any identifier, if we look for the last row, second column: ` .ecl-footer-standardised__row:last-child .ecl-footer-standardised__column:nth-child(2) .ecl-footer-standardised__section`

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,140 @@
...
+  argTypes.hide_contact = {
+    name: system === 'EU' ? 'contact site name' : 'contact us',
+    type: { name: 'boolean' },
+    defaultValue: false,
+    description:
+      system === 'EU'
+        ? 'Hide "Contact site name" section'
+        : 'Hide "Contact us" section',
+    table: {
+      type: { summary: 'boolean' },
```
planctus: mmh..ok that we define these controls as boolean to have a checkbox, but there is no relation with the data in this case, i would simply leave the table "empty", no..? No default value, no "type"

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,140 @@
...
+      defaultValue: { summary: false },
+      category: 'States',
```
planctus: I would reuse the label we have in twig, "use cases"..we use states for the form elements, like, anyway something different, here we are demoing the different content that can feed the component.

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,140 @@
...
+    },
+  };
+
+  if (system !== 'EU') {
+    argTypes.hide_class_name = {
+      name: 'class name',
+      type: { name: 'boolean' },
+      defaultValue: false,
+      description: 'hide "Class name" section',
```
planctus: very small thing, but this lowercased while the other descriptions are capitalized

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,140 @@
...
+    table: {
+      type: { summary: 'boolean' },
+      defaultValue: { summary: false },
+      category: 'States',
+    },
+  };
+
+  if (system !== 'EU') {
+    argTypes.hide_class_name = {
+      name: 'class name',
```
planctus: Maybe "class links", it is not clear what class name refers to, there can be multiple.

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,140 @@
...
   title: 'Components/Footers/Standardised',
-  decorators: [withCode, withNotes, withKnobs],
+  decorators: [withCode, withNotes],
+  parameters: {
+    knobs: { disable: true },
+  },
 };
 
-const data = process.env.STORYBOOK_SYSTEM === 'EU' ? specsEu : specsEc;
+const system = process.env.STORYBOOK_SYSTEM;
```
planctus: Why not using getSystem()..?

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,140 @@
...
-  });
+const getArgTypes = (system) => {
+  const argTypes = {};
+  argTypes.hide_contact = {
+    name: system === 'EU' ? 'contact site name' : 'contact us',
+    type: { name: 'boolean' },
+    defaultValue: false,
+    description:
+      system === 'EU'
+        ? 'Hide "Contact site name" section'
```
planctus: An why not doing the opposite, i mean, we set them as true meaning that we show them, if you hide them they are false, seems more straightforward like that..no?

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,140 @@
...
+const prepareData = (data, args, system) => {
+  correctSvgPath(data);
+
+  if (data.rows[1][0][0].logo) {
+    data.rows[1][0][0].logo.src_mobile = logoEuMobile;
+    data.rows[1][0][0].logo.src_desktop = logoEuDesktop;
+  }
+
+  const res = JSON.parse(JSON.stringify(data));
+  if (args.hide_contact === true) {
```
planctus: I love the solution overall, but why checking for `=== true`?
Wouldn't be enough to say `if (args.hide_contact)`..?

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,140 @@
...
+    knobs: { disable: true },
+  },
 };
 
-const data = process.env.STORYBOOK_SYSTEM === 'EU' ? specsEu : specsEc;
+const system = process.env.STORYBOOK_SYSTEM;
+const data = system === 'EU' ? specsEu : specsEc;
 
-export const Default = () => footer(formatFooter(data));
+export const Default = (args) => footer(prepareData(data, args, system));
```
planctus: Why passing system to prepareData? system is a constant that you define in an higher scope, prepareData can use it, just to make it more readable and to avoid a linter complaint, put in on the top of the file, as we did in other stories.

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,140 @@
 /* eslint-disable no-undef */
```
planctus: You can remove this

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -27,7 +26,9 @@
 
 {# Internal properties #}
 
-{% set _sections = sections|default([]) %}
+{% import _self as footer %}
```
planctus: is this a leftover..? :)

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -27,7 +26,9 @@
...
+
+{% set _rows = rows|default([]) %}
 {% set _css_class = 'ecl-footer-standardised' %}
 {% set _extra_attributes = '' %}
 {% set _internal_wrapper = false %}
```
planctus: it seems that also this can be removed

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +50,94 @@
...
+          {% if _sections.logo is defined and _sections.logo is not empty %}
+            {% set _label %}
+              <img
+                alt="{{ _sections.logo.alt }}"
+                title="{{ _sections.logo.title }}"
+                class="ecl-footer-standardised__logo-image-mobile"
+                src="{{ _sections.logo.src_mobile }}"
+              />
+              <img
+                alt="{{ _sections.logo.alt }}"
```
planctus, line 75: Well, we never did it before, but better start, we will have to do it massively on all the templates, this is an attribute, let's escape it properly with `|e('html_attr')`

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +50,94 @@
...
+                title="{{ _sections.logo.title }}"
```
planctus, line 77: On all the attributes where are passing a variable... :)

planctus (changes requested):
Fantastic work, i would have loved to have also the css switching to flex, but this is great, still a few suggestions to improve the data we really need to build the component and some other things here an there, but thanks a lot, this components might make sense, now.. :)

src/specs/components/footer-standardised/demo/data--ec.js
papegaill: This is questionable, I tried at the beginning to not use custom classes as possible on the sections by using nth-child(). the problem I see is if a row or columns is removed there could be an impact on the behavior of the grid which is not wanted… by using a specific class, the user will have the control?

src/specs/components/footer-standardised/demo/data--ec.js
papegaill: Param `title_with_separator` added 👍

src/specs/components/footer-standardised/demo/data--ec.js
papegaill: To me `links_to_social_networks` looks very confusing, instead I added the param `links_inline` as a boolean

src/specs/components/footer-standardised/demo/data--ec.js
papegaill: I have removed this custom class for the list as suggested 👍  
I'm now only using the section class that I have renamed from
`ecl-footer-standardised__section--small-spacing` 
to 
`ecl-footer-standardised__section--condensed`

src/specs/components/footer-standardised/demo/data--ec.js
papegaill:
```
{% if rows|length == 3 and row_index == 2 %}
  {% set _section_class_name = _section_class_name ~ ' ecl-footer-standardised__section--condensed' %}
{% endif %}
```
I just tried this quickly on twig level, then I realised that we have to distinguish between the EC and 
the EU component which does not need this custom class. It seems more complicated than expected to distinguish  EC or EU component at twig level?

src/specs/components/footer-standardised/demo/data--ec.js
papegaill: but clearly, I also want to remove these custom classes on the sections at markup level ... 
but using nth-child seems a bit tricky to me, both for the middle section and these last sections list

src/specs/components/footer-standardised/demo/data--ec.js
planctus: there is no other component needing to be aware of the "system" in use, so i wouldn't introduce for this, it was an idea to further simplify the specs that i had at the last moment, but i see is not viable..

src/specs/components/footer-standardised/demo/data--ec.js
planctus: well, basing on the usage page i'm now seeing that even the "class links" section is mandatory: https://ec.europa.eu/component-library/ec/standardised-template/footer/usage/, so the assumptions we could make are potentially even more...but ok, just we should not present this as a "use case", so the toggler for the class section should not be there.

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,94 @@
...
+        {% set _section_class_name = 'ecl-footer-standardised__section' %}
+        {% if _sections.section_class_name is defined and _sections.section_class_name is not empty %}
+          {% set _section_class_name = _section_class_name ~ ' ' ~ _sections.section_class_name %}
+        {% endif %}
+        <div class="{{ _section_class_name }}">
+          {# logo (EU) #}
+          {% if _sections.logo is defined and _sections.logo is not empty %}
+            {% set _label %}
+              <img
+                alt="{{ _sections.logo.alt | e('html_attr') }}"
```
planctus: it's bad that we don't have a linter in place for twig in ECL:
`_sections.logo.alt|e('html_attr')`

like all the other twig filters we set. The same in all the places where you added it.

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,94 @@
...
+            } only %}
+          {% endif %}
+          {# content before #}
+          {% if _sections.content_before is defined and _sections.content_before is not empty %}
+            <div class="ecl-footer-standardised__content">{{ _sections.content_before }}</div>
+          {% endif %}
+          {# title #}
+          {% if _sections.title is defined and _sections.title is not empty %}
+            {% set _title_class_name = 'ecl-footer-standardised__title' %}
+            {% if _sections.title_with_separator is defined and _sections.title_with_separator %}
```
planctus: isn't it enough: ` {% if _sections.title_with_separator %}`

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,94 @@
...
+            {% endif %}
+          {% endif %}
+          {# description #}
+          {% if _sections.description is defined and _sections.description is not empty %}
+            <div class="ecl-footer-standardised__description">{{ _sections.description }}</div>
+          {% endif %}
+          {# list #}
+          {% if _sections.links is defined and _sections.links is not empty and _sections.links is iterable %}
+            {% set _list_class_name = 'ecl-footer-standardised__list' %}
+            {% if _sections.links_inline is defined and _sections.links_inline %}
```
planctus: As before, it's a boolean, if it's false it would not pass: `{% if _sections.links_inline %}`

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,94 @@
...
+              {% set _list_class_name = _list_class_name ~ ' ecl-footer-standardised__list--inline' %}
+            {% endif %}
+            <ul class="{{ _list_class_name }}">
+              {% for _item in _sections.links %}
+                <li class="ecl-footer-standardised__list-item">
+                  {{- _item.content_before -}}
```
planctus: we normally do like if `strict_variables` is set to true, so we would check for `defined and not empty` for content_before and content_after

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,94 @@
...
+            {% set _label %}
+              <img
+                alt="{{ _sections.logo.alt | e('html_attr') }}"
+                title="{{ _sections.logo.title }}"
+                class="ecl-footer-standardised__logo-image-mobile"
+                src="{{ _sections.logo.src_mobile }}"
+              />
+              <img
+                alt="{{ _sections.logo.alt | e('html_attr') }}"
+                title="{{ _sections.logo.title }}"
```
planctus, line 77: `title` is an attribute too, escape it

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,94 @@
...
+                class="ecl-footer-standardised__logo-image-desktop"
+                src="{{ _sections.logo.src_desktop }}"
```
planctus, line 79: and `src` as well

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,94 @@
...
+        {% if _sections.section_class_name is defined and _sections.section_class_name is not empty %}
+          {% set _section_class_name = _section_class_name ~ ' ' ~ _sections.section_class_name %}
+        {% endif %}
+        <div class="{{ _section_class_name }}">
+          {# logo (EU) #}
+          {% if _sections.logo is defined and _sections.logo is not empty %}
+            {% set _label %}
+              <img
+                alt="{{ _sections.logo.alt | e('html_attr') }}"
+                title="{{ _sections.logo.title }}"
```
planctus, line 69: this `title` too needs to be escaped

planctus (changes requested):
The test is failing due to the removed classes in the middle row for EC, but like this it looks like we are changing the design, to me you can use the nth-child solution, the three rows are mandatory for the footer-standardised..
Some remarks on the twig template, still, but should be quick to get rid of them.. ;)

planctus:
I've just noticed that we broke the EU demo, `getSystem()` returns a lowercased string.. :)

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
planctus: And to be strict we should also check if the key is defined before trying to render it...

src/specs/components/footer-standardised/demo/data--ec.js
papegaill: Ok I'm removing the toggler 👍  The class name is not mandatory for the harmonised footer, so I prefer to keep same class and css logic for both footer as they are the same.

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,102 @@
...
+          {# list #}
+          {% if _sections.links is defined and _sections.links is not empty and _sections.links is iterable %}
+            {% set _list_class_name = 'ecl-footer-standardised__list' %}
+            {% if _sections.links_inline %}
+              {% set _list_class_name = _list_class_name ~ ' ecl-footer-standardised__list--inline' %}
+            {% endif %}
+            <ul class="{{ _list_class_name }}">
+              {% for _item in _sections.links %}
+                <li class="ecl-footer-standardised__list-item">
+                  {% if _sections.content_before is defined and _sections.content_before is not empty %}
```
planctus: you're checking for `_sections.content_before`, it should be `_item.content_before` :)

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,102 @@
...
+                    {{- _item.content_before -}}
+                  {% endif %}
+                  {% include '@ecl/link/link.html.twig' with _item|merge({
+                    link: _item.link|default({})|merge({ type: 'standalone' }),
+                    extra_classes: 'ecl-footer-standardised__link',
+                  }) only %}
+                  {% if _sections.content_after is defined and _sections.content_after is not empty %}
```
planctus: same here ;)

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
```diff
@@ -1,190 +1,114 @@
...
+  argTypes.hide_relate_site = {
+    name: system === 'eu' ? 'optional links' : 'related sites',
+    type: { name: 'boolean' },
+    defaultValue: true,
+    description:
+      system === 'eu'
+        ? 'Show "Optional links" section'
+        : 'Show "Related sites" section',
+    table: {
+      category: 'Use cases',
```
planctus: I know i proposed you to name it like this, but maybe we can do better naming it "Optional sections", what do you think?

planctus (changes requested):
I like it a lot, from the data we feed the component with, the template we have now and the html we render, it all makes sense to me.
For the attributes escaping, we will have to "rework" them soon, i guess, but fine that we anticipated a bit this "track". Fix just this small problem we introduced and for me it's good, let's then have Romain take a look at it and then we go for it!!
Ah, there is the `Migrating-v3.md` file in the docs folder that probably needs an update with the changes we made here, these are surely breaking ones :)

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
papegaill: Oups !

src/implementations/twig/components/footer-standardised/footer-standardised.story.js
papegaill: Thks 👍  I also prefer it makes more sense in accordance with the usage page.

papegaill:
Many thanks @planctus for this long and useful review 👍 
@emeryro do you want to take a look before updating other footers in this way?

src/implementations/twig/components/footer-standardised/footer-standardised.html.twig
```diff
@@ -49,103 +47,102 @@
 {% endif %}
 
 {# Print the result #}
-<footer class="{{ _css_class }}"{{ _extra_attributes }}>
-  {# Main container #}
+<footer class="ecl-footer-standardised">
```
emeryro: we should keep the option to have extra classes and attributes here

src/implementations/vanilla/components/footer-standardised/footer-standardised-ec.scss
```diff
@@ -155,96 +168,7 @@ $_column-gap: map.get(theme.$grid-gutter-width, 'm');
     margin-top: map.get(theme.$spacing, 'xl');
   }
 
-  .ecl-footer-standardised__list--columns {
-    column-count: 2;
-    column-gap: $_column-gap;
-  }
-
   .ecl-footer-standardised__link {
```
emeryro, line 176: I guess that the `.ecl-footer-standardised__list--columns` class (or something equivalent) will still be needed for the core footer, but we'll see then.

emeryro (changes requested):
Thanks for this, it looks nice!
I just have a couple of remarks:
- there is a minor difference on EC footer (I guess it is just a css class to be added, if it is more than that we may see with designers)
before
footer-ec-1-before
after
footer-ec-1-after
- I find the specs files harder to read, with lots of nested `[` and `{`. I hope that this structure won't be an issue for the users

papegaill:
> Thanks for this, it looks nice!
> I just have a couple of remarks:
> 
> * there is a minor difference on EC footer (I guess it is just a css class to be added, if it is more than that we may see with designers)
>   before
>   footer-ec-1-before
>   after
>   footer-ec-1-after
> * I find the specs files harder to read, with lots of nested `[` and `{`. I hope that this structure won't be an issue for the users

Thks @emeryro, I agree, lots of nested `[` and `{`, but it reflects the grid structure in the end.

emeryro (approved):
for me it's fine.
I would just suggest to add some comments in the specs files, to help understanding in which row/columns we are, but this could be done when dealing with other footers

planctus:
I was about to merge this @papegaill, but it looks like we have some conflicts to fix and a little issue with prettier introduced by some recent change in the story file.. ;)
Ah, but i see you want first to check also the accessibility remarks we received about the component, so i will wait for those.

papegaill:
> I was about to merge this @papegaill, but it looks like we have some conflicts to fix and a little issue with prettier introduced by some recent change in the story file.. ;)
> Ah, but i see you want first to check also the accessibility remarks we received about the component, so i will wait for those.

Thanks, I fixed prettier issue and updated the PR according to accessibility remarks, see FRONT-1455

planctus (approved):

PR #1903, opened 2021-01-26
feat(text-field): review text field and add eu version

emeryro:
specs:
- EC: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/b5328d41-9ea4-4a5f-bda0-ee748d9e10ef
- EU: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/2c68ac5e-aab0-4f1b-94ac-08df60d4261f

it has been validated by designers


src/implementations/twig/components/text-input/text-input.html.twig
```diff
@@ -101,6 +103,12 @@
 
   {% if _invalid and _invalid_text is not empty %}
     <div class="ecl-feedback-message">
+      {% if _invalid_icon is defined and _invalid_icon is not empty %}
```
planctus: We define `_invalid_icon` with a default value, therefore no need to check for `is defined`

planctus:
I read that the design has been approved already, so this comment is not about it but as it was in ECL v2 the disabled state opacity involves the whole component markup while in the specs on abstract only the input field is. Couldn't we ask to change then the specs if this is the desired result?

src/implementations/twig/components/text-input/README.md
```diff
@@ -11,11 +11,12 @@ npm install --save @ecl/twig-component-text-input
...
-- **"invalid_icon_label"** (string) (default: '')
+- **"invalid_icon"** (object of type "icon") (default: {})
 - **"required"** (boolean) (default: false)
 - **"name"** (string) (default: '')
 - **"type"** (string) (default: '')
-- **"width"** (string) (default: '') Input width size (s, m or l)
+- **"placeholder"** (string) (default: '')
+- **"width"** (string) (default: 'm') Input width size (s, m or l)
 - **"extra_label_classes"** (string) (default: '') Extra classes for the label (space separated)
 - **"extra_group_classes"** (optional) (string) (default: '') Extra classes (space separated) for the text-input group
```
planctus, line 21: there is actually a default for this parameter, 'ecl-form-group', maybe we can update this "table"

src/implementations/twig/components/text-input/text-input.story.js
```diff
@@ -1,19 +1,21 @@
...
 import withCode from '@ecl/storybook-addon-code';
 
-import dataDefault from '@ecl/specs-component-text-input/demo/data';
+import specsEc from '@ecl/specs-component-text-input/demo/data--ec';
+import specsEu from '@ecl/specs-component-text-input/demo/data--eu';
 
 import textInput from './text-input.html.twig';
 import notes from './README.md';
 
+const dataDefault = process.env.STORYBOOK_SYSTEM === 'EU' ? specsEu : specsEc;
```
planctus: Don't we use `getSystem() ` now to retrieve the system in use? Not a big deal..eh :)

src/implementations/twig/components/text-input/text-input.test.js
```diff
@@ -1,10 +1,13 @@
...
+import specDefaultEu from '@ecl/specs-component-text-input/demo/data--eu';
 
-const dataInvalid = { ...dataDefault, invalid: true };
-const dataDisabled = { ...dataDefault, disabled: true };
+const specInvalidEc = { ...specDefaultEc, invalid: true };
+const specDisabledEc = { ...specDefaultEc, disabled: true };
+const specInvalidEu = { ...specDefaultEu, invalid: true };
+const specDisabledEu = { ...specDefaultEu, disabled: true };
 
-describe('Text field', () => {
```
planctus, line 7: The structure of the test is a bit weird, i guess it was already, but we now added EU and they doubled, we test again the `extra whatever` for the `disabled` and `invalid` use case, which looks redundant, no reason why these would fail if the default doesn't. Wouldn't it be better to have also the `disabled` and the `invalid` as the `required` currently is, basically? Instead of "describing" them as `Default` we could describe them as `Text input EC` and `Text input EU` directly, also...

planctus (changes requested):
Pretty minor stuff, also i haven't checked the design specs since they were already approved, just gave them a look.

src/implementations/twig/components/text-input/README.md
emeryro: `ecl-form-group` is kind of hardcoded in the template, so that's not really the default value of any parameter

planctus (approved):

PR #1905, opened 2021-02-03
feat(taxonomy): add taxonomy to several components - FRONT-1721

emeryro:
specs: https://citnet.tech.ec.europa.eu/CITnet/confluence/pages/viewpage.action?spaceKey=NEXTEUROPA&title=Taxonomy++-+Design+specifications

the main idea for the implementation is to add a new variant to the description list, called "taxonomy", and to update the other components to add an optional description list in it.

updated components/compositions
- description list
- content item
- file download
- card


src/systems/ec/implementations/vanilla/packages/ec-component-card/ec-component-card.scss
```diff
@@ -121,6 +126,31 @@
...
+    max-width: 100%;
+
+    .ecl-description-list__term {
+      width: auto;
+    }
+
+    .ecl-description-list__definition {
+      margin-left: 0;
+      margin-top: $ecl-spacing-2-xs;
+      width: auto;
```
planctus, line 141: image

I found this a bit weird, in terms of design, then i looked at the code and i see that we are "overriding" the width set for this terms and definitions in the card component specifically, setting it to auto, but since there is a max width still the elements are limited and i wonder, is this on purpose? Checked the "specs" but i could not entirely clear this doubt myself, so i ask you.. ;)

src/website/src/pages/ec/components/file/docs/code.mdx
```diff
@@ -42,3 +43,14 @@ import demoContentThumbnail from '@ecl/ec-specs-file/demo/data--thumbnail';
 >
   <FileDownload {...demoContentThumbnail} variant="thumbnail" />
 </Playground>
+
```
planctus, line 46: I find the definition of this new "example" for the file confusing. For how it is built it's only going to be applicable to the thumbnail variant, but then why not updating the existing example, also in storybook? Like it is one could be tempted to consider this specific one (taxonomy) as a variant, applying the taxonomies to a "standard" file component and that would not work as expected.
Basing on this in twig, regardless of the data provided for a description list with taxonomies, the rendering of this list would only happen if the "thumbnail" variant has been "defined" and i would not add another independent example for it, just update the existing one for the "thumbnail", maybe.

planctus (approved):
To me it's good, it's also pretty clean, what i mean is that the new layout of the description list is reused with almost no overrides, just in the card i see we need some adaptations in terms of style, i left two comments, you can answer but still for me it's fine as it is.. ;)

PR #1922, opened 2021-02-09
feat(text-area): Review text area and add eu version - FRONT-1944

papegaill:
- ec: https://ecl-preview-1922--europa-component-library.netlify.app/playground/ec/?path=/story/components-forms-text-area--default
- eu: https://ecl-preview-1922--europa-component-library.netlify.app/playground/eu/?path=/story/components-forms-text-area--default


src/implementations/twig/components/text-area/text-area.test.js
```diff
@@ -124,45 +65,37 @@ describe('Text area', () => {
...
-        ],
-      });
-
-      return expect(render(optionsWithExtraClasses)).resolves.toMatchSnapshot();
+      return expect(render(dataInvalid)).resolves.toMatchSnapshot();
     });
   });
+};
+
+describe('Text field EC', () => {
```
planctus: Isn't this the test for the text area component..? :)

docs/Migrating-v3.md
```diff
@@ -166,3 +166,26 @@ Most of the old v2 presets have been removed or modified:
 | eu-preset-website | preset removed (replaced by an optional css) |
 | eu-preset-legacy | preset removed (no legacy content) |
 | eu-preset-legacy-website | preset removed (no legacy content) |
+
```
planctus: i'm a bit surprised by seeing all these lines added here, didn't check the version of the file before this update, but it looks like there should only be the lines related to the text area added here

src/presets/ec-core/src/ec-core.scss
```diff
@@ -3,7 +3,8 @@
 
 // EC specific
 // Forms
-@use '@ecl/vanilla-component-form/form-ec';
+@use '@ecl/vanilla-componeyarnnt-form/form-ec';
```
planctus: There is this "yarn" that sneaked into this string

src/implementations/vanilla/components/text-area/_text-area.scss
```diff
@@ -0,0 +1,110 @@
+/**
+ * Textarea component
+ * @define text-area
+ */
+
+@use "sass:map";
```
planctus, line 6: I see there is no rule involving the .ecl-form-group here, so most likely this is coming from v2, but i see in the specs (forget me for noticing this, i find it bit insane) that the margin between the help text and the textarea is different in the two systems...
In the form group css there is this about the textarea: 
```
.ecl-form-group > *:not(.ecl-u-sr-only) + .ecl-text-area {
    margin-top: 0.75rem; 
}
```
which is fine for EC while in EU the margin in the specs is 8px...

image

src/implementations/vanilla/components/text-area/_text-area.scss
```diff
@@ -0,0 +1,110 @@
...
+
+.ecl-text-area--l {
+  width: 100%;
+
+  @include breakpoints.up('l') {
+    width: map.get(theme.$form, 'l');
+  }
+}
+
+.ecl-text-area[disabled],
```
planctus, line 82: I say this but i guess it has been already decided to accept and live with this difference from the specs and the implementation, the opacity of the form group elements (label, help text..) is not altered when the form input is disabled, in the EU specs it is the case while in the EC ones it's not.
I just wonder, in case, this is already "agreed", the reason why the specs keep presenting this difference.

src/implementations/vanilla/components/text-area/_text-area.scss
```diff
@@ -0,0 +1,110 @@
+/**
+ * Textarea component
+ * @define text-area
+ */
+
+@use "sass:map";
+@use '@ecl/theme-dev/theme';
+@use '@ecl/vanilla-layout-grid/mixins/breakpoints';
+
```
planctus, line 9: This is also in the from group, but from what i see in the specs the label font-size is different than the one for the span (optional), in EC only, just mentioning this, i'm getting sick in seeing this minimal differences in the specs for the two systems, actually...

src/website/src/pages/ec/components/forms/text-area/demo/index.js
```diff
@@ -1,4 +1,4 @@
-import demoContentDefault from '@ecl/specs-component-text-area/demo/data';
+import demoContentDefault from '@ecl/specs-component-text-area/demo/data--ec';
```
planctus, line 1: Looks like we have to change something here, now we have default which is optional and then a duplicated example with the same optional textarea.

src/implementations/vanilla/components/text-area/_text-area.scss
```diff
@@ -0,0 +1,110 @@
...
+.ecl-text-area--m {
+  width: 100%;
+
+  @include breakpoints.up('m') {
+    width: 50%;
+  }
+
+  @include breakpoints.up('l') {
+    width: map.get(theme.$form, 'm');
```
planctus, line 70: didn't find anything in the specs in abstract about the different widths, did i miss it? Are we assuming that this is unchanged from v2? Just asking.. :)

src/implementations/vanilla/components/text-area/_text-area.scss
```diff
@@ -0,0 +1,110 @@
...
+  border: $_border-width solid $_border-color;
+  border-radius: $_border-radius;
+  box-shadow: $_shadow;
+  box-sizing: border-box;
+  color: $_color-input;
+  display: block;
+  font: map.get(theme.$font, 'm');
+  margin: 0;
+  max-width: 100%;
+  padding: calc(#{$_padding-vertical} - #{$_border-width})
```
planctus, line 38: Considering the differences in the border sizes in the different states i wonder a bit about the reason why we use border-box here, it forces us in making these crazy calculations and it actually alters the padding applied which is never corresponding to the one in the specs, but ok...

planctus (changes requested):
It's a small component and it was already super difficult to review it, feel that we should tell something about the fact that most of the differences designers set in the specs are irrelevant and due to them we are forced to define two different css, parametrize everything, i feel this s far away from the initial goal for the v3 approach.
Most of the comments i left in my own opinion can be ignored, surely there is the example in the website that needs to be updated, for the rest it is fine, i'd say, but all these little differences they make it really hard to say it with confidence..

src/implementations/vanilla/components/text-area/_text-area.scss
papegaill: yes and I also referred to the text-field component which is already validated for EU.

src/implementations/vanilla/components/text-area/_text-area.scss
papegaill: yes indeed, I think we will have to ask designer directly. I followed the same logic as `text field` already validated for EU.

src/implementations/vanilla/components/text-area/_text-area.scss
papegaill: yes I think this is to prevent all the elements below in the page from moving if we focus or different states.

src/implementations/vanilla/components/text-area/_text-area.scss
papegaill: Updated and verify for text field, same spacings 👍

planctus: isn't this still the case? in EU the text-area margin-top is supposed to be 0.5rem, 8px, i see 12px
Ah, but it seems you just swapped the changes, it was correct in EC and you changed it there, now it's 8px in EC while it should be 12px, none of these problems would exist if they would avoid such irrelevant misalignment between the two systems..

planctus: It seems to me that his H associated in EC to red color of the "invalid text" on hover means red hover, the same applied to the textarea when hovering it, but also in EU the color on hover should change consistently in the invalid text and the border of the form input
image

planctus:
Not sure about this, but maybe we should avoid changing style of the form input when it's disabled, i see the focus is still provoking a change in the border that then disappears because the element is disabled, as i said i'm not sure about this, it doesn't seem to be specified in the specs, but my feeling is that the element should stay "steady" when it's disabled.

papegaill:
> Not sure about this, but maybe we should avoid changing style of the form input when it's disabled, i see the focus is still provoking a change in the border that then disappears because the element is disabled, as i said i'm not sure about this, it doesn't seem to be specified in the specs, but my feeling is that the element should stay "steady" when it's disabled.

it is already the case with the form elements of EC components. I can ask the designer to clarify this point.

src/implementations/vanilla/components/text-area/_text-area.scss
papegaill: Swapped the form element top margin 👍 
Not sure to understand your last comment for the hover color of the invalid text. after checking the border colors correspond to the defined specs.

planctus: the border yes, but in the specs i see also the invalid text changing color consistently on hover, which is not happening in our implementation, also because we trigger the hover status of the text-area only. So, basically, it looks like this behaviour could be standardised in the form-group css, so that when hovering the form-group container both the input border and the invalid text would get the same, different color.

planctus (approved):

PR #1924, opened 2021-02-10
feat(search-form): review search form and add eu version - FRONT-1943

emeryro:
- ec: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/b5d91b9d-6145-4bc0-bcbf-030bec552a8f
- eu: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/7d05575a-7aee-4e96-ad5e-15764a13d55c


planctus (commented):
Since we rely on the designer's approval i haven't been looking deeply at the specs. The look and feel of the component when it's "invalid" is a bit odd, but i see no specification about it in the files on abstract, so clearly we rely on the standard behaviour of the text-field.
Just a small thing, we didn't disable the knobs in the story file, maybe we can do it now..

PR #1925, opened 2021-02-10
fix(page-headers): Update sass to have a padding-top if not breadcrumb - FRONT-1996

papegaill:


planctus (commented):
I think i've never looked before at all these pages with templates, it surely doesn't depend on these changes that are just improving them, but it seems they would need some other intervention, there is weird spacing under the page title depending on the chosen template, maybe we can create a ticket for handling this.
About this pr, instead, i think we can apply the same fix also to the deprecated page header, platform switched to the families but we still offer that "example" in storybook for ECL and release the twig template.
But cool, i don't know why we never went for fixing this and instead created dozens of pages with a "patch" on them...

HAL-Patch-INNO (dismissed):

planctus (approved):

PR #1931, opened 2021-02-24
feat(menu): Restyling EC, adding EU version - FRONT-1946

planctus:
Some considerations:

## EU

Responsiveness:

* The text will only wrap when there is no space available, the examples in the specs file are not realistic, we should set a max-width for that to happen but this is not advisable

* The dropdown styles for the different states are not specified as in others cases, the desktop styles reflect the "mobile first" definition of them

### Typography:

* mobile XL, not sure what that is about, no change has been applied to the font sizes in the mobile viewport.

* there is a variable definition of the hover and active state regarding the underlining, shouldn't this be a behaviour defined for all the links and not depending instead on the context in which this link appears? The attempt was to reproduce what's in the specs, anyway

### Shapes and colours

* Since the height of the menu in a desktop viewport is 4rems while the item height is 3.75rem it has been deduced there is a 4px margin/padding not specified anywhere.

* the font-size of the "sub-elements" in the examples for the mobile slide menu is smaller, in the typography file there is no difference, instead

### Spacing:.

* Might be me not knowing how to use abstract, but the smaller spacing is not readable in the downloadable pngs, in abstract if i zoom i can't then move the image to see what i want to look at

### Typography:

* Slide-in level 1 items have no difference between default and hover, is that how it is intended to be?

* Slide-in level 2 items are bold in the typography file table, not in the other example where only the active item seems to be bold, this is how it is in the implementation

* The file on abstract seems to have disappeared


src/implementations/vanilla/components/menu/_menu.scss
```diff
@@ -7,23 +7,89 @@
...
+$_menu-sublink-desktop-padding: null !default;
+$_menu-sublink-mobile-vertical-padding: null !default;
+$_menu-sublink-outline-offset: null !default;
+$_menu-sublink-active-color: null !default;
+$_menu-sublink-active-weight: null !default;
+$_menu-sublink-active-background: null !default;
+$_menu-sublink-hover-color: null !default;
+$_menu-title-font-weight: null !default;
+$_mobile-item-color: null !default;
+$_mobile-item-hover-background: null !default;
```
emeryro, line 74: wow, that's a lot of variables!
that's a good example of "small differences" between EC and EU going out of control

emeryro (changes requested):
Thanks for this, I know that working on the menu component is always something complex...

I spotted a few things already:
EC:
- there is an issue with the selected item in mobile menu
image

EU
- the design is kind of odd, with lots of empty space, but it seems to be conform to the specs... Maybe we could use some more realistic label for menu items to make it look better
- there is some css transition when using the tab key to focus links. I wonder if we could get rid of it
- desktop: I assume that there should be a grey background and no underline when hovering items in the dropdown menu (but I agree that it isn't very clear in the specs)
- mobile: there should also be a grey background and no underline on hover on the second level navigation

emeryro (approved):

PR #1934, opened 2021-03-01
fix(gallery): fix gallery optional elements - FRONT-2060

emeryro:
prevent js error when some optional elements are not set


planctus (approved):
Cool, i think it would have been good also to remove meta and description from at least one item in the specs, this would have revealed that, like we are doing in twig, also in react we still render the <figcaption> that appears empty when hovering the item in the gallery.
Anyway, since we don't have knobs in storybook for removing these props set from the specs, there is no real issue.

PR #1935, opened 2021-03-01
feat(checkbox): Review EC, adding EU version - front-2054

planctus:
### Remarks:

1. The specs introduced some change in the behaviour of the component
-  when there is a single checkbox the error text is rendered below the form element

2. The specs introduced some conditional spacing based on the presence of the help text. This is handled by setting the spacing for the single checkbox and reducing it by a negative margin set on the helper-text when it's present.

3. The specs introduced a specific style for the  "indeterminate" state of the checkbox, **this has not been implemented**

4. In Eu there is a different spacing for the first form element, a new variable has been introduced in the form group sass file.


src/implementations/vanilla/components/checkbox/_checkbox.scss
```diff
@@ -74,47 +83,29 @@ $_outline-width: 2px;
...
   font: map.get(theme.$font, 'xl');
   font-weight: map.get(theme.$font-weight, 'bold');
   padding-left: map.get(theme.$spacing, '2xs');
 }
-// Checked
+// Checked and indeterminate
 // stylelint-disable no-descending-specificity
-.ecl-checkbox__input:checked {
+.ecl-checkbox__input:checked,
+.ecl-checkbox__input:indeterminate {
```
emeryro: does it makes sense to keep this indeterminate here?

src/implementations/twig/components/checkbox/checkbox.story.js
```diff
@@ -1,14 +1,27 @@
...
+import dataEu from '@ecl/specs-component-checkbox/demo/data--eu';
 
 import checkboxGroup from './checkbox-group.html.twig';
 import notes from './README.md';
 
-const dataInvalid = { ...dataDefault, invalid: true };
-const dataOptional = { ...dataDefault, required: false };
+const system = getSystem();
+
+let dataDefault = { ...dataEc };
```
emeryro: can't we have something like 
`let dataDefault = system === 'eu' ? { ...dataEu } : { ...dataEc };`?
and so we don't have to add the icon afterward

src/implementations/vanilla/components/checkbox/_checkbox.scss
planctus: no, i didn't even remember i put this in place.. :)

src/implementations/twig/components/checkbox/checkbox.story.js
planctus: wow, thanks for noticing this, i really can't understand the reason why i was doing this that way...

emeryro (approved):

PR #1938, opened 2021-03-01
feat(footer-harmonised): Review markup and add controls - FRONT-1845

papegaill:


src/implementations/twig/components/footer-harmonised/eu-footer-harmonised.story.js
```diff
@@ -0,0 +1,92 @@
...
+    res.rows[0].splice(2, 1);
+  }
+  if (!args.hide_contact && !args.hide_follow) {
+    res.rows[0].splice(1, 1);
+  }
+
+  return res;
+};
+
+export const Default = (args) => footer(prepareDataG1(specsEu, args));
```
emeryro, line 75: do we really need to have this whole file, if we don't display EU harmonized footer at all?

src/implementations/vanilla/components/footer-harmonised/footer-harmonised-eu.scss
```diff
@@ -0,0 +1,176 @@
+/**
+ * Footer Harmonised component
```
emeryro, line 2: same question here: is this css actually used somewhere?

src/specs/components/footer-harmonised/demo/data--eu.js
```diff
@@ -0,0 +1,349 @@
+module.exports = {
+  rows: [
```
emeryro, line 2: same here :)

src/implementations/twig/components/footer-harmonised/footer-harmonised.html.twig
```diff
@@ -1,21 +1,19 @@
 {% spaceless %}
 
-{#
+{# 
   Parameters:
   - "group" (string) (default: 'group1'): 'group1', 'group2' or 'group3'
-  - "sections" (array of objects) (default: []):
+  - "rows" (array of objects) (default: [])
```
emeryro, line 6: we should update the readme file of the component too

emeryro (changes requested):
thanks for this!
I spotted a small design difference on EC harmonized group 1:
before:
image

after:
image

src/implementations/twig/components/footer-harmonised/eu-footer-harmonised.story.js
papegaill: Thanks @emeryro
The story for EU has a different name from the one of EC 
if you use the EC | EU switcher, then yes you end up with a blank page.

EU: https://ecl-preview-1938--europa-component-library.netlify.app/playground/eu/?path=/story/components-footers-core--default

emeryro (commented):
that's good to me. Could you just update the tests snapshots? Then we can merge it

emeryro (approved):

PR #1939, opened 2021-03-02
feat(link): update EC link and add EU

emeryro:
- spec EC: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/8ffc7526-37c9-4146-9157-6d30e11fe18c
- specs EU: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/3cf127d4-0d51-43bb-90aa-f2f5ba76029e

notes:
- as the link component were used in lots of different other components, I had to keep the css call order, to prevent side effects. So the css files are now called directly in the corresponding preset (ec-core and eu-core), and no longer in dev preset


planctus (commented):
just a few considerations.. :)

- out of curiosity, this was already defined as such, but I fail to understand the need for the two cases of the story as a string or as a node, do you know when the story is going to be a "node"? :)
- i think we have an issue with the "viewport" in v3, we reset all the padding i guess but when you see the standalone link with focus only half of the outline is going to be visible
- I find a bit confusing the presence of an example for the negative variant and the control that applies the negative in all the examples, but anyway, we should at least disable the control for the negative story itself and the cta one, where it is not meant to be used
- I love the fact that we have now css to separate the link and the icon, i do wonder why it was not always the case, i think the approximation you made is correct, but the result with 0.5 rem is visually increasing the spacing more than the "pixels" would say, maybe 0.25 rem would give less the impression that something has changed (i know it is not, in math terms..)
- I'm ok with moving the list of components to include into the themes, there should not be any risk of proliferation of those themes..
- I guess that next step would be to apply this new variant where it fits, no..? like footers, but i guess you already considered this.
- I did not check the specs, i leave that to the designers

emeryro:
to answer a few of your points:
> out of curiosity, this was already defined as such, but I fail to understand the need for the two cases of the story as a string or as a node, do you know when the story is going to be a "node"? :)

I was wondering the same.. I kept it as I was not the one introducing this test before, but if we agree that there is no point, I'll remove it

> i think we have an issue with the "viewport" in v3, we reset all the padding i guess but when you see the standalone link with focus only half of the outline is going to be visible

I'd say that in most cases, it is better to not have padding (especially for full width elements). But we could add some exceptions in specific stories I guess

> I find a bit confusing the presence of an example for the negative variant and the control that applies the negative in all the examples, but anyway, we should at least disable the control for the negative story itself and the cta one, where it is not meant to be used

We'll see what COMM requests for the example

> I love the fact that we have now css to separate the link and the icon, i do wonder why it was not always the case, i think the approximation you made is correct, but the result with 0.5 rem is visually increasing the spacing more than the "pixels" would say, maybe 0.25 rem would give less the impression that something has changed (i know it is not, in math terms..)

I took the value 0.5rem from the specs

> I guess that next step would be to apply this new variant where it fits, no..? like footers, but i guess you already considered this.

Indeed, that will be done in future tickets

PR #1946, opened 2021-03-08
fix(card): Preventing text overflow - FRONT-2092

planctus:


emeryro (approved):
works fine.
This should be applied to ECL 3 too

PR #1947, opened 2021-03-08
feat(select): Review EC, adding EU version - FRONT-2078

papegaill:
EC
- specs: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/fe2a43ce-5f28-4591-805d-44b6a53abc0e
- preview: https://ecl-preview-1947--europa-component-library.netlify.app/playground/ec/?path=/story/components-forms-select--single

EU
- specs: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/f2d0c51d-0ade-4ad6-bffa-2f174896061f
- preview: https://ecl-preview-1947--europa-component-library.netlify.app/playground/eu/?path=/story/components-forms-select--single


planctus:
I see that focusing on the element "moves" the text inside the box
select

src/implementations/vanilla/components/select/select.js
```diff
@@ -424,11 +440,31 @@ export class Select {
...
+    if (visible.length === 0 || visible.length !== checked.length) {
       this.selectAll.querySelector('input').checked = false;
     } else {
       this.selectAll.querySelector('input').checked = true;
     }
+    // Display no-results message.
+    const noResultsElement = this.searchContainer.querySelector(
+      '.ecl-select__multiple-no-results'
+    );
+    if (visible.length === 0 && noResultsElement === null) {
```
planctus: Being a parameter feeding the component, the availability of the text for the "no result" case might not be present, i think we can add a condition here so that we don't process it in case there is no label for it (unless we want to set a default on our side..)

src/implementations/vanilla/components/select/select.js
```diff
@@ -361,6 +373,10 @@ export class Select {
    */
   handleClickSelectAll(e) {
     e.preventDefault();
+    // Early returns.
```
planctus, line 376: i don't understand this "early return" logic, how is the "select all" accessible when the select is disabled?

src/implementations/vanilla/components/select/select.js
planctus: It's a correct check  `=== null` but why not simply writing `!noResultsElement`? Up to you, as i said there is nothing wrong about this. Same happens a few line later and same approach can be used.

planctus:
i see that we are fixing some behaviours of the multiple select, cool, but to me the biggest problem it has is still in place, i wonder if this could be the moment to tackle it. I find that the lack on any handling of the ui when multiple values are selected is pretty bad.
image

This was already the case but i see that while focusing we have this now:
image
and this seems new.

planctus:
I wonder why we don't add also a parameter for the "unselect all", the checkbox keeps showing the same label while having an opposite behaviour. Un-checking "select all" could be intended as the possibility of checking everything quickly for then being able to remove a single option, for instance, the way our "select all" checkbox works instead is un-checking all the "visible" checkboxes, so it should have a different label, imho. I wonder why this concern was not raised by the usability team...

src/implementations/twig/components/select/__snapshots__/select.test.js.snap
```diff
@@ -618,7 +619,8 @@ exports[`Select With error renders correctly 1`] = `
   <div
     class="ecl-feedback-message"
   >
-    This is the error message
+
```
planctus: Why do we have this empty line added in the test..?

src/implementations/twig/components/select/select.html.twig
```diff
@@ -113,6 +118,12 @@
 
   {% if _invalid and _invalid_text is not empty %}
     <div class="ecl-feedback-message">
+      {% if _invalid_icon is not empty %}
```
planctus: This is the reason why we have the empty line in the test.. :) We need to use dashes `-` in this case to avoid it.

src/implementations/vanilla/components/select/_select.scss
```diff
@@ -100,7 +125,9 @@ $_arrow-width: 2.75rem;
 
 .ecl-select__icon {
   align-items: center;
-  background-color: $_border-color;
+  background-color: $_arrow-background-color;
+  border-bottom-right-radius: $_border-radius;
```
planctus, line 139: Isn't it easier setting the shortcut border-radius..? One line instead of two, but just a thing i've noticed, nothing wrong with this..

planctus (changes requested):
I haven't checked anything related to the specs, just left some comments here about some visual bug that i noticed, a few suggestions on the code and some remarks on the component itself for how it already was in v2. Thanks!

src/implementations/vanilla/components/select/_select.scss
papegaill: no otherwise we would have rounded edges on the left, and the caret button only has one on the right for EU.

papegaill:
> I wonder why we don't add also a parameter for the "unselect all", the checkbox keeps showing the same label while having an opposite behaviour. Un-checking "select all" could be intended as the possibility of checking everything quickly for then being able to remove a single option, for instance, the way our "select all" checkbox works instead is un-checking all the "visible" checkboxes, so it should have a different label, imho. I wonder why this concern was not raised by the usability team...

Thank you for your review @planctus, I will submit this remark to the designer during validation!

PR #1962, opened 2021-03-18
feat(site-header-core): Review EC, adding EU version - FRONT-1947

papegaill:
EC:

- preview: https://ecl-preview-1962--europa-component-library.netlify.app/playground/ec/?path=/story/components-site-headers-core--default
- specs: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/a47e2bc7-65ae-4a20-b7c0-ad54a88921d9

EU:

- preview: https://ecl-preview-1962--europa-component-library.netlify.app/playground/eu/?path=/story/components-site-headers-core--default
- specs: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/cd8844a3-e74e-4bd7-9a6f-1a2c9df6dbe0

### DIFF
--- a/src/implementations/twig/components/site-header-core/site-header-core.story.js
+++ b/src/implementations/twig/components/site-header-core/site-header-core.story.js
@@ -35,17 +35,15 @@ const getArgTypes = () => {
     },
   };
 
-  if (system === 'eu') {
-    argTypes.menu = {
-      type: { name: 'boolean' },
-      defaultValue: true,
-      description: 'Toggle menu visibility',
-      table: {
-        type: { summary: 'object' },
-        defaultValue: { summary: '{}' },
-      },
-    };
-  }
+  argTypes.menu = {
+    type: { name: 'boolean' },
+    defaultValue: true,
+    description: 'Toggle menu visibility',
+    table: {
+      type: { summary: 'object' },
+      defaultValue: { summary: '{}' },
+    },
+  };
 
   return argTypes;
 };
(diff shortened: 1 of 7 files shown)


src/implementations/twig/components/site-header-core/site-header-core.story.js
```diff
@@ -35,17 +35,15 @@ const getArgTypes = () => {
...
-        type: { summary: 'object' },
-        defaultValue: { summary: '{}' },
-      },
-    };
-  }
+  argTypes.menu = {
+    type: { name: 'boolean' },
+    defaultValue: true,
+    description: 'Toggle menu visibility',
+    table: {
```
planctus: Small thing, but i think that when the controls are nothing related to the data we should simply remove any information from the "table", it doesn't make sense to specify the data type of the control and the default value..

planctus (approved):
I've been just looking at the changes in the story file, basically, i leave the review of the design specs to the designers.
There is just one comment about the controls we set, but it's really a small thing.

src/implementations/twig/components/site-header-core/site-header-core.story.js
papegaill: Thks @planctus, PR updated.

emeryro (changes requested):
there seems to be a display issue with the EC logo size on mobile (it may be very big)
image

papegaill:
> there seems to be a display issue with the EC logo size on mobile (it may be very big)
> image

@emeryro I put back the rule for the logo in mobile format for the EC version, it should be better now.

emeryro (approved):

PR #1963, opened 2021-03-18
feat(website): Adding EU structure - FRONT-2100

planctus:


emeryro (commented):
To me that's fine.
One update that we could do if the structure on EU pages is the same as EC is to be able to switch from one system to another without losing the current page

planctus:
interesting idea..to be seen ;)

emeryro:
I see that the jira task is already closed.
So should we do anything on this PR?

planctus:
We should merge it at some point, so to make the pages available for jean to work on them.. :)
But i still need to have a confirmation on her side that the pages are ok, she should do this today, so let's wait.. ;)

PR #1971, opened 2021-03-31
feat(radio): Adding EU styles, revising EC - FRONT-2103

planctus:
Some notes:

- there is a little inconsistency about the margin for the error message in EU, i kept as it is after we aligned all the form elements, but in the specs for the radio the spacing is xs instead of 2xs.

- there is one use case in the specs that has not been covered, there is no such a thing as a single radio button with associated "required" or "optional" information.

- There is one use case which is not entirely clear, hovering and invalid, checked radio button. I don't think it even make sense as such, checked and invalid, anyway the hovering at the moment is not changing the border color like it does for a non checked radio button.


emeryro (changes requested):
there is a small misalignment on binary radio:
image

emeryro (approved):

PR #2010, opened 2021-04-14
docs(migration): update migration doc - FRONT-2180

emeryro:


docs/Migrating-v3.md
```diff
@@ -2,51 +2,159 @@
 
 The following guidelines aim to facilitate migration between ECL v2 to v3.
 
-## System requirements
```
planctus, line 5: We should add this system requirements paragraph in the main README, we had already a couple of situations were users had problems due to the version of node they were using.
I wanted to merge this @emeryro , but now that we merged the eu icons we introduced some conflict here, let's fix those and we update this doc, thanks for taking care of it ;)

emeryro:
Conflict solved

PR #2021, opened 2021-04-20
feat(messages): Adding EU styles, revamping EC - FRONT-2152

planctus:
This pr has been already approved by dg-comm.


src/implementations/vanilla/components/message/_message.scss
```diff
@@ -31,26 +48,31 @@ $_close-color: map.get(theme.$color, 'black-100');
...
   bottom: map.get(theme.$spacing, 'm') - map.get(theme.$spacing, 's');
   color: $_close-color;
-  font: map.get(theme.$font, 's');
+  font: map.get(theme.$font, 'm');
+  font-weight: $_close-font-weight;
   position: absolute;
   right: 0;
+  // stylelint-disable-next-line plugin/selector-bem-pattern
+  .ecl-icon {
+    fill: map.get(theme.$color, 'blue-100');
```
emeryro: ecl-icon is using `fill: currentColor;`, which is also blue-100, so I don't think that we need to override it here

src/implementations/twig/components/message/message.html.twig
```diff
@@ -53,7 +53,7 @@
     extra_classes: 'ecl-message__icon',
   } only %}
   {% if _close.icon is defined and _close.icon is not empty %}
-    {% set _icon = _close.icon|merge({ size: 's', name: 'close' }) %}
+    {% set _icon = _close.icon|merge({ name: 'close-filled' }) %}
```
emeryro, line 56: We should add a line in the migration doc to say that the size of the icon is no longer set by default in the template; this may need some action on integration side

emeryro (changes requested):
Small things

emeryro (approved):

PR #2024, opened 2021-04-21
Add wt-btn exclusion rule for ecl-editor

haringsrob:
Hi,

As seen on: https://europa.eu/year-of-rail/index_en the ecl-editor overwrites some styles provided by a webtools component.

I proposed a fix, but please feel free to implement a different solution.

Kind regards,

### DIFF
--- a/src/systems/ec/implementations/vanilla/packages/ec-editor/_ec-editor--button.scss
+++ b/src/systems/ec/implementations/vanilla/packages/ec-editor/_ec-editor--button.scss
@@ -12,7 +12,7 @@
   $padding-vertical: $ecl-spacing-m
 ) {
   /* stylelint-disable selector-no-qualifying-type */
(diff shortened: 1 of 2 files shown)


planctus:
Hi @haringsrob!! It seems we already handled this here: https://github.com/ec-europa/europa-component-library/pull/1957
It's not yet released, though, will be done in 2.38.0
If you're fine with the solution we applied there you can close this pr ;)
Thanks!

haringsrob:
Hey @planctus!

Thanks for the quick reply, if it is fixed it is fixed :). I will update the story on our end and point to the already made fix.

Is there any eta on 2.38.0? 

Kind regards,

planctus:
Apparently the "ETA" for 2.38.0 is beginning of May, Rob..

haringsrob:
Great thank you!

PR #2037, opened 2021-04-23
feat(timeline): Adding EU styles, revamping EC - FRONT-2151

planctus:
### Remarks:

- It is not specified in the specs, but the hover effect (background color) has been kept for the bullets.

- Due to the differences in the item spacing in the two systems, the position of the toggle button is different, it is implemented as in the specs but i'm not sure it looks how intended.

- The dashed border for the toggle is not anything we can customise via css, it could be done potentially with a background-image but due to the small visual impact of the element this has not been implemented.

- There is some missing specification, like the vertical position of the arrow in the EU items, it has been placed more or less where it seems to be in the mocks.

- The overall max-width of the component, not specified in the mocks, has been kept as in v2 (633px)

- The overall behavior of the toggler has not been altered, still configurable with two parameters, "hide from", "hide to".

- Due to the styles requested for EU additional markup has been added to the item of the timeline, this is also present in EC but not styled to keep one single template without system "awareness".


src/implementations/twig/components/timeline/timeline.html.twig
```diff
@@ -73,11 +73,14 @@ Parameters:
         {% if _item.id is defined and _item.id is not empty %}
           id="{{ _item.id }}"
         {% endif %}>
-        <div class="ecl-timeline__label">{{ _item.label }}</div>
-        {% if _item.title is defined and _item.title is not empty %}
-        <div class="ecl-timeline__title">{{ _item.title }}</div>
-        {% endif %}
-        <div class="ecl-timeline__content">{{ _item.content|default('') }}</div>
+        <div class="ecl-timeline__tooltip">
```
emeryro, line 76: we should add a line about this markup update in the migration doc

emeryro (changes requested):

emeryro (approved):

PR #2039, opened 2021-04-26
feat(media-container): Adding EU styles, revamping EC - FRONT-2177

papegaill:
Previews:

- EC: https://ecl-preview-2039--europa-component-library.netlify.app/playground/ec/?path=/story/components-media-container--image
- EU: https://ecl-preview-2039--europa-component-library.netlify.app/playground/eu/?path=/story/components-media-container--image

Specs: 

- EC: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/50cbf540-67ef-4f85-a760-7231b68adac4
- EU: https://app.abstract.com/projects/2ffb1208-5760-4ed8-ab75-a3ec8e2a4ed2/branches/master/collections/d11b49da-2df9-4912-af9e-092a231b86bd


planctus (commented):
Interesting use case, basically here you considered that there is no common ground between EC and EU and you simply defined two separated scss files.
I haven't looked at the specs, but it seems it's the opposite for the print use case, we can live out of a common style..

emeryro (changes requested):
For the scss files, we should use the same structure as other components:

- _media-container.scss (common style and variables)
- media-container-ec.scss (implementation of variables for ec)
- media-container-eu.scss (implementation of variables for ec)

The only exceptions were for branding components

papegaill:
> For the scss files, we should use the same structure as other components:
> 
> * _media-container.scss (common style and variables)
> * media-container-ec.scss (implementation of variables for ec)
> * media-container-eu.scss (implementation of variables for ec)
> 
> The only exceptions were for branding components

What about the little visual yellow border? 
Make a `display: none` for the EC version. 
I'm not sure it's very optimal..

emeryro:
What has been done in the timeline PR is to add this kind of very specific element in the EU css of the component
https://github.com/ec-europa/europa-component-library/pull/2037/files#diff-22585d50a67d7727c69667b76057bdf58abed00320c4097121c404f11151c818R43

emeryro (approved):

PR #2041, opened 2021-04-26
feat(fact-figures): Adding EU styles, revamping EC - FRONT-2169

papegaill:
Previews:

- EC: https://ecl-preview-2041--europa-component-library.netlify.app/playground/ec/?path=/story/components-fact-figures--columns-3
- EU: https://ecl-preview-2041--europa-component-library.netlify.app/playground/eu/?path=/story/components-fact-figures--columns-3

Specs: 

- EC: https://share.goabstract.com/c83619ca-ceb4-4a43-8e05-6b4684629fed
- EU: https://share.goabstract.com/2ae34360-af9e-4833-b599-9907f5210ee8


src/implementations/twig/components/fact-figures/fact-figures.story.js
```diff
@@ -18,7 +18,7 @@ const viewAll = { ...data3Col.view_all };
 const getArgTypes = (data) => {
   return {
     icon: {
-      defaultValue: data.items[0].icon.name,
+      defaultValue: system === 'ec' ? 'digital' : data.items[0].icon.name,
```
emeryro: why do we force the use of this specific icon?

src/implementations/vanilla/components/fact-figures/fact-figures-ec.scss
```diff
@@ -14,20 +14,16 @@
 }
```
emeryro: on the mockups there is no background color on EC

emeryro (changes requested):
For the scss files, we should use the same structure as other components:

- _fact-figures.scss (common style and variables)
- fact-figures-ec.scss (implementation of variables for ec)
- fact-figures-eu.scss (implementation of variables for ec)

src/implementations/vanilla/components/fact-figures/fact-figures-ec.scss
papegaill: Missing on the mockups, designers 'll be updated.

src/implementations/twig/components/fact-figures/fact-figures.story.js
papegaill: On EC version, they are using the same `digital` icon everywhere, on EU they are using a different icon for each items.

emeryro: I don't think we need to keep this difference, that's just an example.
We could use the same content for EC and EU (including the icons)

emeryro (approved):

PR #2053, opened 2021-05-03
chore(release):  Making packages public, updating license

planctus:


emeryro:
shouldn't we change the package version too, before any release?
something like 3.0.0-alpha.1

emeryro (approved):

PR #2055, opened 2021-05-03
feat(media-container): add full width - FRONT-2224

emeryro:
Add full width feature to media container

preview: https://ecl-preview-2055--europa-component-library.netlify.app/playground/ec/?path=/story/components-media-container--image

note: controls for banners are handled in the dedicated PR


papegaill (changes requested):
I don't quite understand the logic of the full-width select here. why not just have a toggle with the value "full width inside a grid container" to demonstrate this case only? I have the impression that the other two cases are not necessary and confusing.

same approach as what was done yesterday for v2?

emeryro:
@papegaill the "recommended" way to have the media container full width is to use it outside the grid container, that's why I wanted to have an example like that.
Then I added the full width example, when inside a container (the main goal of this update), and the one inside the container, to clearly see the difference.
Hope it clarify things

papegaill:
> @papegaill the "recommended" way to have the media container full width is to use it outside the grid container, that's why I wanted to have an example like that.
> Then I added the full width example, when inside a container (the main goal of this update), and the one inside the container, to clearly see the difference.
> Hope it clarify things

@emeryro I understand but for me it seems clearer to just have a toggle with the full-width option inside a grid container.

emeryro:
I updated the control to make it clearer, after a suggestion of Davide.
I think that it could be a good compromise

PR #2064, opened 2021-05-10
fix(menu): Review menu spacings - FRONT-2156

papegaill:
- EC: https://ecl-preview-2064--europa-component-library.netlify.app/playground/ec/?path=/story/components-navigation-menu--default
- EU: https://ecl-preview-2064--europa-component-library.netlify.app/playground/eu/?path=/story/components-navigation-menu--default

### DIFF
--- a/src/implementations/vanilla/components/menu/_menu.scss
+++ b/src/implementations/vanilla/components/menu/_menu.scss
@@ -139,14 +139,6 @@ $_mobile-item-hover-background: null !default;
 }
 
 /* stylelint-disable-next-line order/order */
-@include breakpoints.up('m') {
-  .ecl-menu {
-    padding-bottom: map.get(theme.$spacing, 'm');
-    padding-top: map.get(theme.$spacing, 'm');
-  }
-}
-
-/* stylelint-disable-next-line order/order */
 @include breakpoints.up('l') {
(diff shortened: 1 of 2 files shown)


src/implementations/vanilla/components/menu/_menu.scss
```diff
@@ -138,14 +138,6 @@ $_mobile-item-hover-background: null !default;
   }
 }
 
-/* stylelint-disable-next-line order/order */
-@include breakpoints.up('m') {
-  .ecl-menu {
```
emeryro, line 143: it seems that this spacing change is still needed, at least on EU:
image
On EC, it seems to be the case too, even if that's not very clear...
image

emeryro (changes requested):
The menu items are fine now, thanks.

src/implementations/vanilla/components/menu/_menu.scss
papegaill: Thks for review @emeryro, this was already validated by Erika in the comment of the ticket. Specs not being clear during implementation. After checking, Erika also found that the spacing for tablet was much too large, especially without a site name. Do you want to have this checked by the designers one last time?

emeryro: If you say that they validated, it is fine for me

PR #2065, opened 2021-05-12
feat(file): Adding EU styles, revamping EC - FRONT-2178

papegaill:
Previews:
- EC: https://ecl-preview-2065--europa-component-library.netlify.app/playground/ec/?path=/story/components-file--without-translation
- EU: https://ecl-preview-2065--europa-component-library.netlify.app/playground/eu/?path=/story/components-file--without-translation

Specs:
- EC: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/ECL/EC+file+download
- EU: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/ECL/EU+file+download

The `tags` and `description-list` in the taxonomy variant have not been adapted for the EU version, I guess we have to wait for the EU style of these components that have not yet been processed.


src/implementations/twig/components/file/file.html.twig
```diff
@@ -88,13 +88,13 @@
...
   <div class="ecl-file__container">
   {% if _variant == 'thumbnail' %}
-    {% if _label is not empty %}
-      <div class="ecl-file__label">
-        {% include '@ecl/label/label.html.twig' with _label only %}
-      </div>
-    {% endif %}
     <div class="ecl-file__detail">
       <div class="ecl-file__detail-info">
+      {% if _label is not empty %}
```
emeryro, line 93: we should put in the migration doc this markup update

src/implementations/vanilla/components/file/file-eu.scss
```diff
@@ -0,0 +1,312 @@
+/*
+ * File
+ * @define file; weak
+ */
+
```
emeryro: we are duplicating the whole css here. Isn't it possible to have the same structure as the other components?

emeryro (changes requested):

src/implementations/vanilla/components/file/file-eu.scss
papegaill: I started with this option, but if you do a diff between the two scss files, you'll notice that there are many differences, a lot of properties are set for one version and not the other also in the media queries we have a lot of differences. I therefore preferred to leave two separate files to keep minimum css declaration for both version.

emeryro: well, I did run a diff, and to me the structure is kind of similar. There are of course some differences in terms of spacing, color and shape (border and border radius), but that's the case for the other components too.
Also if there are are other differences, like the thumbnail behavior on mobile, we should challenge them with designers, to see if we could have the same thing on both systems

emeryro (dismissed):

emeryro (approved):

PR #2071, opened 2021-05-18
feat(tags): Adding EU styles, revamping EC - FRONT-2124

papegaill:
Specs:

- EC: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/ECL/EC+Tags
- EU: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/ECL/EU+Tags

Previews:

- EC: https://ecl-preview-2071--europa-component-library.netlify.app/playground/ec/?path=/story/components-tag--display
- EU: https://ecl-preview-2071--europa-component-library.netlify.app/playground/eu/?path=/story/components-tag--display


emeryro (dismissed):
Fine for me, we just need to get validation from designers

src/implementations/twig/components/tag/tag.html.twig
```diff
@@ -67,11 +67,10 @@
   <a href="{{ _tag.path }}" class="{{ _css_class }}" {{ _extra_attributes|raw }}>
     {{- _tag.label -}}
   </a>
-{% else %}
-  <span class="{{ _css_class }}">
+{% elseif _tag.type == 'removable' %}
+  <a href="#" class="{{ _css_class }}" {{ _extra_attributes|raw }}>
```
emeryro: not sure why we changed it to a link for removable tags.. Was that a request from COMM?
in any case, the href shouldn't be hardcoded here

papegaill: @emeryro, thank you.. I changed the hardcoded link 👍 
Yes this is a request from DGcomm: "_As the only action the user can do with the removable tag is to remove the tag, the full tag should be "actionable"/clickable, which is why the hover and focus state should be on the full tag and not only the X icon."_

PR #2072, opened 2021-05-19
feat(html tags): add default styling - FRONT-2213

emeryro:
Add optional styling for some HTML tags.
The documentation has been updated, to provide additional information.

Notes:
- Optional css files are provided in both packages. They are names `ecl-ec-default` and `ecl-eu-default`
- Compilation is done using Sass placeholder and %extend:
  - the placeholder is put in the component's css, to select the rules that have to be used (no refactoring needed)
  - it is then called in the `default` scss file (no duplication)
  - the main drawback is that some unneeded css rules are integrated in the compiled css. This won't break anything, but is not very clean
- I had to disable the linter for BEM in some files, as it didn't like this structure...
- I consider it as a first version, to be improved, but it should works well already


src/implementations/twig/utilities/default/index.story.js
```diff
@@ -0,0 +1,74 @@
+import withCode from '@ecl/storybook-addon-code';
+
+export default {
+  title: 'Utilities/HTML tag styling',
+  decorators: [withCode],
+  parameters: {
```
planctus: I think we should enable the html tags css in this story by default.

src/implementations/vanilla/components/blockquote/_blockquote.scss
```diff
@@ -49,52 +51,61 @@ $_text-max-width: 65ch; // Approximately 80 characters with the current font
   }
 }
 
-.ecl-blockquote__attribution {
+.ecl-blockquote__attribution,
+%ecl-blockquote-author {
```
planctus: You didn't use a "BEM" syntax here to avoid the `percent-placeholder-pattern` error that you were getting..? I pushed a configuration of stylelint so that we don't need to disable the checks by our BEM syntax plugin in our components but it would also be possible to do a different thing: adapting the current checks by our plugin to also check the selectors with the placeholder for their "BEM" syntax and disable the `percent-placeholder-pattern`. This is how i initially did this, then i noticed that your selectors were not "bem" compliant, so i pushed  just an ignore for our plugin.

src/website/src/pages/ec/utilities/html-tag/docs/usage.md
```diff
@@ -0,0 +1,32 @@
+---
```
planctus, line 1: Two things, this page should be added also in `src/website/public/admin/config.yml` for the cms to handle it and since we have now the EU sections in the website we should also add one there `src/website/src/pages/eu/utilities/html-tag`, finally also this usage page in EU should be added in the cms configuration :)

src/website/src/pages/ec/utilities/html-tag/docs/usage.md
```diff
@@ -0,0 +1,32 @@
...
+title: Usage
+order: 1
+---
+
+import { Link, Paragraph } from '@ecl/website-components';
```
planctus: You created an md file, correctly for the cms to read it, but then we should not use the react components, the style of the paragraph will be respected by simply having a <p> tag, because it's the first, i don't think i see any link in the page, but in case it can be created in markdown syntax [label](link)

src/presets/ec/src/ec-default.scss
```diff
@@ -0,0 +1,141 @@
...
+@use "@ecl/vanilla-component-link/link-ec";
+@use "@ecl/vanilla-component-ordered-list/ordered-list";
+@use "@ecl/vanilla-component-table/table";
+@use "@ecl/vanilla-component-unordered-list/unordered-list";
+@use "@ecl/vanilla-utility-typography/typography";
+
+// stylelint-disable no-descending-specificity
+.ecl button:not([class*='ecl-']),
+.ecl button:is([class*='ecl-u-']) {
+  @extend %ecl-button;
```
planctus, line 16: I'm not really sure we need to creysate these placeholder selectors, i mean, their main idea is that they might be used to define some style that, if they are not extended, would not generate css, but here we are applying those selectors always in addition to some existing ones, so why not extending the one we have directly..?

planctus (changes requested):
I think it's pretty pretty nice, but i wonder why we need those additional placeholders selectors. i left a few comments and i've pushed a change in the stylelint configuration to let you use these selectors without disabling the bem checks, but let's see if we want to change the approach about the placeholders, in case i will adapt this conf.

src/implementations/vanilla/components/blockquote/_blockquote.scss
emeryro: the reason why I use placeholders is because that's highly recommended to do so when using extend. It prevent unwanted side effects and potential duplications in compiled CSS.
see for instance:
- https://webinista.com/updates/dont-use-extend-sass/
- https://stackoverflow.com/questions/54084022/what-does-extend-must-be-used-with-a-placeholder-mean

planctus: Yeah, ok, fine for the placeholders, still i would use the same convention that we use for the classes, so BEM, if we decide so, i will make the needed changes in the eslint configuration for that not to make the lint check fail.

src/implementations/twig/utilities/default/index.story.js
emeryro: do we have an example somewhere, where we activate a css for a specific story?

src/implementations/vanilla/components/blockquote/_blockquote.scss
emeryro: the linter fails even with BEM conventions (it just doesn't like the % thingy...).
but indeed, I will update the placeholder names

emeryro: it seems like BEM patterns like '__' or '--' are not valid for placeholders by default.
We could override the linter configuration, but I guess it is cleaner to just follow the default rule here and keep current naming

planctus: but even normal selectors (classes) using BEM would not be "allowed" in the default configuration of stylelint we have, we do this to prevent the linter from complaining:     
```
// Allow underscores in class names (BEM)
    'selector-class-pattern': null,
```
and we have our own plugin to check the compliance of the BEM selectors we define.
So, i think for consistency, we can just treat the selectors the same way, whether they are classes or placeholders

planctus: So, if we agree about this i will push the configuration for including the placeholders in the same check we do for the BEM syntax of our classes and i will disable the rule that checks the placeholders in stylelint.

src/implementations/twig/utilities/default/index.story.js
planctus: Surely we have it in OEL. you have to override the cssresources in the parameters, basically..
https://github.com/planctus/oe-library/blob/master/src/components/header/header.story.js

src/website/src/pages/ec/utilities/html-tag/docs/usage.md
planctus: I am thinking that we will handle this page on our side, so i'd say that the inclusion in the cms is only optional, no real need for that...

planctus (dismissed):
To me it's good, just that there is a small typo "recommanded" in the usage pages and that we still have to select the right css manually in the storybook example.

src/website/src/pages/ec/utilities/html-tag/docs/usage.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Usage
+order: 1
+---
+
+<p>
+  ECL provides an easy way to style some default HTML tags (like links or paragraphs).<br />
+  Note that this approach comes with limitations (no possible variant for elements, limited number of HTML tags, ...), and should be considered as an edge case. The recommanded way to use ECL is still to apply the corresponding CSS classes where needed.
```
emeryro:
```suggestion
  Note that this approach comes with limitations (no possible variant for elements, limited number of HTML tags, ...), and should be considered as an edge case. The recommended way to use ECL is still to apply the corresponding CSS classes where needed.
```

PR #2082, opened 2021-05-26
feat(gallery): Adding EU styles, revamping EC - front-2255

planctus:
REMARKS:

- The size of the previous and next icon of the overlay in the "specs" is L, but visually is clearly smaller, it has been set to "S"
- The styles of the previous and next buttons in the overlay are not specified, the specs request to follow the button's default but there is no such a default, the hover color has been set to "yellow-25"
- Styles for the close button of the overlay are not specified, specs say to follow the default for buttons but there is no such a default in EC, the hover color has been set to yellow-25
- The color for the caption in the overlay in the specs is "grey-140" but it's clearly lighter in the mocks, it has been set to "grey-60"
- There is an unspecified spacing under the description in the overlay for the mobile viewport in EC and EU, a min-height has been set on the description container has been set to reproduce this.
- A small spacing has been added in the left of the view_all button in order for the hover and focus styles in EU to be correct
-  There is no mention of the ecl-gallery__footer element (link to further media items) in the specs, the element has been maintained as it was, except for a spacing that has been removed in order to maintain a consistent style.


src/implementations/vanilla/components/gallery/_gallery.scss
```diff
@@ -101,15 +108,19 @@ $_description-color: map.get(theme.$color, 'white-100');
...
+  color: $_gallery-info-color;
   float: right;
-  font: map.get(theme.$font, 'xs');
-  padding: map.get(theme.$spacing, 'xs') map.get(theme.$spacing, 's');
+  font: $_gallery-info-font;
+  padding: map.get(theme.$spacing, 'm') map.get(theme.$spacing, 's');
+}
+
+/* stylelint-disable-next-line plugin/selector-bem-pattern */
+[data-ecl-gallery-all] {
```
emeryro: we should use a proper css class here instead of targeting the data-* attribute.
If there is none on the element, that's the good time to add one

src/implementations/vanilla/components/gallery/gallery-ec.scss
```diff
@@ -0,0 +1,23 @@
+@use "sass:map";
+@use '@ecl/theme-dev/theme';
+@use 'gallery' with (
+  $_actions-margin: map.get(theme.$spacing, 'xs'),
+  $_actions-margin-desktop: map.get(theme.$spacing, 'm'),
+  $_arrow-margin: map.get(theme.$spacing, 'm'),
+  $_arrow-margins: map.get(theme.$spacing, 'xs') map.get(theme.$spacing, '2xs'),
+  $_close-button-color: map.get(theme.$color, 'white-100'),
+  $_description-background: map.get(theme.$color, 'grey-100'),
+  $_description-icon-small: 1.5rem,
```
emeryro: can't we use the icon sizes defined in the scss map?

src/implementations/vanilla/components/gallery/gallery-ec.scss
```diff
@@ -0,0 +1,23 @@
...
+  $_description-margin-desktop: map.get(theme.$spacing, 'm'),
+  $_detail-background: map.get(theme.$color, 'grey-100'),
+  $_detail-color: map.get(theme.$color, 'white-100'),
+  $_gallery-info-color: map.get(theme.$color, 'grey'),
+  $_gallery-info-font: map.get(theme.$font, 'xs'),
+  $_gallery-meta-color: map.get(theme.$color, 'white-100'),
+  $_icon-small: 1rem,
```
emeryro: same here

emeryro (changes requested):
A few little things

emeryro:
Thanks for the update @planctus 
Can you just add a few lines in the migration doc for the most impacting updates in markup and data structure? (mostly the new class and the icon update)? Then we'll be good

emeryro (dismissed):

papegaill (commented):
Is it normal: 
- that the overlay is not fully fullscreen, there is a space at the bottom EC and EU?
- to have the text in yellow on the close, next and previous buttons of the overlay in hover? same on EU for next/prev btns?
- to have a yellow focus border for gallery items in EU? it should follow the general blue focus right?

planctus:
For the first point yes, these styles were added in the implementation although missing in the specs, if you read the remarks in this pr designers were alerted about this and they approved those styles, they will update the specs basing on them.
For the other two point i think we need to fix those things, should be fairly easy, i will do it, wait before merging this, then.. ;)

PR #2084, opened 2021-05-28
feat(stories): Set padding by default for all stories and make some fullscreen - FRONT-2211

papegaill:
I opted to make all the stories padded by default, and to manually render certain stories with the fullscreen layout option:
https://storybook.js.org/docs/react/configure/story-layout


src/implementations/twig/components/site-header-standardised/site-header-standardised.story.js
```diff
@@ -143,18 +143,27 @@ export const Default = (args) =>
 
 Default.storyName = 'default';
 Default.argTypes = getArgTypes(englishData);
-Default.parameters = { notes: { markdown: notes, json: englishData } };
```
planctus: but if all the stories in a component have to be displayed "fullscreen" while not adding the parameter in the `export default` instead of repeating it for each story..?

papegaill: done

src/playground/ec/.storybook/preview.js
```diff
@@ -40,7 +40,7 @@ addParameters({
     },
   ],
   controls: { expanded: true },
-  layout: 'fullscreen',
+  layout: 'padded',
```
emeryro, line 43: we should also apply this to EU storybook

emeryro (changes requested):
that's fine for me, just a small thing to fix

src/playground/ec/.storybook/preview.js
papegaill: Oups, thank you 👍

emeryro (dismissed):

PR #2089, opened 2021-06-01
feat(language-list): Revamping EC and EU - FRONT-2250

papegaill:
Previews:

- EC: https://ecl-preview-2089--europa-component-library.netlify.app/playground/ec/?path=/story/components-language-list--splash
- EU: https://ecl-preview-2089--europa-component-library.netlify.app/playground/eu/?path=/story/components-language-list--splash

Specs:

- EC: https://share.goabstract.com/34b7ec51-0c20-46d0-954a-6670620e49cd
- EU: https://share.goabstract.com/7a0bbfd6-d343-4258-a27b-b7713734610c


planctus (commented):
For me is ok, i just have two small remarks:

1. In the readme file the title is still defined as a string, but:
2. If the newly added icon has to me of size `m` in both systems, i think we should hard-code that in the template, also the icon used doesn't seem to be "optional", so also it's name could/should be hard-coded in the template and then i wonder: we already have the` icon_path` parameter so couldn't we simply render the icon without asking anything additional and at that point leave the `title` as it was..?
3. If so, then also the logic in the template could change a bit and the icon rendered in any case, even when the `title` is not set

papegaill:
> For me is ok, i just have two small remarks:
> 
> 1. In the readme file the title is still defined as a string, but:
> 2. If the newly added icon has to me of size `m` in both systems, i think we should hard-code that in the template, also the icon used doesn't seem to be "optional", so also it's name could/should be hard-coded in the template and then i wonder: we already have the` icon_path` parameter so couldn't we simply render the icon without asking anything additional and at that point leave the `title` as it was..?
> 3. If so, then also the logic in the template could change a bit and the icon rendered in any case, even when the `title` is not set

Thank you @planctus! You are absolutely right, I updated the PR accordingly 👍

PR #2096, opened 2021-06-11
fix(icons): Icons not displaying on footer standardised - FRONT-2328

tibi2303:
Which is the correct version of icons? I included the latest sprite (icon-media-social)
Also we have a problem here, as the links has color white, the icons get filled white but linkedin and facebook are not displaying properly still the twitter icon needs to get filled white in order to be shown. In my opinion, we need another set of icons for fb and linkedin.

### DIFF
--- a/src/playground/addons/story-utils/index.js
+++ b/src/playground/addons/story-utils/index.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-param-reassign */
 import iconPathEc from '@ecl/resources-ec-icons/dist/sprites/icons.svg';
 import iconPathEu from '@ecl/resources-eu-icons/dist/sprites/icons.svg';
+import iconSocialPath from '@ecl/resources-ec-social-icons/dist/sprites/icons-social.svg';
 import iconMediaSocialPath from '@ecl/resources-social-media-icons/dist/sprites/icons-social-media.svg';
 import getSystem from '@ecl/builder/utils/getSystem';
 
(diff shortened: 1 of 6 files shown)


emeryro:
you should use the negative version of the icons, as it is on a dark background.
You can see the icons here: https://v3--europa-component-library.netlify.app/ec/guidelines/iconography/

src/playground/addons/story-utils/index.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable no-param-reassign */
 import iconPathEc from '@ecl/resources-ec-icons/dist/sprites/icons.svg';
 import iconPathEu from '@ecl/resources-eu-icons/dist/sprites/icons.svg';
-import iconSocialPath from '@ecl/resources-ec-social-icons/dist/sprites/icons-social.svg';
+import iconMediaSocialPath from '@ecl/resources-social-media-icons/dist/sprites/icons-social-media.svg';
```
emeryro, line 5: icons from `@ecl/resources-ec-social-icons` are still used in some components (namely social media follow and social media share).
So we should have both resources available: footer uses icons from `@ecl/resources-social-media-icons`, the other two from `@ecl/resources-ec-social-icons`

emeryro (changes requested):

src/playground/addons/story-utils/index.js
tibi2303: I added all 3 cases in this function (based on social-media, social and fallback cases).

emeryro (approved):

PR #2127, opened 2021-06-22
feat(select-multiple): ECL v3 - Dispatching change event on user selection - FRONT-2268

planctus:


emeryro:
That seems good to me.
Do you think that you could document this feature somewhere, with a small example as the one you provided in the v2 PR?
Ideally in the API, worse case in the showcase page I think.

planctus:
Thanks @emeryro for reviewing this, to be honest i don't think there is anything to document, this not affecting the ecl apis, basically it's just recreating via "code" something that would normally happen in the DOM, so it's kind of totally transparent for the user. If anyone wants to listen for a change in the select multiple values, he would now be able to do it while before no event was propagated after the user selection.

PR #2139, opened 2021-06-30
feat(typography): update EU typography

emeryro:
Update website page and headings styling for EU

specs: https://citnet.tech.ec.europa.eu/CITnet/confluence/pages/viewpage.action?pageId=1023169001


planctus (commented):
It seems all good to me, there is just a small difference in the "specs", which is:
`The exception to this rule are headings`, the thing is that the same text appears also in the EC guildelines, so i don't know, if we have to change it here it should change also there...
The other thing i noticed is that from the "usage" page in EU all the links to the guidelines are pointing to EC, maybe we can fix this in this pr.. ;)

emeryro:
I fixed every fault link that I found in usage pages
Also, I created https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-2424 to update "getting started" pages

planctus (approved):

PR #2172, opened 2021-08-05
fix(cards): fix center display - FRONT-2708

emeryro:
fix alignment of image on EC cards


src/implementations/vanilla/components/card/_card.scss
```diff
@@ -56,7 +56,7 @@ $_list-padding: null !default;
   margin-left: $_image-margin;
   margin-right: $_image-margin;
   margin-top: $_image-margin;
-  width: 100%;
+  width: calc(100% - (2 * #{$_image-margin}));
```
planctus, line 59: Ok, this works, but i wonder the reason why this is not handled as a padding on the container element, wouldn't it be easier that way..?

planctus (approved):

PR #2175, opened 2021-08-13
fix(playground): Fixing optional css loading in EC and EU - FRONT-2792

planctus:


emeryro:
It works fine on netlify, but on production only the print css checkbox, in CSS resources,  seems to do something (at least the 404 errors are fixed indeed).
Do you have the same behavior on your side?

planctus:
@emeryro, wow, yes, it's a total mess...but it's not because of these recent deployments, i've checked in many "archived" versions and still the behavior of css resources is kind of erratic, the main css cannot be disabled while the print works as expected.
If this was not the case before the migration to the asset manager then it's something related to that, otherwise we have been living with this issue for a very long time, i don't know...

planctus:
I write here like to take notes:
the misbehaviour seems to be indeed the presence of two <links> to the same css, but it seems to only occur in the archived packages.
The guess would then be that this is due to the function and the if that we use to identify the "open canvas in new tab" use case to be able to load the css because cssresources would not, when in an archived package (subfolder), i've checked in both instances, archived and production, the code we have seems to work consistently...

There is an additional confusion due to the fact that in v2 we load in the "open canvas in new tab" a different set of css resources from the ones we have loaded in the iframe, in one case is the website-legacy preset while in the second is the website preset.
In any case the i

emeryro (approved):

PR #2180, opened 2021-08-18
feat(page-header): add overlay for background image - FRONT-2886

emeryro:
Add overlay (light and dark) for core and standardised page headers (EC/EU)

### DIFF
--- a/src/implementations/twig/components/page-header-standardised/__snapshots__/page-header-standardised.test.js.snap
+++ b/src/implementations/twig/components/page-header-standardised/__snapshots__/page-header-standardised.test.js.snap
@@ -3,7 +3,7 @@
 exports[`Page Header Standardised Background image - renders correctly 1`] = `
 <jest>
   <div
-    class="ecl-page-header-standardised ecl-page-header-standardised--image"
+    class="ecl-page-header-standardised ecl-page-header-standardised--image ecl-page-header-standardised--overlay-none"
   >
     <div
       aria-hidden="true"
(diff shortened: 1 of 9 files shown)


src/specs/components/page-header-standardised/demo/data--background-image.js
```diff
@@ -9,4 +9,5 @@ module.exports = {
   },
   background_image_url:
     'https://inno-ecl.s3.amazonaws.com/media/examples/example-image.jpg',
+  overlay: 'none',
```
planctus: mmh..."none" is just a value that we add in the control's select to reset it, it doesn't really exists, indeed in the "docs" comment of the tempalte we only mention "dark" and "light".
If we don't want to use any overlay in the example of the standardised page header we should simply not pass the parameter and in the story file we should use  'none' as the defaultValue of the control.

src/implementations/twig/components/page-header-standardised/__snapshots__/page-header-standardised.test.js.snap
```diff
@@ -3,7 +3,7 @@
 exports[`Page Header Standardised Background image - renders correctly 1`] = `
 <jest>
   <div
-    class="ecl-page-header-standardised ecl-page-header-standardised--image"
+    class="ecl-page-header-standardised ecl-page-header-standardised--image ecl-page-header-standardised--overlay-none"
```
planctus: Yep, this is the consequence of the fact that we pass the 'none' value for the overlay parameter..

src/implementations/twig/components/page-header-standardised/page-header-standardised.story.js
```diff
@@ -95,6 +92,21 @@ const getArgTypes = (data) => {
...
       },
     };
+
+    argTypes.overlay = {
+      name: 'image overlay',
+      type: 'select',
+      defaultValue: data.overlay,
+      description: 'Overlay on top on background image',
+      table: {
+        type: { summary: 'string' },
```
planctus, line 102: we miss the defaultValue for the "table", which is an empty string.

planctus (changes requested):
it works fine, but we introduced a small issue...

planctus (approved):

PR #2196, opened 2021-09-08
feat(site-header): add optional CTA - FRONT-2514

emeryro:
Add optional call-to-action link to standardised site header
This is done in two phases:
- in the menu, for mobile display
- in the site header for desktop


src/implementations/twig/components/menu/menu.story.js
```diff
@@ -34,12 +40,31 @@ const getArgTypes = () => {
         type: 'text',
       },
     },
+    cta_link: {
+      name: 'call to action',
+      type: { name: 'boolean' },
+      defaultValue: false,
```
planctus: this `defaultValue` is not needed, we now set it via the `getArgs()` which you did, so you can remove it here .. ;)

src/implementations/twig/components/site-header-standardised/site-header-standardised.story.js
```diff
@@ -86,6 +91,20 @@ const getArgTypes = () => {
       category: 'Content',
     },
   };
+  argTypes.cta_link = {
+    name: 'call to action',
+    type: { name: 'boolean' },
+    defaultValue: false,
```
planctus: Same as before, the object we build in `getArgs()` is replacing the `defaultValue` for each control as we were used to set them.

planctus (changes requested):
Two lines to be removed, for the rest for me is fine.. ;)

PR #2197, opened 2021-09-08
feat(quote-image): added image in blockquote - FRONT-2463

tibi2303:


src/implementations/twig/components/blockquote/README.md
```diff
@@ -14,13 +14,19 @@ npm install --save @ecl/twig-component-blockquote
 - **"extra_attributes"** (optional) (array) (default: []) Extra attributes
   - "name" (string) Attribute name, eg. 'data-test'
   - "value" (string) Attribute value, eg: 'data-test-1'
+- **"image"** (optional) (object) (default: {}) Blockquote image
```
planctus: This comment is unclear, it should come earlier than extra_attributes and it shouldn't have "attribute name" and "attribute value" which were probably copied from the previous one.

src/implementations/twig/components/blockquote/blockquote.html.twig
```diff
@@ -39,8 +44,23 @@
 {# Print the result #}
 
 <blockquote class="{{ _css_class }}"{{ _extra_attributes|raw }}>
```
planctus: i don't know, i think it's a bit questionable what are doing with this markup, the blockquote tag should be used to "wrap" the quote, here we build a layout inside of it, i think i would prefer having a `<figure>` as the component root and the `<blockquote>` tag only wrapping the real quote. At this point the selector  `ecl-blockquote__body` could be used instead of the newly added selector `ecl-blockquote__inner `and the `<blockquote>` could be defined as an element like `ecl-blockquote__quote`, but i'd let you choose if this would be a more rational definition for this component or not, i understand it might a bit out of the scope of this "extension" of the component look&feel

src/implementations/twig/components/blockquote/blockquote.html.twig
```diff
@@ -39,8 +44,23 @@
...
+  {% if image is defined and image is not empty %}
+    <img
+      {% if image.path is not empty %}
+        src="{{ image.path }}"
+      {% endif %}
+      {% if image.alt is not empty %}
+        alt="{{ image.alt }}"
+      {% endif %}
+        class="ecl-blockquote__image"
+        width="120"
```
planctus: interesting way of fixing the size of this image, hardcoding it here, but i think we should do this in the css more than in the template.

src/implementations/twig/components/blockquote/blockquote.story.js
```diff
@@ -35,6 +35,17 @@ const getArgTypes = (data) => {
       type: 'text',
     },
   };
+  argTypes.image = {
```
planctus: i think this control should be a boolean, letting you show/hide the image and it should be set to false by default, we are adding the possibility of using an image in the component, not necessarily as its "default" implementation

src/implementations/vanilla/components/blockquote/_blockquote.scss
```diff
@@ -109,4 +131,12 @@ $_text-max-width: 65ch; // Approximately 80 characters with the current font
     font: $_author-font-desktop;
     font-weight: $_author-font-weight;
   }
+
+  .ecl-blockquote__image,
+  %ecl-blockquote__image {
+    float: left;
```
planctus: Ok, it works this way floating the element, but i wonder if we wouldn't prefer using a flex layout for the component instead, with `row-reverse` as the flex-direction and `justify-content: left`.

src/specs/components/blockquote/demo/data.js
```diff
@@ -3,4 +3,7 @@ module.exports = {
   citation:
     'An interconnected grid will help deliver the ultimate goal of the Energy Union, to make sure affordable, secure and sustainable energy, and also growth across the EU.',
   author: 'President Juncker',
+  image: {
+    path: 'https://inno-ecl.s3.amazonaws.com/media/examples/example-image3.jpg',
```
planctus: don't be have another example image to use for this..? :) something with a european flag or so..?

planctus (changes requested):
The changes i really think we should make are minor ones, just the Readme text and the control used in the demo,  the others are kind of "optional", for you to evaluate them. ;)

src/implementations/twig/components/blockquote/blockquote.html.twig
tibi2303: This is a bit hard to implement, as i need an inner container that should have those borders left and on responsive the border should change to the outer container, also i'm not sure about a figure tag containing all the markup. Maybe i did not understood corectly

src/specs/components/blockquote/demo/data.js
tibi2303: I need a 1:1 ratio or around there, i will try to find something else

src/implementations/twig/components/blockquote/blockquote.html.twig
planctus: what i'm saying is just that instead of wrapping everything into a <blockquote> tag this tagcould be only used to wrap the quote, i see that we need some markup to handle the look and feel, the rest of the proposal is indeed only about what selector should identify what, not changing the current wrappers you have.
So,  the ` __body` element could be what currently is the `__inner` element, and the `<blockquote>` could be identified by a new element `__quote`.
The resulting structure would be:
```
<figure class="ecl-blockquote>
  <div class="ecl-blockquote__body>
    <blockquote class="ecl-blockquote__quote"><p></p></blockquote>
    <footer class="ecl-blockquote__attribution></footer>
  </div>
  <img class="ecl-blockquote__image />
</figure>
```

src/specs/components/blockquote/demo/data.js
planctus: i like this image better.. :)

src/implementations/twig/components/blockquote/blockquote.html.twig
tibi2303: Should be done.

src/specs/components/blockquote/demo/data.js
emeryro: fyi, we have a 1:1 image available: https://inno-ecl.s3.amazonaws.com/media/examples/example-image-square.jpg

src/implementations/twig/components/blockquote/blockquote.story.js
planctus: ok, good, but can you give the control a "human readable" label..basically without the underscore.. :)

src/specs/components/blockquote/demo/data.js
planctus: even better ;)

planctus:
i don't normally check the styles, i leave that to the review by the designers but i've noticed that we add a big margin in mobile under the image, basically you should remove its margin-bottom, cause it adds to the padding of the wrapper

src/implementations/twig/components/blockquote/blockquote.story.js
```diff
@@ -40,10 +41,28 @@ const getArgTypes = () => {
         type: 'text',
       },
     },
+    show_image: {
+      name: 'show_image',
+      type: { name: 'boolean' },
+      defaultValue: false,
```
planctus: i see you added the show_image default value correctly in the args object, this one should be removed, then.

src/implementations/twig/components/blockquote/blockquote.story.js
```diff
@@ -40,10 +41,28 @@ const getArgTypes = () => {
...
+        category: 'Content',
+      },
+    },
   };
 };
 
 const prepareData = (data, args) => {
+  if (args.show_image) {
+    data.image = {
+      path: 'https://inno-ecl.s3.amazonaws.com/media/examples/example-image-square.jpg',
```
planctus, line 61: well, like this if we change the image again the specs you also need to update it here.
You can clone this value from the data object into a constant that then you can use here, instead.

src/implementations/twig/components/blockquote/blockquote.story.js
```diff
@@ -40,10 +41,28 @@ const getArgTypes = () => {
...
+    };
+  } else {
```
planctus: what about a ternary operator instead of this if..? :)

src/implementations/twig/components/blockquote/blockquote.story.js
```diff
@@ -40,10 +41,28 @@ const getArgTypes = () => {
...
       },
     },
+    show_image: {
+      name: 'show_image',
+      type: { name: 'boolean' },
+      defaultValue: false,
+      description: 'Show image',
+      table: {
+        type: { summary: 'boolean' },
+        defaultValue: { summary: 'false' },
```
planctus: also this small thing, this is not a parameter in the template, so please don't set any default value in the table which might look confusing...

src/implementations/twig/components/blockquote/blockquote.story.js
planctus: Maybe you changed something, but not in the right way, what should be "readable" is the "name" of the control, so the label that appears in storybook.
The description for this parameter should be 
`"Toggle image visibility"`

planctus (approved):

PR #2198, opened 2021-09-08
feat(storybook): Refactoring stories (select options and defaultValue) - FRONT-2722

planctus:


emeryro (changes requested):
there seems to be an issue with icons, they are not displayed in storybook

planctus:
in which story did you notice this @emeryro ..? I've checked a few ones, starting from the "icons" components in the netlify preview and i see them correctly..

emeryro:
well, I reopened the stories, and now it seems to work.. maybe an issue on my side...

emeryro (approved):

PR #2209, opened 2021-09-17
fix(datepicker): Fix display issues on mobile - FRONT-3174

papegaill:


emeryro (commented):
It's already looking way better, thanks.
While we're at it, I noticed that there is a width issue on the datepicker, on mobile
image
It's already the case in v2 too. Do you think that's something we can fix easily? If not we should have another ticket to fix this later

papegaill:
> It's already looking way better, thanks.
> While we're at it, I noticed that there is a width issue on the datepicker, on mobile
> image
> It's already the case in v2 too. Do you think that's something we can fix easily? If not we should have another ticket to fix this later

@emeryro   JS fix added to fix picker size that exceeds vw on mobile.

emeryro (approved):

PR #2210, opened 2021-09-20
fix(default-css): fix conflict with webtools - FRONT-3171

emeryro:


papegaill (approved):
Approved to fix this particular bug on the map. but we are not immune to other conflicts. As discussed, we must see directly with webtools team so that they generalize the naming of elements of these type in all their widgets .

PR #2211, opened 2021-09-21
fix(date-picker): Styling and JS adaptations - FRONT-3183

papegaill:
Previews:
- EC: https://ecl-preview-2211--europa-component-library.netlify.app/playground/ec/?path=/story/components-forms-datepicker--default
- EU: https://ecl-preview-2211--europa-component-library.netlify.app/playground/eu/?path=/story/components-forms-datepicker--default


src/specs/components/checkbox/demo/data--eu.js
```diff
@@ -9,7 +9,8 @@ module.exports = {
   name: 'checkbox-default',
   invalid_icon: {
     path: '/icons.svg',
-    name: 'warning',
+    type: 'notifications',
```
emeryro: If I'm not wrong, we don't need to have this 'type' field anymore, all icons are in the same category now (same for the other data files)

emeryro (changes requested):
almost good, just a small thing:

src/specs/components/checkbox/demo/data--eu.js
papegaill: cool thanks for the info, as i saw some specified i wanted to align them all up ;) i removed them everywhere..

emeryro (approved):

PR #2222, opened 2021-10-04
feat(footers): Adding EC logo to all the footers - FRONT-3224

planctus:


src/implementations/twig/components/footer-core/__snapshots__/footer-core.test.js.snap
```diff
@@ -17,6 +17,18 @@ exports[`Footer Core EC default renders correctly 1`] = `
           <div
             class="ecl-footer-core__section"
           >
+            <a
```
papegaill, line 20: Strange focus state actually on the logo:

Screenshot 2021-10-08 at 14 33 40

already better by having `display:block` to the `href`

Screenshot 2021-10-08 at 14 34 03

I think it would be better to set the bottom margin on the parent `href` instead of the logo `img`so the focus will be better around the logo only. as made on standardised?

src/implementations/twig/components/footer-core/footer-core.story.js
```diff
@@ -27,7 +29,12 @@ const prepareData = (data) => {
 export default {
```
papegaill, line 29: Any control to show or hide the logo? 

Here is ticket description:
_"EU footers already have the logo. For EC, it must appear by default, but there should be an option to hide it."_

Same for standardised and harmonised, EU also?

src/implementations/twig/components/footer-standardised/__snapshots__/footer-standardised.test.js.snap
```diff
@@ -300,6 +300,18 @@ exports[`Footer Standardised EC default renders correctly 1`] = `
           <div
             class="ecl-footer-standardised__section"
           >
+            <a
```
papegaill, line 303: On the mocks it seems that for desktop they want first the heading "European Commission" then the logo. And reverse on mobile. maybe flex order? 

Same for harmonised.

papegaill (changes requested):

src/implementations/twig/components/footer-standardised/__snapshots__/footer-standardised.test.js.snap
planctus: are you looking on abstract..? The "mocks" have been updated after a request from us to align the behaviour of this section, the decision was to always have the logo on top, not sure the files have been updated also there, maybe it was done only on figma: https://www.figma.com/file/MZCXQtdJmpx8gzEvYlvwYn/EC-Components?node-id=170%3A3719

src/implementations/twig/components/footer-core/footer-core.story.js
planctus: I've added a control to hide the logo in harmonised and standardised, not in core where no other "section" can be hidden.

src/implementations/twig/components/footer-core/__snapshots__/footer-core.test.js.snap
planctus: You're right, the element was inline and that's why the margin was applied to the image, now this behaves as the others.

src/implementations/twig/components/footer-standardised/__snapshots__/footer-standardised.test.js.snap
papegaill: yes looking on abstract ;)

papegaill (approved):

PR #2226, opened 2021-10-08
fix(site-headers): Using only one site_name in site-headers - FRONT-3227

planctus:
PROBLEM:

-  in order to embed the menu into the site-header and support the "design" in the different viewport, the solution we adopted was to duplicate the site name, one in the site header and one in the menu, we used to show the first only in desktop, the second only in mobile.
This created different issues in the different "family", but although it could have been working properly simply passing the site name to the menu from the site-header template, at the same time this involved lots of duplicaiton, the menu was having a "site name" and a "cta" link that didn't belong to it, for instance.

SOLUTION:

- The solution is to use only one site name, as a parameter of the site header and ensuring the correct positioning of the menu when in mobile via css, the menu is absoluted positioned in mobile aligning this with the site header banner, holding the site name.


src/implementations/twig/components/menu/README.md
```diff
@@ -31,36 +30,36 @@ npm install --save @ecl/twig-component-menu
...
-      label: "Menu item", 
-      path: "example", 
-      is_current: false, 
-      children: [ 
+{% include '@ecl/menu/menu.html.twig' with {
+  title: 'Menu',
+  close: 'Close',
+  back: 'Back',
+  icon_path: '/icons.svg',
+  site_name: 'Site name',
```
emeryro:
```suggestion

```

emeryro (changes requested):
that's good for me, much cleaner that way!
just a minor thing to fix

emeryro (dismissed):

emeryro (approved):

PR #2232, opened 2021-10-18
feat(card, file): add highlight label - FRONT-3276

emeryro:


src/implementations/twig/components/card/__snapshots__/card.test.js.snap
```diff
@@ -129,6 +151,28 @@ exports[`Card Default renders correctly with extra attributes 1`] = `
...
+            class="ecl-label ecl-label--highlight ecl-card__label"
+          >
+            highlighted
+          </span>
+        </li>
+        <li
+          class="ecl-card__label-item"
+        >
+          <span
+            class="ecl-label ecl-label--medium ecl-card__label"
```
papegaill, line 170: on playground this label is highlight, blue color? normal? https://ecl-preview-2232--europa-component-library.netlify.app/playground/ec/?path=/story/components-card--card

papegaill: for me as an example it is better to use the "high important" variant for this label, as in figma `ecl-label--high`

papegaill (changes requested):

src/implementations/twig/components/file/__snapshots__/file.test.js.snap
```diff
@@ -16,6 +16,15 @@ exports[`File Thumbnail renders correctly 1`] = `
...
         >
+          <div
+            class="ecl-file__label"
+          >
+            <span
+              class="ecl-label ecl-label--highlight"
+            >
+              highlighted
+            </span>
+          </div>
```
papegaill, line 27: for the file, should we also have a yellow “high important” label as in figma?

src/implementations/twig/components/card/__snapshots__/card.test.js.snap
emeryro: yes I noticed this, but I didn't want to over-complicate the controls. They only offers the possibility to change the label, so the color is fixed on the example

src/implementations/twig/components/file/__snapshots__/file.test.js.snap
emeryro: I used this wiki page as a reference: https://citnet.tech.ec.europa.eu/CITnet/confluence/pages/viewpage.action?pageId=948443705
But in any case, the twig file expect a label component structure, so it is a matter of data to use one variant or another

PR #2252, opened 2021-11-04
feat(file): label(s) also in the default variant - FRONT-3350

planctus:


emeryro (dismissed):
Fine for me, we should just get the validation from the designer

emeryro (approved):

PR #2253, opened 2021-11-09
feat(menu): allows more menu items - FRONT-3354

emeryro:
Changed the way menu items are displayed in the mega menu; we no longer use fixed grid, but rely on the column-count property.

This PR also include accessibility improvement requested on https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-3347


papegaill (approved):
Nice 👍  keyboard tested and it works fine with focus.

PR #2255, opened 2021-11-15
feat(spinner): Adding component - FRONT-3223

planctus:
This is a twig implementation of a "spinner" component.
It is not clear how to make this usable in an application, 
1. if we have specific components to apply this to we could include this template, then the visibility of the spinner can be handled through the addition of a class.
2. Otherwise this could be used as a "template" only for a functionality to be implemented at the application level
3. We could potentially set the markup through js so that given a selector we could inject the spinner markup, but it doesn't look like an elegant nor a failsafe approach.

UPDATE:
After talking with Francesco it seems that this will ony be used as a example markup that needs to be copied into a drupal function, so not using any twig template.
Therefore the current approach seems potentially correct.


src/implementations/twig/components/spinner/spinner.html.twig
```diff
@@ -0,0 +1,82 @@
+{% spaceless %}
+{#
+  Parameters:
+  - "variant" (string) options:
+    [ primary, negative ] (default: primary)
+  - "size" (string) options:
+    [ small, medium, large ]
+  - "text" (string) (default: '')
+  - "visible" (boolean) (default false)
```
emeryro, line 10: missing "centered" and "overlay" parameters

src/implementations/twig/components/spinner/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Parameters:
+
+- **"variant"** (string) (default: 'primary')
+  options: [ primary, negative ]
+- **"size"** (string) (default: 'medium')
+  options: [ small, medium, large ]
+- **"text"** (string) (default: '')
+- **"visible"** (boolean) (default: false)
+- **"centered"** (boolean) (default: true))
```
emeryro, line 17: missing "overlay" parameter

src/implementations/twig/components/spinner/spinner.html.twig
```diff
@@ -0,0 +1,82 @@
...
+{% set _extra_attributes = '' %}
+
+{% set _css_classes = _css_classes ~ ' ' ~ 'ecl-spinner--' ~ _variant  ~ ' ecl-spinner--' ~ _size %}
+
+{% if _centered %}
+  {% set _css_classes = _css_classes ~ ' ' ~ 'ecl-spinner--centered' %}
+{% endif %}
+
+{% if _visible %}
+ {% set _css_classes = _css_classes ~ ' ecl-spinner--active' %}
```
emeryro: as the parameter is called "visible", maybe we could use the same name for the class (".ecl-spinner--visible"). Same for the overlay

src/implementations/twig/components/spinner/spinner.story.js
```diff
@@ -0,0 +1,114 @@
...
+    text: {
+      type: { name: 'string' },
+      description: 'Text below the loading indicator',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '' },
+        category: 'Content',
+      },
+    },
+    variant: {
```
emeryro, line 32: when selecting negative spinner on storybook (in the controls), we should apply a background (similar to what has been done for the links), otherwise it is not visible (in primary story)
Or we just don't show this control

src/implementations/vanilla/components/spinner/_spinner.scss
```diff
@@ -0,0 +1,122 @@
...
+    width: map.get($_spinner-sizes, 'small');
+  }
+
+  .ecl-spinner__text {
+    margin-left: map.get(theme.$spacing, 'xs');
+    margin-top: 0;
+  }
+}
+// stylelint-disable-next-line plugin/selector-bem-pattern
+.path {
```
emeryro: can't we apply the namespace/bem pattern for this css class too?

emeryro (changes requested):
General remarks:
- the order of pages on the website is not correct
image
- the overlay in storybook is not visible for primary. Maybe we could have some dummy content somewhere to see the difference

emeryro (dismissed):

emeryro (approved):

PR #2271, opened 2021-11-30
fix(news-ticker): Fix showcase iframe resize issue - FRONT-3425

papegaill:
The `heightCalculationMethod` option for the showcase iframes is causing the issues.
Using `max` value instead of `lowestElement` solve the issues.
see: http://davidjbradshaw.github.io/iframe-resizer/


emeryro (changes requested):
this update indeed fixes the news ticker issue. But it introduced a new bug on the site-header/menu showcase
menu
It is also slightly visible when closing the multi select

papegaill:
> this update indeed fixes the news ticker issue. But it introduced a new bug on the site-header/menu showcase menu It is also slightly visible when closing the multi select

Thank you for spotting this! 
I'm now applying the `max` heightCalulationMethod only to the news-ticker showcase iframe.

emeryro (approved):

PR #2281, opened 2021-12-08
fix(site-header): Mobile display of site-header harmonised without site_name - FRONT-3497

papegaill:
- Fix displays of Group1 and Group2 without site_name.
- This also fixes the menu display for the Group2 variant. Currently, the menu is not displayed on mobile.


emeryro (changes requested):
it is good for EC site headers, but there is still an issue on EU harmonised/standardised
image

papegaill:
> it is good for EC site headers, but there is still an issue on EU harmonised/standardised image

Thanks for the review @emeryro PR updated accordingly.
 I also added the fix for FRONT-3508

emeryro (approved):

PR #2289, opened 2021-12-17
fix(file): Review component spacings - FRONT-3424

papegaill:


emeryro (commented):
it seems that designers changed the separators in this component too.
this is not linked to the spacing, but I assume that they will report it

emeryro (approved):

PR #2292, opened 2021-12-20
fix(site-header): Correctly rename the cta class in the standardised variant - FRONT-3519

papegaill:
CSS assets limit increased by 1kb


emeryro (changes requested):
to avoid breaking changes, I suggest to keep also the previous class name in the css, and mark it as deprecated.
We could remove it when we do the revamp of the component next year

emeryro (approved):

PR #2297, opened 2022-01-03
chore(deps): Updates 03-01-22

planctus:


emeryro (changes requested):
I noticed a small issue:
- font family not applied on some links, in both website and components (i assume that is it some auto formatting issue, there are some extra ```"```)
image
image

planctus:
There were issues introduced by recent prs, the syntax for the select controls changed in storybook 6.4, options is an array and in case we need the labels to be different than the values we need to pass a labels object inside the control property of the argType, basically...

emeryro (approved):

PR #2310, opened 2022-01-13
chore(deps): Major updates, using node 14 - v2

planctus:


emeryro (approved):
As much as I can see, it seems to work :)

PR #2328, opened 2022-02-02
feat(file-upload): Add dynamic upload file list example - FRONT-3525

papegaill:
Composition previews: 

- EC: https://ecl-preview-2328--europa-component-library.netlify.app/playground/ec/?path=/story/compositions-file-upload-status--default
- EU: https://ecl-preview-2328--europa-component-library.netlify.app/playground/eu/?path=/story/compositions-file-upload-status--default


emeryro (commented):
that's a tricky task here...
On one side, putting the html in the story helps keeping the template clean, on the other side we still add some specific css which will not be used in most cases. So we are kind of in-between two positions here.

I would say that if we provide the css, we could also provide the structured html coming with it (so for instance having already the "ecl-file-upload__item-action" section, even if empty)
Or we don't provide the css either (relying on utilities)

papegaill:
> that's a tricky task here... On one side, putting the html in the story helps keeping the template clean, on the other side we still add some specific css which will not be used in most cases. So we are kind of in-between two positions here.
> 
> I would say that if we provide the css, we could also provide the structured html coming with it (so for instance having already the "ecl-file-upload__item-action" section, even if empty) Or we don't provide the css either (relying on utilities)

Many thanks for the review @emeryro ! 
You are totally right, I reviewed the markup using only utilities classes and I removed the extra css 👍

emeryro (dismissed):

src/implementations/twig/compositions/file-upload-status/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@ecl/twig-composition-file-upload-status",
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "3.2.2",
```
emeryro:
```suggestion
  "version": "3.2.3",
```

src/implementations/twig/compositions/file-upload-status/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "description": "ECL File upload status",
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@ecl/twig-component-spinner": "^3.2.2",
```
emeryro:
```suggestion
    "@ecl/twig-component-spinner": "^3.2.3",
```

src/implementations/twig/compositions/file-upload-status/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "@ecl/twig-component-link": "^3.2.2"
```
emeryro:
```suggestion
    "@ecl/twig-component-link": "^3.2.3"
```

emeryro (commented):
just the version number to update and that's fine

emeryro (approved):

PR #2333, opened 2022-02-10
feat(site-footer): add component - FRONT-3541

papegaill:
Previews: 

- EC: https://ecl-preview-2333--europa-component-library.netlify.app/playground/ec/?path=/story/components-site-wide-site-footer--core
- EU: https://ecl-preview-2333--europa-component-library.netlify.app/playground/eu/?path=/story/components-site-wide-site-footer--core


src/implementations/twig/components/site-footer/ec-site-footer.story.js
```diff
@@ -0,0 +1,140 @@
...
+import dataCore from '@ecl/specs-component-site-footer/demo/data-core--ec';
+import dataStandardised from '@ecl/specs-component-site-footer/demo/data-standardised--ec';
+import dataHarmonised from '@ecl/specs-component-site-footer/demo/data-harmonised--ec';
+import logoEc from '@ecl/resources-ec-logo/negative/logo-ec--en.svg';
+import footer from './site-footer.html.twig';
+import notes from './README.md';
+
+const getArgs = () => {
+  const args = {
+    hide_contact: true,
```
emeryro: in the other component we name the switcher the other way ("show_contact" instead of "hide_contact")

src/specs/components/site-footer/demo/data-harmonised--ec.js
```diff
@@ -0,0 +1,246 @@
...
+              },
+            },
+          ],
+        },
+      ],
+    ],
+    [
+      [
+        {
+          links: [
```
emeryro, line 140: I didn't check the specs, but I guess that this section is optional (there is no control to disable it)

src/implementations/twig/components/site-footer/eu-site-footer.story.js
```diff
@@ -0,0 +1,33 @@
...
+  decorators: [withCode, withNotes],
+  parameters: {
+    layout: 'fullscreen',
+    controls: {
+      disable: true,
+    },
+  },
+};
+
+export const Core = () => footer(prepareData(dataCore));
```
emeryro: for EU there are no other footer than the Core? No Harmonised?

src/implementations/twig/components/site-footer/README.md
```diff
@@ -0,0 +1,199 @@
+# ECL Site Footer component
```
emeryro, line 1: this component is named "site footer", and not "footer" as previously. Is this a request from COMM?
no real issue for me, but I just wanted to make sure that it was on purpose

src/implementations/twig/components/site-footer/site-footer.html.twig
```diff
@@ -0,0 +1,166 @@
+{% spaceless %}
+
+{#
+  Parameters:
+  - "variant" (string) (default: '') core, standardised or harmonised
```
emeryro: do we still need this "variant" parameter (and corresponding css classes)?

emeryro (changes requested):

src/implementations/twig/components/site-footer/ec-site-footer.story.js
papegaill: What component are you referring to? I didn't really touch the controls, I kept the same logic, but here all the controls are in hide mode? would you like me to put them all in show mode?

src/implementations/twig/components/site-footer/README.md
papegaill: Yes the new component should be named **site-footer**, see ECL-460

src/implementations/twig/components/site-footer/eu-site-footer.story.js
papegaill: Yes, for me there is only Core for the EU version, see description of ECL-460

src/implementations/twig/components/site-footer/site-footer.html.twig
papegaill: I actually wanted to see this point with you, a good approach would be a “negative” boolean param?
instead of having the `—core`, `—standardised` or even `—harmonised` classes, we would just have a `—negative` class. this could also apply for the new **page-header** (https://github.com/ec-europa/europa-component-library/pull/2337) as well as the **breadcrumb**.

src/specs/components/site-footer/demo/data-harmonised--ec.js
papegaill: for the actual standardised version, this section is mandatory, so I guess we should keep it like this. https://ec.europa.eu/component-library/ec/standardised-template/footer/usage/

src/implementations/twig/components/site-footer/ec-site-footer.story.js
emeryro: Yes, we didn't update the control of headers/footer when doing the global cleanup a few weeks ago, as we knew that they would be modified soon.
You can check for instance the content item (https://github.com/ec-europa/europa-component-library/blob/v3-dev/src/implementations/twig/components/content-item/content-item.story.js) or the card (https://github.com/ec-europa/europa-component-library/blob/v3-dev/src/implementations/twig/components/card/card.story.js)

src/specs/components/site-footer/demo/data-harmonised--ec.js
emeryro: for the harmonised I don't think that it is mandatory. To be confirmed with designers I guess

src/implementations/twig/components/site-footer/site-footer.html.twig
emeryro: yes I think that it would be a good approach. That way the design does not depend on the family, but the family just use one design or another

papegaill: as discussed, I confused all the footers have the blue background, so there is no negative version to differentiate them.. the difference is in the grid which is different for the core version. So for now I added a param `split_columns` that generates a `--split-columns` class modifier only used for core EC.

src/implementations/twig/components/site-footer/eu-site-footer.story.js
papegaill: I added the EU harmonised version

src/specs/components/site-footer/demo/data-harmonised--ec.js
papegaill: new control `show_class_names` added

src/implementations/twig/components/site-footer/ec-site-footer.story.js
papegaill: done

emeryro (commented):
this is good for me
please just update the version of packages, as we released 3.2.3

emeryro (dismissed):

emeryro (approved):

PR #2337, opened 2022-02-15
feat(page-header): add site-wide components - FRONT-3546

papegaill:
- EC: https://ecl-preview-2337--europa-component-library.netlify.app/playground/ec/?path=/story/components-site-wide-page-headers--core
- EU: https://ecl-preview-2337--europa-component-library.netlify.app/playground/eu/?path=/story/components-site-wide-page-headers--core

PR #2347, opened 2022-02-21
feat(breadcrumb): add site-wide components - FRONT-3552

papegaill:
ideally the following PR https://github.com/ec-europa/europa-component-library/pull/2337 should be merged first in order to review the include of the new page-header according to this new breadcrumb negative variant: https://github.com/ec-europa/europa-component-library/blob/9261c6ebe9de7b9a63d25aa988c6c3abf65de6de/src/implementations/twig/components/page-header/page-header.html.twig#L76-L84

EC: https://ecl-preview-2347--europa-component-library.netlify.app/playground/ec/?path=/story/components-site-wide-navigation-breadcrumb--core
EU: https://ecl-preview-2347--europa-component-library.netlify.app/playground/eu/?path=/story/components-site-wide-navigation-breadcrumb--core


src/implementations/twig/components/breadcrumb/breadcrumb.story.js
```diff
@@ -0,0 +1,93 @@
...
+Core.args = getArgs(dataCore);
+Core.argTypes = getArgTypes(dataCore);
+Core.parameters = {
+  notes: { markdown: notes, json: dataCore },
+};
+
+export const Standardised = (args) =>
+  breadcrumb(prepareData(dataDefault, args));
+
+Standardised.storyName = 'standardised';
```
emeryro: there is no page header standardised for EU, so I guess that it is the same for breadcrumb

src/implementations/twig/components/breadcrumb/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@ecl/twig-component-breadcrumb",
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "3.2.3",
```
emeryro: version should be 3.2.4 everywhere

src/implementations/twig/components/breadcrumb/breadcrumb.html.twig
```diff
@@ -0,0 +1,145 @@
...
+  },
+  extra_classes: 'ecl-breadcrumb__icon',
+  extra_attributes: [{ name: 'role', value: 'presentation' }]
+} %}
+
+{% set _css_class = 'ecl-breadcrumb' %}
+
+{# Internal logic - Process properties #}
+
+{% if _variant is not empty and _variant == 'negative' %}
```
emeryro, line 60: do we really need to test if the variant is "negative"? we could just append to the class name, that's what is done in most components

src/implementations/vanilla/components/breadcrumb/breadcrumb-print.scss
```diff
@@ -0,0 +1,8 @@
+/*
+ * Breadcrumbs Standardised
+ * @define breadcrumb-standardised
+ */
+
+.ecl-breadcrumb-standardised {
```
emeryro: wrong copy/paste I guess, this should not be standardised here

src/website/src/pages/ec/components/site-wide/navigation/breadcrumb/docs/api.mdx
```diff
@@ -0,0 +1,43 @@
+---
+title: API
+order: 3
+---
+
+import BreadcrumbJSDoc from '../../../../../_jsdoc/breadcrumbcore.md';
```
emeryro: should not be breadcrumb core I guess

emeryro (changes requested):
There is something wrong with the website structure now, with two entries for the breadcrumb
image
To be checked with COMM (this does not block the PR)

papegaill:
> There is something wrong with the website structure now, with two entries for the breadcrumb image To be checked with COMM (this does not block the PR)

Indeed, for now, I have removed the navigation folder under site-wide, I think it will  avoid confusion.

emeryro (changes requested):
The standardised example should be removed from the website (EU)
image

emeryro (approved):

PR #2350, opened 2022-03-02
fix(form): fix checkbox and radio display on multiple lines - FRONT-3562

emeryro:
Fix display of checkboxes and radios when the label takes multiple lines (it didn't wrap correctly)
I also ensured that the fix we did for https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-3190 was still working.
I had to add an extra span in the label to achieve this, so markup is slightly altered.
Here is the result
image

### DIFF
--- a/src/implementations/twig/components/checkbox/__snapshots__/checkbox.test.js.snap
+++ b/src/implementations/twig/components/checkbox/__snapshots__/checkbox.test.js.snap
@@ -145,6 +145,52 @@ exports[`Checkbox EC default renders correctly 1`] = `
         Helper text for an option
       </div>
     </div>
+    <div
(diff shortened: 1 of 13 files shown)


papegaill (commented):
I'm not convinced about the vertical alignment of the checkbox which is centered when there are multiple lignes. Isn't it better to align it at the start with a slight margin-top of `0.125rem`?

I tested by removing all the flex properties from the container and removing the white-space property, just using a `float: left` on the checkbox, without altering the markup, and simplifying the css:

Screenshot 2022-03-07 at 12 57 36

To see it may not be ideal in terms of label alignment, I think we should ask COMM for confirmation?

PR #2358, opened 2022-03-15
feat(rtl): right-to-left batch 1 - FRONT-3565

emeryro:
- card
- content-item
- list-illustration
- inpage-navigation
- link
- menu
- spinner
- tabs
- tag

Notes
- there are a few elements that will be handled in another batch (for instance description list). So display may not be fully operational yet (would need a final review when all batch are done)
- the javascript for tabs has to be updated to fix mobile display. For now I focused on the css, we will handle the few tricky components after


src/implementations/vanilla/components/tabs/tabs.js
```diff
@@ -190,7 +190,7 @@ export class Tabs {
       newOffset = maxScroll;
     }
     this.list.style.transitionDuration = '0.4s';
-    this.list.style.transform = `translate3d(-${newOffset}px, 0px, 0px)`;
+    this.list.style.transform = `translate3d(${newOffset}px, 0px, 0px)`;
```
papegaill: this breaks the navigation for the non-rtl tabs in mobile, the pagination is in reverse, I won't apply this change now?

papegaill (changes requested):
Really nice, good result already with only little extra css!
I just noticed 2 small things:
Screenshot 2022-03-15 at 18 18 19
For the inpage, active background/border should be reversed too, but it may already be too much detail..
Screenshot 2022-03-15 at 18 20 32
for tabs, the spacing of the icon on the more button should be corrected, to see if this will be taken into account in another batch with the js fixes?

src/implementations/vanilla/components/tabs/tabs.js
emeryro: yes, I did it for testing and forgot to remove... I'll update

emeryro:
I updated the inpage nav, to have something working well in both cases, and reverted the wrong js update.
For the button, it is indeed handled in another PR

PR #2359, opened 2022-03-16
feat(rtl): right-to-left batch 2 - FRONT-3566

emeryro:
- carousel
- featured-item
- gallery
- hero-banner
- icon
- media-container
- page-banner
- social-media-follow
- social-media-share

Notes:
- there are a few elements that will be handled in another batch (for instance links and button). So display may not be fully operational yet (would need a final review when all batch are done)
- carousel has to be be fixed in js, at a later stage

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -22,7 +22,7 @@
   {
     "path": "dist/packages/eu/styles/ecl-eu.css",
     "webpack": false,
-    "limit": "33 KB"
+    "limit": "34 KB"
   },
   {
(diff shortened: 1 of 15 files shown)


papegaill (changes requested):
I only see a spacing issue on the social links
Screenshot 2022-03-17 at 11 41 20

emeryro:
I merged with the first batch containing the link, it should be good for social share now

papegaill (approved):
just noticed an issue with EU gallery modal background, support ticket created as a follluw-up

PR #2369, opened 2022-03-24
fix(storybook): Adding static folder for rtl css - FRONT-3575

planctus:


emeryro (approved):
thanks for this, that would be really helpful when working on the rtl css

PR #2372, opened 2022-03-28
feat(select): Correctly reset the multiselect on form reset - FRONT-3583

papegaill:
Not so obvious to fix. To achieve this, I use the existing event `handleClickOutside` by identifying if the clicked element is a button of the `reset` type.

To test PR locally, simply wrap `select.html.twig` code with basic form tag: 

```
<form action="#">
  ....
  <input type="reset" value="Reset">
  <input type="submit" value="Submit">
</form>
```


emeryro:
this solution should work indeed, but can't you achieve this with the onreset event?
https://www.w3schools.com/jsref/event_onreset.asp

papegaill:
> this solution should work indeed, but can't you achieve this with the onreset event? https://www.w3schools.com/jsref/event_onreset.asp

Thank you, @emeryro. Yes, it could also work, but I wanted to avoid having listeners on other elements such as forms or having to intervene on all the forms that will be present on the page. In the same way, it will be necessary to identify if the multi-select is indeed part of the reset forms. What do you think?

planctus:
Not sure i get your concerns alexis, it seems that we would simply look for a parent element "form" and apply the listener to that, not more than this and you are already doing it from another "point of view".. :)
`const form = e.target.closest('form');`

So it seems pretty straightforward to me to listen for a reset event on the form.

src/implementations/vanilla/components/select/select.js
```diff
@@ -524,6 +531,29 @@ export class Select {
...
     }
   }
+
+  /**
+   * Reset Multiselect.
+   */
+  resetForm() {
+    // A slight timeout is necessary to execute the function just after the original reset of the form.
+    setTimeout(() => {
+      Array.from(this.select.options).forEach((option) => {
```
planctus, line 545: but.. i wonder, shouldn't actually be an autoInit().update operation the one that we should perform on form reset..? Or destroy and init, probably better..? 
Currently we have an issue:
image
but this made me think that a "reset" event on a form might actually be one use case for a destroy method for our components.. :)

papegaill: What's the issue? 
it's normal that Bulgaria is selected after reset, as it is since the beginning? It reset the form element to the default states, and Bulgaria is selected by default.

planctus: the issue is that the label Bulgaria disappears from the select when the form has been reset, like if there is no option selected

papegaill: Strange it works just fine locally here, can you ensure JS is fully recompile?

planctus (approved):

PR #2375, opened 2022-03-30
fix(accessibility): Pagination, carousel, gallery and  news-ticker - FRONT-3591

papegaill:
Fix accessibility issues for the following EC & EU components: 

**Pagination**
https://ec.europa.eu/component-library/playground/ec/?path=/story/components-navigation-pagination--default
- Elements must only use allowed ARIA attributes

**Carousel**
https://ec.europa.eu/component-library/playground/ec/?path=/story/components-carousel--default
- ARIA hidden element must not contain focusable elements

**Media container**
https://ec.europa.eu/component-library/playground/ec/?path=/story/components-media-container--embedded-video
- Frames should be tested with axe-core

**News Ticker**
https://ec.europa.eu/component-library/playground/ec/?path=/story/components-news-ticker--default
- ARIA hidden element must not contain focusable elements
- Buttons must have discernible text

**Button**
- New `hide_label` parameter added for screen reader only


src/implementations/twig/components/news-ticker/news-ticker.html.twig
```diff
@@ -11,6 +11,9 @@
         ...
       ]
     - "counter_label" (string) (default: 'of')
+    - "sr_previous" (string) (default: 'Previous news') screen reader label for previous button
+    - "sr_next" (string) (default: 'Next ,ews') screen reader label for next button
```
emeryro:
```suggestion
    - "sr_next" (string) (default: 'Next news') screen reader label for next button
```

src/implementations/twig/components/news-ticker/news-ticker.html.twig
```diff
@@ -93,56 +102,60 @@
...
-      },
-      extra_classes: 'ecl-news-ticker__prev',
-      extra_attributes: [
-        {name: 'data-ecl-news-ticker-prev'},
-        {name: 'aria-controls', value: _aria_controls},
-        {name: 'tabindex', value: '-1'},
-        {name: 'data-controls', value: 'prev'},
-      ],
-    } only %}
+    <button
```
emeryro: Isn't it worth adding a parameter to the button template, to have a label, but hidden?
Something like hide_label: true/false

emeryro (approved):
That's good for me.
Could you just add in the PR description that we added this new "hide_label" parameter?

PR #2379, opened 2022-04-01
feat(tabs): Adapt JS for RTL support - FRONT-3594

papegaill:
I needed to review the way to remove listeners on specific elements like so: 
https://bobbyhadz.com/blog/javascript-remove-all-event-listeners-from-element


emeryro (dismissed):
It seems to work well. I would just request a second review by @planctus as it would impact the way we handle all javascript components

planctus (approved):

PR #2384, opened 2022-04-05
feat(ext-links): Adding external icon to links based on a parameter -  FRONT-3601

planctus:
The approach used is to give priority to the external icon:
it means that if the link is set to be external it will take precedence on any other "icon" that might have been passed to the component.
This can be tested by setting an icon through the controls, if the link is external nothing should change.


src/specs/components/card/demo/data--card.js
```diff
@@ -12,6 +12,8 @@ module.exports = {
       label: 'Better regulation',
       path: exampleLink,
       level: 1,
+      external: true,
```
emeryro: for the demo to COMM we can have the external link by default, but I guess that on the end we should have the normal link here (and maybe a control to have it external)

src/specs/components/content-item/demo/data--image.js
```diff
@@ -16,6 +16,8 @@ module.exports = {
     type: 'standalone',
     label: 'Title',
     path: exampleLink,
+    icon_path: 'icons.svg',
+    external: true,
```
emeryro: same remark for the default external link

src/specs/components/menu/demo/data--en.js
```diff
@@ -36,25 +36,7 @@ module.exports = {
...
-        { label: 'Item 4.7', path: exampleLink },
-        { label: 'Item 4.8', path: exampleLink },
-        { label: 'Item 4.9', path: exampleLink },
-        { label: 'Item 4.10', path: exampleLink },
-        { label: 'Item 4.11', path: exampleLink },
-        { label: 'Item 4.12', path: exampleLink },
-        { label: 'Item 4.13', path: exampleLink },
-        { label: 'Item 4.14', path: exampleLink },
-      ],
+      external: true,
```
emeryro: we should update the translated data file too

emeryro (changes requested):

planctus:
Thanks Romain for reviewing this, the examples were meant to show where the "functionality" was added to give a chance both to comm and platform to see what the result would have been. I agree that if this gets "validated" by our stakeholders we can simply remove the icon in some of the stories where we are currently showing it, i would also keep the control to activate it only for the link stories.

planctus:
It still not entirely clear how are we going to release this, the pr has been kept against v3, the examples have been removed, the link is the main "entry" point to test the "external icon" using the associated control

planctus:
The external icon has been added also to the tag component, only when the tag is of type "link"

emeryro (approved):

PR #2394, opened 2022-04-14
feat(rtl): batch 6 - FRONT-3570

papegaill:


emeryro (changes requested):
There are probably more utilities which needs update here.
Could you check all utilities using "left" or "right" somewhere? (media and float for instance)

papegaill:
> There are probably more utilities which needs update here. Could you check all utilities using "left" or "right" somewhere? (media and float for instance)

Thanks for the review @emeryro, I checked, but I don't see what to change, float right or left should not change. And for media, it's background alignment. Nothing to change as well. Maybe I'm missing something?

emeryro (approved):

PR #2396, opened 2022-04-21
feat(datepicker): Adapt JS for RTL support - FRONT-3595

papegaill:


src/playground/ec/.storybook/preview-head.html
```diff
@@ -32,6 +32,16 @@
     window.frameElement.id === 'storybook-preview-iframe'
   ) {
     // Let CSS Resources inject the styles
+
+    // Re-initialize component from the parent frame when changing #ecl-rtl CSS resource checkbox
+    window.parent.document.addEventListener('click', (e) => {
+      if (e.target && e.target.getAttribute('id') === 'ecl-rtl') {
+        console.log(e.target.checked);
```
planctus: forgot this line..

src/playground/ec/.storybook/preview-head.html
```diff
@@ -47,6 +57,7 @@
 <script>
   // https://github.com/storybookjs/storybook/issues/6113#issuecomment-473965255
   function runOnPageChange() {
+    ECL.autoInit().destroy();
```
planctus, line 59: Do we really need to destroy everything on each click on in a different story..? I don't necessarily see the need for this, but i might not consider something..

src/tools/dom-utils/autoinit/index.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable no-console */
+/* eslint-disable no-plusplus */
```
planctus: Why adding an exclusion and not changing the syntax to prevent the error..? What was wrong with the previous forEach, by the way..?

src/playground/ec/.storybook/preview-head.html
papegaill: Yes  first of all to lighten the global `ECL.components` array which grows with each component instantiation. And also because I noticed that by switching rtl and navigate between the stories, if the destroy is not done it creates problems..

src/tools/dom-utils/autoinit/index.js
papegaill: The  previous foreach was wrong, removing elements from the array during the loop misleads the final result. 
it's therefore necessary to loop backwards to avoid any breaking..

PR #2397, opened 2022-04-21
feat(content-item): revamp content item and description list - FRONT-3554

emeryro:
We tried to provide backward compatibility for most case, but there may be some potential breaking changes, especially if you are using taxonomy lists

Main changes
- new building block called content block
- content item now relies on the content block instead of card, :warning: so data structure has been updated
- description list now offers new options (standalone links), and :warning: the taxonomy no longer uses tags

Notes
- content block component is only displayed in storybook. To be seen after review if we should hide it or not
- basic backward compatibility has been added for content item meta and image


src/implementations/twig/components/content-block/content-block.html.twig
```diff
@@ -0,0 +1,167 @@
...
+      {% for _meta in _primary_meta %}
+        <li class="ecl-content-block__primary-meta-item">
+          {{- _meta -}}
+        </li>
+      {% endfor %}
+    </ul>
+  {% endif %}
+
+  {% if _title is not empty and _title.label is not empty %}
+    <h1 class="ecl-content-block__title">
```
papegaill, line 81: I see that's already the case for the content-item, but are we sure we want an <h1> here, at the limit an h2, right?

papegaill, line 81: I see that's already the case for the content-item, but are we sure we want an h1 here, maybe an h2 or h3 is best?

src/implementations/twig/components/content-block/content-block.story.js
```diff
@@ -0,0 +1,208 @@
...
+    clone.title.label = args.title;
+  }
+  if (clone.description) {
+    clone.description = args.description;
+  }
+
+  return clone;
+};
+
+export default {
```
papegaill, line 198: Will the story be hidden? I assume it remains visible for validation? 
Do we plan to update before the merge?

papegaill (commented):
Nice work, we'll actually have to see the feedback for the breaking changes.
Also, some tests are failing.

src/implementations/twig/components/content-block/content-block.html.twig
emeryro: Yes an no :)
The idea is to place this content block into an `<article>`. In this case, having an H1 is the way to go (see https://webdesign.tutsplus.com/articles/the-truth-about-multiple-h1-tags-in-the-html5-era--webdesign-16824)
But that' true that the content item should use an `<article>` and not a `<div>`. I'll update that

src/implementations/twig/components/content-block/content-block.story.js
emeryro: Yes, I guess that we will hide it before the merge

papegaill (approved):

pnowak2:
image

looks like breaking change within semantic minor release

pnowak2:
[10:56] NOWAK Piotr Andrzej (DIGIT-EXT)
    shouldn't you avoid breaking changes in semantic versioning between minor releases ?
Modifié​[10:57] NOWAK Piotr Andrzej (DIGIT-EXT)
    for major one its fine, but minor, 3.2.3 to 3.3.0 seems very brutal, should be in 4.0 rather
Modifié​[11:01] NOWAK Piotr Andrzej (DIGIT-EXT)
    its not even marked as breaking change in changelog, our users will now have all cards, navigation list broken
​[11:03] GAILLARD Alexis (DIGIT-EXT)
    Indeed, I suggest you to see this with emery.romain who will be back next week. I think the intention was to avoid any breaking change as much as possible and provide backward compatibilty as described in his pull request.
​[11:04] NOWAK Piotr Andrzej (DIGIT-EXT)
    yes, but card looks broken, and all card content markup changed, so its huge breaking change
​[11:04] NOWAK Piotr Andrzej (DIGIT-EXT)
    anyways, have to deal with that, but id like to agree with you that its not normal to have such breaking changes within minor releases
​[11:05] NOWAK Piotr Andrzej (DIGIT-EXT)
    better approach would be like you guys did in the past with i.e. accordion, you proposed new markup and released ecl-accordion2, still having old accordion intact
(1 mentions J'aime)​[11:05] NOWAK Piotr Andrzej (DIGIT-EXT)
    this seems to be kind with users instead of breaking it not even mentioning it in changelog
​[11:05] NOWAK Piotr Andrzej (DIGIT-EXT)
    fyi @emeryro 
<https://teams.microsoft.com/l/message/19:0daae67a80124e6389c57f036eabf852@thread.tacv2/1658998607764?tenantId=b24c8b06-522c-46fe-9080-70926f8dddb1&amp;groupId=4383e406-0bd7-4524-8872-0e5cec88fa89&amp;parentMessageId=1658997142289&amp;teamName=GRP-Front Support&amp;channelName=ECL Support&amp;createdTime=1658998607764>

PR #2404, opened 2022-04-28
feat(banner): add coypright - FRONT-3616

emeryro:
Add optional copyright to banners and carousel


src/implementations/vanilla/components/hero-banner/_hero-banner.scss
```diff
@@ -36,6 +36,23 @@ $_font-weight: null !default;
   position: relative;
 }
 
+.ecl-hero-banner__copyright {
```
planctus: it's true that we only use this twice, but couldn't we make a mixin and include it where needed, instead of duplicating these rules..?

src/implementations/twig/components/page-banner/page-banner.story.js
```diff
@@ -117,7 +122,16 @@ const getArgTypes = (data) => {
       },
     },
   };
+
   if (data.image) {
+    argTypes.show_copyright = {
```
planctus: this control seems uneffective, you don't use it to control the display of the copyright.

src/implementations/twig/components/page-banner/page-banner.html.twig
```diff
@@ -58,6 +60,9 @@
 <section class="{{ _css_class }}"{{ _extra_attributes|raw }}>
 {% if _image is not empty and _variant in ['image', 'image-shade', 'image-gradient'] %}
   <div class="ecl-page-banner__image" style="background-image:url({{ _image }})"></div>
+{% endif %}
+{% if _copyright is not empty and _variant in ['image', 'image-shade', 'image-gradient'] %}
```
planctus: I wonder if this is the right approach. i mean that i would imagine a use case where the copyright is needed but no text would be associated to it, not sure if it's a valid use case but in this case we should probably have a boolean to control the display and a text parameter to set an optional text, but you know what the requirements are better than me..

planctus (changes requested):

src/implementations/vanilla/components/hero-banner/_hero-banner.scss
emeryro: we are already duplicating all the other rules for the banners.
We could have some generic banner css, and just change what correspond to page or hero banner. But I would keep this for another ticket

planctus (approved):

PR #2469, opened 2022-05-12
feat(cards): refactor cards with content block - FRONT-3606

emeryro:
Update card component, to rely on the new content block element

**Important**:
Data structure has been updated/simplified (), but a temporary backward compatibility has been provided to avoid breaking changes.
Main modifications in data:
- the wrapper "card" object is no longer needed
- "meta" has been renamed to "primary_meta"
- "infos" has been renamed to "secondary_meta"
- "type" has been removed (no longer used)


src/implementations/twig/components/card/card.story.js
```diff
@@ -193,9 +207,23 @@ export default {
   decorators: [withCode, withNotes],
 };
 
-export const Card = (args) => card(prepareData(dataCard, args));
+export const Default = (args) => card(prepareData(dataCardDefault, args));
+
```
planctus, line 211: At this point i wonder about the reason why having multiple stories in card. I find that confusing considering that all the optional content of a card has toggles to show/hide them, i think that one single story "default" would be enough, but you probably based also on requests by comm, so just a personal thought..

planctus (dismissed):
It seems all good to me, i do have a concern regarding the markup, though, but it is coming from the content block, we have now these very long class names ex: ecl-content-block__secondary-meta-container and this code is going to be everywhere making the source code "heavier" although semantically correct.

planctus:
I don't understand how this would be backward compatible, though. I see that we keep populating a _card object with the properties of the card we receive from the data, for backward compatibility, but then we just include content block with the parameters in the root, so i don't see how an old implementation of a card would still work with this template.

planctus:
wouldn't it be good to have a test also with the "old" data to check the backward compatibility of this implementation..?

src/implementations/twig/components/card/card.story.js
emeryro: I won't say that this is a "request", but all their design mention 3 examples of cards (default, tags, list), so I guess that having them with the same naming would avoid some discussions

PR #2554, opened 2022-05-31
fix(forms): Add invalid icons for EC and fix them for EU - FRONT-3651

papegaill:


src/implementations/twig/components/file-upload/file-upload.html.twig
```diff
@@ -134,7 +136,13 @@
...
 {% if _invalid and _invalid_text is not empty %}
   <div class="ecl-feedback-message">
-    {%- block invalid_text _invalid_text -%}
+    {% if _invalid_icon is not empty %}
+      {% include '@ecl/icon/icon.html.twig' with {
+        icon: _invalid_icon,
+        extra_classes: 'ecl-feedback-message__icon',
+      } only %}
+    {% endif %}
+    {{- invalid_text -}}
```
emeryro: it should output a block here, same as what is done on the text field for instance

emeryro (changes requested):

src/implementations/twig/components/file-upload/file-upload.html.twig
papegaill: done, but initially I took this from the date picker where the parameter is not a block, should we create a ticket for date picker as well? 

https://github.com/ec-europa/europa-component-library/blob/43ed0a422e45a71f3462b20eae24325b7895d0f8/src/implementations/twig/components/datepicker/datepicker.html.twig#L106

Also, looking at the definition for the `optional_text` and `required_text` parameters in text-input or text-area, it seems that these are also blocks, but this is not the case in code, it may be necessary to review the definition of the blocks globally on all the forms component? What do you think?

https://github.com/ec-europa/europa-component-library/blob/43ed0a422e45a71f3462b20eae24325b7895d0f8/src/implementations/twig/components/text-input/text-input.html.twig#L22
https://github.com/ec-europa/europa-component-library/blob/43ed0a422e45a71f3462b20eae24325b7895d0f8/src/implementations/twig/components/text-input/text-input.html.twig#L23

emeryro: yes, let's just do it for the element now, we should align with platform to make sure that we provide what they expect

emeryro (approved):

PR #2561, opened 2022-06-07
feat(icons): add telegram icons - FRONT-3650

emeryro:


planctus:
Questions:
1) how does it come that we have a @ecl/resources-ec-social-icons package but no EU..? Maybe it's just a bad naming for the package since we use the same for EU..?
2) I noticed this because i was expecting these icons for telegram to be available in the "social" components, social-media-follow, social-media-share, but there we have two icons per social where the second is used on "hover", are we sure that this is going to provide the expected resources, then..?

emeryro:
@planctus there are two things here:
- the social media icons provided by ECL, available in the package `@ecl/resources-social-media-icons`, and visible here https://ecl-preview-2561--europa-component-library.netlify.app/ec/guidelines/iconography/
These icons are the same for EC and EU for now, but COMM intend to provide a different set for EU (to be seen later how we will handle this)
- the "legacy" social media icons, used in the social media components (share and follow), and available in the package `@ecl/resources-ec-social-icons`.
These icons have an "hover" version too. In the end, they may not stay in ECL if Webtools take over these two components

planctus (approved):

PR #2565, opened 2022-06-09
feat(site_header): Add aria label to language switcher - FRONT-3658

emeryro:


papegaill (approved):
breaking change tag?

emeryro:
@papegaill there should be no breaking change here, I just added an extra parameter.
I also fixed the documentation which was wrongly describing a "name" parameter, while it was in fact already called "label" in the code

papegaill (approved):

PR #2566, opened 2022-06-09
fix(carousel): Improve accessibility - FRONT-3657

papegaill:
- fix tab key flow + deactivate focus on hidden slides (`inert` + `tabindex='-1'` as a fallback)
- add play & pause buttons instead of only a toggle autoplay button
- pause carousel on mouseover
- disable autoplay if focus is on a slide CTA
- activate only touch events on the slides and remove mouse drag events to be able to select slide content
- if only one slide, hide controls and don't initialize carousel
- deactivate autoplay on mobile < 768
- update carousel styles:
  - review position of the controls for each breakpoint, 
  previous & next buttons are moved in the control bar for mobile format (JS)
  - controls hover, active and focus stats

todo: 
- testing for reflow with the ARC toolkit, the arrows of the carousel disappear


src/implementations/twig/components/carousel/carousel.html.twig
```diff
@@ -7,7 +7,8 @@
     - "sr_previous" (string) (default: 'Previous slides') screen reader label for previous button
     - "sr_next" (string) (default: 'Next slides') screen reader label for next button
     - "sr_navigation" (string) (default: 'Go to slide %d') screen reader label for navigation buttons
-    - "sr_autoplay" (string) (default: 'Carousel auto play') screen reader label for autoplay button
+    - "sr_play" (string) (default: 'Play carousel') screen reader label for the play button
```
emeryro, line 10: we could add a backward compatibility check, to use the value of the previous "sr_autoplay" by default
similar to what was done here https://github.com/ec-europa/europa-component-library/blob/v3.4.0-dev/src/implementations/twig/components/content-item/content-item.html.twig#L42

src/implementations/vanilla/components/carousel/_carousel.scss
```diff
@@ -85,71 +85,78 @@ $_outline-width: null !default;
...
-.ecl-carousel__next {
-  right: 0;
-
-  &:focus {
-    margin-inline-end: $_outline-width;
-    padding-inline-end: calc(
-      #{map.get(theme.$spacing, 's')} - #{$_outline-width}
-    );
-  }
+.ecl-carousel__autoplay {
```
emeryro, line 108: so user no longer has the possibility to play/pause carousel on mobile? Is it on purpose?

src/implementations/vanilla/components/carousel/_carousel.scss
```diff
@@ -225,6 +264,37 @@ $_outline-width: null !default;
...
-    display: block;
+    padding: map.get(theme.$spacing, 's');
+    position: absolute;
+    top: calc(50% - 22px);
+    transform: translateY(-50%);
+    z-index: 1;
+  }
+
+  .ecl-carousel__prev {
+    left: 0;
```
emeryro, line 275: I guess that it should be inverted in the rtl css

src/implementations/vanilla/components/carousel/_carousel.scss
```diff
@@ -225,6 +264,37 @@ $_outline-width: null !default;
 
   .ecl-carousel__prev,
```
emeryro, line 265: these buttons should be hidden on print

src/implementations/vanilla/components/carousel/_carousel.scss
```diff
@@ -85,71 +85,78 @@ $_outline-width: null !default;
...
-  right: 0;
-
-  &:focus {
-    margin-inline-end: $_outline-width;
-    padding-inline-end: calc(
-      #{map.get(theme.$spacing, 's')} - #{$_outline-width}
-    );
-  }
+.ecl-carousel__autoplay {
+  display: none;
```
emeryro, line 109: the buttons don't disapear at the the same breakpoint as the mobile version, so there are some cases where you have the mobile display, but still have the button

src/implementations/vanilla/components/carousel/carousel.js
```diff
@@ -31,14 +31,17 @@ export class Carousel {
   constructor(
     element,
     {
-      toggleSelector = '.ecl-carousel__toggle',
+      playSelector = '.ecl-carousel__play',
+      pauseSelector = '.ecl-carousel__pause',
```
emeryro, line 35: not sure why we use css class selector here instead of data attributes as on the other components, but that would require some quite big changes...
Probably for next major version of this component

emeryro (changes requested):
I'm not really sure that it wa what the accessibility expert had in mind when talking about "two buttons", but if that's validated by COMM...

src/implementations/vanilla/components/carousel/_carousel.scss
papegaill: Yes it is, here are the specs for touch screen: https://www.figma.com/file/wAuZXP2e9NUrsT5C1EKNEk/Carousel?node-id=428%3A32040

src/implementations/vanilla/components/carousel/_carousel.scss
papegaill: Already done in RTL CSS resource

src/implementations/vanilla/components/carousel/_carousel.scss
papegaill: Yes that's correct, https://www.figma.com/file/wAuZXP2e9NUrsT5C1EKNEk/Carousel?node-id=428%3A32040

src/implementations/vanilla/components/carousel/carousel.js
papegaill: I think I wanted to lighten the markup in a way... but indeed we'll have to align this as other components

src/implementations/vanilla/components/carousel/_carousel.scss
papegaill: They update the spec, so no autoPlay on mobile anymore 👍

PR #2568, opened 2022-06-13
fix(inpage-navigation): Improve accessibility - FRONT-3660

papegaill:


src/implementations/twig/components/inpage-navigation/inpage-navigation.html.twig
```diff
@@ -61,8 +63,7 @@
         extra_classes: 'ecl-icon--rotate-180 ecl-inpage-navigation__trigger-icon',
       } only %}
     </button>
-    <ul class="ecl-inpage-navigation__list" aria-labelledby="ecl-inpage-navigation-trigger"
-        data-ecl-inpage-navigation-list="true" id="ecl-inpage-navigation-list">
+    <ul class="ecl-inpage-navigation__list" data-ecl-inpage-navigation-list="true" id="ecl-inpage-navigation-list">
```
emeryro: we have an id here too, and another one for the trigger, which are hardcoded.
I would suggest to use the id you added in parameter in all cases, and add a suffix to it ("-title", "-list", "-trigger")

emeryro (changes requested):

emeryro (approved):

PR #2571, opened 2022-06-20
chore(links): Randomize links used in the stories - FRONT-3666

planctus:


src/implementations/twig/components/card/__snapshots__/card.test.js.snap
```diff
@@ -273,7 +273,7 @@ exports[`Card Default renders correctly 1`] = `
               Lorem ipsum dolor sit amet, 
               <a
                 class="ecl-link"
-                href="#"
+                href="/example"
```
emeryro, line 276: the random part is not applied here

src/implementations/twig/components/timeline/__snapshots__/timeline.test.js.snap
```diff
@@ -483,29 +483,13 @@ exports[`Timeline renders correctly with extra attributes 1`] = `
...
-          class="ecl-timeline__label"
-        >
-          28-29 June 2018
-        </div>
-        <div
-          class="ecl-timeline__content"
-        >
-          Commission contribution to the European Council: 
+          : European Commission's legislative proposals for spending programmes',
+    },
```
emeryro: there is some issue with the data here

emeryro (changes requested):
@planctus there are a few cases where the random part is not applied. I highlighted one of them, but it seems to be the case in other components too (see modified files)

src/implementations/twig/components/card/__snapshots__/card.test.js.snap
planctus: this is a snapshot, the alteration of the link only happens in storybook...

src/implementations/twig/components/timeline/__snapshots__/timeline.test.js.snap
planctus: Oh yeah, there was, but this wouldn't have worked in the previous implementation, i adapted ithe utility so that it will work also when the "/example" is part of a longer string and not just a prop value per se... ;)

emeryro (approved):

PR #2573, opened 2022-06-21
fix(default-css): Fix webtools highchart dropdown conflict - FRONT-3667

papegaill:


src/presets/ec/src/ec-default.scss
```diff
@@ -58,17 +58,17 @@
   @extend %ecl-paragraph;
 }
 
-.ecl ul:not([class*='ecl-'], [class*='wt-']),
+.ecl ul:not([class*='ecl-'], [class*='wt-'], [class*='highcharts-*']),
```
emeryro: I don't think we need the second '*'. The rule will already target any class starting with 'highcharts-'
```suggestion
.ecl ul:not([class*='ecl-'], [class*='wt-'], [class*='highcharts-']),
```

emeryro (commented):
That's quite a shame to be forced to add this kind of exceptions in ECL...
It's almost fine (just see comment below), but in the future we may see with webtools to ask them to add their namespace ('wt-') to these classes too

emeryro (approved):

PR #2576, opened 2022-06-27
fix(tabs): Improve accessibility - FRONT-3661

papegaill:


src/implementations/twig/components/tabs/tabs.html.twig
```diff
@@ -33,7 +33,8 @@
 {% set _css_class = 'ecl-tabs' %}
 {% set _extra_attributes = [
```
planctus, line 34: i see that we normally define our internal variable for attributes as a string that gets populated by looping into the extra_attributes param, and it's ok since it's going to be rendered as a string. But when we have "internal" attributes to set, it would feel natural to merge these values into the extra_attributes array and get the _extra_attributes string as the result of the internal and external attributes processing.

src/implementations/vanilla/components/tabs/tabs.js
```diff
@@ -334,6 +374,100 @@ export class Tabs {
...
+        flag = true;
+        break;
+
+      case 'End':
+        this.moveFocusToTab(this.lastTab);
+        flag = true;
+        break;
+
+      default:
+        break;
```
planctus: never seen a break on a default case for a switch, what would that be doing..? :)

src/implementations/vanilla/components/tabs/tabs.js
```diff
@@ -334,6 +374,100 @@ export class Tabs {
...
+    }
+
+    if (flag) {
+      e.stopPropagation();
```
planctus: not sure about what we would prevent here, but ok, just that for how the logic is it seems you could set the variable to true and then in the default case set it to false, not a big difference but less repetitions..

src/implementations/twig/components/tabs/tabs.html.twig
```diff
@@ -92,6 +104,7 @@
             transform: 'rotate-180'
           },
           extra_classes: 'ecl-tabs__toggle',
+          extra_attributes: [{ name: 'tabindex', value: '-1' }],
```
planctus, line 107: i understand that we are introducing navigation through keyboard arrows, but why disabling the standard navigation via Tab..? It feels really weird to access the first tab with a Tab and then being moved out of the tabs typing again Tab. This way the user really have to know that the only way of navigating our tabs would be using arrows.
Also because the standard navigation would be working decently in most of the viewports and only when some tabs get hidden it would be really problematic.

src/implementations/twig/components/tabs/tabs.html.twig
```diff
@@ -59,27 +60,38 @@
 {# Print the result #}
 
 {% if _items is not empty and _items is iterable %}
-<nav class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+<div class="{{ _css_class }}"{{ _extra_attributes|raw }}>
```
planctus: not sure the reason why of this change, but i guess there is.. :)

src/implementations/vanilla/components/tabs/tabs.js
```diff
@@ -334,6 +374,100 @@ export class Tabs {
...
+      }
+      return;
+    }
+
+    if (currentTab === this.lastTab) {
+      this.moveFocusToTab(this.firstTab);
+    } else {
+      this.moveFocusToTab(this.tabsKey[index + 1]);
+    }
+  }
```
planctus, line 447: i don't see the new listeners being destroyed together with the others in the destroy method..

src/implementations/vanilla/components/tabs/tabs.js
```diff
@@ -334,6 +374,100 @@ export class Tabs {
...
+    const tgt = e.currentTarget;
+    let flag = false;
+
+    switch (e.key) {
+      case 'ArrowLeft':
+        this.moveFocusToPreviousTab(tgt);
+        flag = true;
+        break;
+
+      case 'ArrowRight':
```
planctus, line 393: i thnk this navigation has a minor flaw, when you reach the "collapsed" tabs you would maybe expect to be able to browse the items also using arrowDown and arrowUp, no..? At least this is what i instinctively tried to do when i first tested this.

src/implementations/vanilla/components/tabs/tabs.js
```diff
@@ -334,6 +374,100 @@ export class Tabs {
...
+      this.dropdown.classList.remove('ecl-tabs__dropdown--show');
+    }
+    currentTab.focus();
+    return this;
+  }
+
+  /**
+   * @param {HTMLElement} currentTab tab element
+   */
+  moveFocusToPreviousTab(currentTab) {
```
planctus: Do we really need separated callbacks foran operation like moving focus between the tags..? It's not only because of the almost duplicated code, but it's about the logic, i feel that we could handle this in a single function, but it's just a suggestion and a preference..

src/implementations/vanilla/components/tabs/tabs.js
```diff
@@ -334,6 +374,100 @@ export class Tabs {
...
+   * @param {HTMLElement} currentTab tab element
+   */
+  moveFocusToTab(currentTab) {
+    if (currentTab.closest('.ecl-tabs__dropdown')) {
+      this.dropdown.classList.add('ecl-tabs__dropdown--show');
+    } else {
+      this.dropdown.classList.remove('ecl-tabs__dropdown--show');
+    }
+    currentTab.focus();
+    return this;
```
planctus: Why returning the instance here..?

planctus (changes requested):
A few comments, minor fixes and a few questions, but the navigation with arrows works fine, just that i don't understand why preventing users from using Tabs as well, it creates the impression of a "bug" and you need to discover yourself the alternative way of navigating between the different tabs.

src/implementations/twig/components/tabs/tabs.html.twig
papegaill: Not sure, I properly understand the issue here, we are doing same in expandable and news-ticker

src/implementations/vanilla/components/tabs/tabs.js
papegaill: :) This code was basically taken from https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html https://softwareengineering.stackexchange.com/questions/201777/break-on-default-case-in-switch

src/implementations/vanilla/components/tabs/tabs.js
papegaill: Sure you are right, I have removed this prevents condition 👍

src/implementations/twig/components/tabs/tabs.html.twig
papegaill: This is basically the request, to only be able to navigate with arrows and thus be able to go to the content more quickly without having to tab all tabs. https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html

src/implementations/twig/components/tabs/tabs.html.twig
papegaill: Was following markup from w3: https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html but I guess we can keep `<nav>`

src/implementations/vanilla/components/tabs/tabs.js
papegaill: done

src/implementations/vanilla/components/tabs/tabs.js
papegaill: I was first trying doing this, but it's more simple now the dropdown opens automatically if you try to go to the next tabs (collapse). Otherwise, you need to tab open the dropdown navigate with up and down arrows and tab again to come back to previous tabs. this was agreed with COMM.

src/implementations/vanilla/components/tabs/tabs.js
papegaill: removed 👍

src/implementations/vanilla/components/tabs/tabs.js
papegaill: Merged the 2 functions

src/implementations/twig/components/tabs/tabs.html.twig
planctus: not an issue, just that we are requesting extra attributes as objects with name and value for then processing them to generate a string. if we are about to create internal attributes i would merge these with the extra_attributes array in order to process them in one place consistently. I know that we have other use cases like this, but that doesn't make it right. it was a suggestion, if you like it like this, keep it that way.

src/implementations/vanilla/components/tabs/tabs.js
planctus: i don't absolutely agree with the user who wrote that comment, default is always executed at the end of the switch, breaking after it is completely irrelevant, i'd say logically wrong, for how i see it. But again, fine if you like it this way.

src/implementations/twig/components/tabs/tabs.html.twig
planctus: i don't understand, really..

src/implementations/vanilla/components/tabs/tabs.js
planctus: i was only sugggesting to "add" the possibility of browsing the dropdown also with up and down arrows, nothing wrong with the fact that it opens and close using left and right arrows. Actually what i would find correct would be to focus on the "more" element and open the dropdown for then using arrow up and down inside of it.

planctus (dismissed):

src/implementations/twig/components/tabs/tabs.html.twig
papegaill: We can discuss this on the next weekly if you want ;)

src/implementations/vanilla/components/tabs/tabs.js
papegaill: removed ;)

PR #2577, opened 2022-06-29
feat(social icons): add legacy icons for Mastodon and Telegram

emeryro:


planctus:
These svgs are not really consistent, some have width and height defined some they don't, then for the way we use them is not an issue and these icons are really just for the sake of this component...

planctus (approved):

PR #2582, opened 2022-07-04
feat(list-illustrations): Reviewing spacing - FRONT-3681

planctus:


emeryro (commented):
the icon size is different in our implementation (32 for us, 40 in the specs). It makes the alignment of the title a little odd.
I guess that we should update the size on our side...

emeryro (approved):

PR #2585, opened 2022-07-06
feat(social-media-follow): Revamp - FRONT-3679

planctus:


src/implementations/twig/components/social-media-follow/social-media-follow.story.js
```diff
@@ -9,9 +9,11 @@ import notes from './README.md';
 // Preserve original data.
 const dataHorizontal = { ...specs };
 const dataVertical = { ...specs, variant: 'vertical' };
+const others = { ...specs.links[5] };
```
emeryro: it may be safer to get the last item of the array, instead of a specific number (if we have to add or remove an item in 6 months or so)

src/implementations/twig/components/social-media-follow/social-media-follow.story.js
```diff
@@ -25,9 +27,27 @@ const getArgTypes = () => ({
...
 });
 
-const prepareData = (data, args) => Object.assign(correctPaths(data), args);
+const prepareData = (data, args) => {
+  const isItThere = data.links.slice(-1)[0].label === 'Other social networks';
+
+  if (!args.toggle_other && isItThere) {
+    data.links.pop();
+  } else if (args.toggle_other && !isItThere) {
+    data.links[5] = others;
```
emeryro: same thing here

emeryro (changes requested):
some small update to the story, but you can already ask for designer review

src/implementations/twig/components/social-media-follow/social-media-follow.story.js
planctus: Yep, makes sense, although i discovered that there is an "editorial" limit set to 6 items, that's the reason why i removed the mastodon icon, it was astrid requesting that.. :)

emeryro (approved):

PR #2586, opened 2022-07-07
feat(social-media-share): Revamp - FRONT-3680

planctus:


src/implementations/twig/components/social-media-share/social-media-share.html.twig
```diff
@@ -2,6 +2,7 @@
 
 {#
   Parameters:
+    - variant (string) (default: horizontal)
```
papegaill:
```suggestion
    - "variant" (string) (default: 'horizontal')
```

papegaill (changes requested):

emeryro (approved):

PR #2591, opened 2022-07-11
fix(file-upload): Fix focus styles for the browse button - FRONT-3689

papegaill:


src/implementations/vanilla/components/file-upload/_file-upload.scss
```diff
@@ -45,20 +51,33 @@
...
+    position: initial;
+    width: auto;
     margin: 0;
 
     // stylelint-disable max-nesting-depth
     &:active,
     &:focus {
-      outline: 3px solid map.get(theme.$color, 'yellow');
-      outline-offset: -3px;
+      + .ecl-file-upload__button-container .ecl-file-upload__button {
```
emeryro: I don't think that this code should be in the no-js section

emeryro (commented):
we can indeed access the hidden file upload input, but there is no visual indication of it (as it is hidden...)
Is it what is expected?

emeryro (approved):

PR #2597, opened 2022-07-14
fix(table): Fix focus for table sort arrows - FRONT-3693

papegaill:


emeryro (commented):
on my side, I have the sort buttons duplicated, so I have to tab twice to get to the next one.
Is it on purpose?
image

papegaill:
> on my side, I have the sort buttons duplicated, so I have to tab twice to get to the next one. Is it on purpose? image

Thank you, indeed, the component was double init() for EU version. 
This was previously fixed in EC, but not aligned to EU in the preview-head.hml file.

emeryro (approved):

PR #2598, opened 2022-07-18
fix(icon): Removing default, unneded but harmful transformation - FRONT-3702

planctus:
Really surprising, personally i don't have an explanation, feels like this is a chrome bug. Any item in one of the "additional" columns in the menu set as external would get the external icon "transformed" in Chrome. It was enough to remove the default transformation set in the css, scale 0 rotate 0, which looks pretty useless, to get it fixed. Examples in the menu have been updated to "prove" this, to test it you can clone one item in production and put it in a column which is not the first.


emeryro (approved):
that's a really strange bug for sure...
I don't see side effects of your fix, so let's go for it

PR #2605, opened 2022-08-04
feat(tests): Adding accessibilty tests using jest-axe - FRONT-3710

planctus:


emeryro:
@planctus how can it be tested?

planctus:
The tests are already running, they are part of the already existing jest tests.
So if you access the "tests" check here in github you will see the "passes the accessibility tests" message in most of the existing tests. I had to remove these tests from the tabs component, though, cause there is some issue non trivial there.
From the command line we have the test:components script or you can simply run yarn jest

emeryro (approved):

PR #2610, opened 2022-08-10
fix(inpage-navigation): fix z-index - FRONT-3708

emeryro:
Fix inpage navigation z-index on desktop.
It can be tested on https://codepen.io/serge321/pen/wvmpVyN


src/implementations/vanilla/components/inpage-navigation/_inpage-navigation.scss
```diff
@@ -29,6 +29,7 @@ $_outline: null !default;
     display: block;
     position: sticky;
     top: 0;
+    z-index: 0;
```
planctus: mmh..i'm not sure we want to set z 0 z-index on an element like the inpage navigation. This would only be safe in case nothing else, basically, is present in the left sidebar, otherwise any element could easily go on top of the inpage, it's enough a relative positioning for this to happen.

emeryro: that's true.. I'll test another solution

planctus:
shouldn't we do this also for the new site header..?

planctus (approved):

PR #2611, opened 2022-08-11
feat(links): Fixing display of links as buttons - FRONT-3713 

planctus:


emeryro (commented):
That seems to be fine, but I'd rather waiting for Alexis to be back. I assume thet there was a reason why he applied this dipslay only to the icon links

papegaill:
> That seems to be fine, but I'd rather waiting for Alexis to be back. I assume thet there was a reason why he applied this dipslay only to the icon links


`display` property was somehow removed when creating the other CTA variants: https://github.com/ec-europa/europa-component-library/commit/b03ea834301b4be0b1d2b9f7c6dadca3cae93513#diff-487ae4b002b7b2c86b6fc54e2d0012b3679dea1afab5e9ba0b6458394ac85399L153
I only intervened on the icon classes after/before, not wanting to break the existing properties: 
 https://github.com/ec-europa/europa-component-library/commit/4551437406aaca3b035c67a1899789b96bd861b2

papegaill (approved):

PR #2613, opened 2022-08-17
fix(description-list): Aligning expected data for the taxonomy links use case - FRONT-3724

planctus:


emeryro:
the change is fine for me, but as that would be a breaking change isn't it?
can't we have some backward compatibility here?

emeryro (approved):

PR #2623, opened 2022-08-22
fix(carousel): Force navigation and pagination initialization - FRONT-3707

papegaill:


planctus:
Totally unrelated to this pr, just wonder whether we want to take the chance to tackle this now or we create a separate ticket for this, in "tablet" kind of viewport i see this:
image

papegaill:
> Totally unrelated to this pr, just wonder whether we want to take the chance to tackle this now or we create a separate ticket for this, in "tablet" kind of viewport i see this: image

Just tested with chrome, FF & safari with different viewport and I can't reproduce this? Can you give more information?

planctus (approved):

PR #2624, opened 2022-08-22
fix(rating-field): Fix focus for safari - FRONT-3706

papegaill:
`focus-visible` style is not supported on radio elements when navigating with keyboard arrows on safari

### DIFF
--- a/src/implementations/vanilla/components/rating-field/_rating-field.scss
+++ b/src/implementations/vanilla/components/rating-field/_rating-field.scss
@@ -50,7 +50,7 @@
 }
 
 // Focus, active state
-.ecl-rating-field__input:focus-visible,
+.ecl-rating-field__input:focus,
 .ecl-rating-field__input:active {
   + .ecl-rating-field__label {
     outline: 2px solid map.get(theme.$color, 'blue');


planctus:
couldn't we use somthing like this is such a scenario..?
```
.ecl-rating-field__input:focus { /* some exciting button focus styles */ }

@supports selector(:focus-visible) {
    .ecl-rating-field__input:focus { /* undo all the above focused button styles */ }
    .ecl-rating-field__input:focus-visible { /* and then reapply the styles here instead */ }
}
```

papegaill:
> couldn't we use somthing like this is such a scenario..?
> 
> ```
> .ecl-rating-field__input:focus { /* some exciting button focus styles */ }
> 
> @supports selector(:focus-visible) {
>     .ecl-rating-field__input:focus { /* undo all the above focused button styles */ }
>     .ecl-rating-field__input:focus-visible { /* and then reapply the styles here instead */ }
> }
> ```

`focus-visible` is currently supported by Safari, it's just that when you switch an element, it doesn't take on the focus-visible style. Safari applies styles differently for this type of element.

planctus (approved):

PR #2633, opened 2022-08-26
fix(news-ticker): Fixing position of the external icon - FRONT-3735

planctus:


emeryro:
This change goes against the design choices made by COMM.
It has to be reviewed by them

PR #2641, opened 2022-09-13
fix(carousel): allow click on cta - FRONT-3747

emeryro:
Fixes the carousel, to allow click on the call to action; on current version is is not possible to click on it because of the drag function.

Can be tested on storybook (fullscreen), with responsive display of the inspector

### DIFF
--- a/src/implementations/vanilla/components/carousel/carousel.js
+++ b/src/implementations/vanilla/components/carousel/carousel.js
@@ -291,6 +291,7 @@ export class Carousel {
     e = e || window.event;
 
     if (e.type === 'touchmove') {
+      e.preventDefault();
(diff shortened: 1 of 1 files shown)


planctus (commented):
I have a very little knowledge around touch events, but it seems that the problem we faced is well known and the recommended solution is to prevent the propagation of the event on "touchmove" and not in "touchstart", more than "giving up" about it.
https://developer.mozilla.org/en-US/docs/Web/API/Touch_events

I'm not really aware of the amount of propagated events that we would or should prevent, but it seems good practice to do so, so we could consider adding the preventDefault() only on touchmove in the dragStart() method.
```
    if (e.type === 'touchmove') {
      e.preventDefault();
    }
```

planctus (approved):

PR #2644, opened 2022-09-15
fix(social-media): Fixing and aligning variant selectors - FRONT-3749

planctus:


src/implementations/twig/components/social-media-follow/social-media-follow.html.twig
```diff
@@ -37,13 +37,11 @@
 {% set _extra_attributes = '' %}
 {% set _description = description|default('') %}
 {% set _links = links|default({}) %}
-{% set _variant = variant|default('') %}
+{% set _variant = variant|default('horizontal') %}
```
emeryro: the class `ecl-social-media-share--horizontal` does not exists in our css. Couldn't we keep it empty by default, and have only `vertical` as a possible modifier?

planctus: Ok, the idea was mainly to align the two templates, i did it applying only the vertical variant in the two templates, now. ;)

emeryro (approved):

PR #2647, opened 2022-09-19
feat(multiselect): Adding toolbar and selection counter - FRONT-3746

planctus:


src/implementations/vanilla/components/select/select.js
```diff
@@ -105,13 +117,18 @@ export class Select {
     this.handleFocusout = this.handleFocusout.bind(this);
     this.handleSearch = this.handleSearch.bind(this);
     this.handleClickOutside = this.handleClickOutside.bind(this);
+    this.handleClickOnClose = this.handleClickOnClose.bind(this);
     this.moveFocus = this.moveFocus.bind(this);
     this.resetForm = this.resetForm.bind(this);
+    this.resetValues = this.resetValues.bind(this);
```
emeryro: shouldn't we call it handleClickOnClearAll, to be consistent?

src/implementations/vanilla/components/select/select.js
```diff
@@ -288,6 +322,39 @@ export class Select {
...
+        this.closeButton = document.createElement('button');
+        this.closeButton.textContent = this.closeButtonLabel;
+        this.closeButton.classList.add('ecl-button', 'ecl-button--primary');
+        this.closeButton.addEventListener('click', this.handleClickOnClose);
+        this.closeButton.addEventListener(
+          'keydown',
+          this.handleKeyboardOnClose
+        );
+        this.dropDownToolbar.appendChild(this.closeButton);
+        this.dropDownToolbar.style.display = 'none';
```
emeryro: why is this styling inside the if for the close button label?

src/implementations/vanilla/components/select/select.js
```diff
@@ -547,6 +667,13 @@ export class Select {
     }
   }
 
+  /**
+   * Close the dropdown when the use clicks on the button
+   */
+  handleClickOnClose() {
```
emeryro: it seems to be the same thing as handleEsc()
Maybe we could use the same function for both

emeryro (commented):
That's a little strange to not have the "close" button displayed by default, even if nothing is selected, but if it has been validated as is...

planctus:
Thanks for reviewing Romain, i've also added some ifs in the js since i've noticed that we generate the toolbar only if we have buttons to render.
About the behaviour, it's them who requested it as such, initially i was always showing the buttons.. :)

emeryro (approved):

PR #2649, opened 2022-09-22
docs(website): update webtools documentation - FRONT-3742

emeryro:


src/website/src/pages/ec/resources/webtools/index.md
```diff
@@ -3,98 +3,14 @@ order: 3
...
-- documentation: https://webgate.ec.europa.eu/fpfis/wikis/display/webtools/Social+bookmarking+and+networking
-- wizard: https://europa.eu/webtools/mgmt/wizard/ui.php?uec=sbkm
-
-### Twitter feeds
-
-Embed a series of specific tweets, free of cookies. Choose between tweets from a certain user, list, search result, a collection or embed a single tweet.
-
-- wizard: http://europa.eu/webtools/mgmt/wizard/ui.php?uec=smk
-- documentation: https://webgate.ec.europa.eu/fpfis/wikis/display/webtools/Social+Media+Kit+-+Technical+details
+Official website: https://webgate.ec.europa.eu/fpfis/wikis/display/webtools
```
planctus: but this page is not publicly accessible one.. Don't they have any mirror where it can be accessed without login..? We should at least specify that EU login is needed to access this link, imho,  if there is none...

emeryro: well I can add a line for that, but in any case, Webtools widgets works only on Europa network, so people have EU login there

planctus (approved):

PR #2653, opened 2022-09-27
feat(table): add parameter for enhanced mobile - FRONT-3754

emeryro:
Add a parameter to disable the enhanced mobile display. If set, table will only have an horizontal scroll on small screens

⚠️ Markup change: to have a default responsive behaviour, a wrapper had to be added around the table. This wrapper should be set in all cases (enhanced or not)


src/implementations/twig/components/table/table.story.js
```diff
@@ -10,30 +10,58 @@ import notes from './README.md';
...
+      'Enhanced display for mobile. This may depends on your needs and the data passed.',
+    table: {
+      type: { summary: 'boolean' },
+      defaultValue: { summary: false },
+      category: 'Display',
+    },
+  },
+});
+
+const prepareData = (data, args) => {
```
planctus: this can simply be `const prepareData = (data, args) => Object.assign(data, args);`, no..?

src/website/src/pages/eu/components/table/docs/code.mdx
```diff
@@ -34,3 +40,21 @@ import { tableDefault, tableMulti, tableSortable, tableZebra } from '../demo';
 >
   <Html markup={tableSortable} />
 </Playground>
+
+## Enhanced mobile table
+
+Enhanced tables offers a more user friendly display on mobile.
```
planctus, line 40: But why we don't specify what are the "tables" where this can be safely enabled..? I mean, for this to work we need the first tr to contain th(s) as a basic requirement.
Then this might not be enough in 100% of the cases, so this is why making this optional.

planctus (commented):
To me, while making this optional is a really good way to avoid breaking potentially correct tables, this way we lose any "responsiveness" for tables.
Shouldn't we introduce something like bootstrap does..? And maybe apply this by default, if the user wants he can use the enhanced when it's going to improve the mobile look and feel. 
https://getbootstrap.com/docs/5.0/content/tables/#responsive-tables

planctus (dismissed):
There is one thing that we are not handling, but i don't see how could we, if created in the wyswyg editor the table will not be responsive, unless they do something on the implementation side..

PR #2660, opened 2022-10-10
fix(print): fix display of default html tag for print - FRONT-3699

emeryro:
Added an option css to print elements inside the .ecl class


planctus (commented):
just one question..is this really meant to be an optional css to be added to the print version of our css or is this just creating an overhead, both on our and the customers side..? 
i think this could actually be part of the print css so that no adaptation would be needed for the users to benefit from this, no new resource to be generated and provided with examples every release for it to be included when using ECL.

emeryro:
@plactus to me that's the same thing as the other "html-default" we have. It will only be used by EWPP to cover their need in the wysiwyg.
As we decided to keep this hacky css out of the main one for screen display, I guess it makes sense to do the same for print

planctus (approved):

PR #2661, opened 2022-10-10
feat(focus): New styles for checkboxes and radios - FRONT-3770

planctus:


emeryro (commented):
on EC, focus element have an outline in any case, but on EU only the selected checkbox has an outline, the other element only change color
Is that the intended behaviour?
image
image

planctus:
thanks @emeryro , yep, this is the file on figma with the specs:https://www.figma.com/file/Bo9MKSWsatH7qSdjZwxMEe/Form-Elements for eu, as in the initial comment for this pr the things that are not fully aligned with what you find there should be the focus on error state and the size of the selected-focused checkbox in eu.

emeryro (approved):

PR #2669, opened 2022-10-18
fix(site header): fix z-index for mobile - FRONT-3768

emeryro:
Fix z-index conflict for the mobile menu. This had to be done at the level of the site header for the z-index to be applied correctly
Can be tested on https://finance.ec.europa.eu/digital-finance_en

### DIFF
--- a/src/implementations/vanilla/components/site-header/site-header-ec.scss
+++ b/src/implementations/vanilla/components/site-header/site-header-ec.scss
@@ -22,7 +22,7 @@ $_search-width-xl: 31.5rem;
   background-color: map.get(theme.$color, 'white-100');
   margin: 0;
   position: relative;
-  z-index: 2;
+  z-index: map.get(theme.$z-index, 'modal');
 
   .ecl-menu {
     background: transparent;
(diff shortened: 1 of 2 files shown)


planctus (commented):
The issue in the website where the problem has been identified with the inpage on top of the menu in mobile is not coming from ECL, but from the markup used, the inpage navigation where we set a z-index: 1 in ECL has this overridden by the ecl-u-z-navigation utility used on it. (z-index: 10)

I'm not sure that it was a good idea, the one to set the z-index of the site header to any particular value, it feels like z-index: auto is more what we would do there instead of raising its value to overcome issues like the one coming from the laco widget, that doesn't have a z-index set, per se.
Using auto would also fix the global banner overlapping the menu..what do you think, Romain..?

emeryro:
Thanks @planctus 
This issue with the z-index utility class has been raised to EWPP in another ticket. But anyways, I did some more test, and unfortunately z-index: auto does not work in all situation. That's quite fine for EC header, but for EU it causes issue with the shadows.
You can test it here https://next-generation-eu.europa.eu/make-it-green_en (it's also slightly visible on the main europa.eu site)

planctus:
You refer to the shadow of the site-header being overlapped by the breadcrumb..? It is part of the page-header, and the page-header has a white background, so we would not need to set also the background of the breadcrumb, in this variant.
In any case, for me we can go with this "arbitrary" z-index, but if feels we have to adjust things from time to time because of the z-indexes we set and it would feel better knowing that we can let the browser create the right layering itself.

planctus (approved):

PR #2682, opened 2022-11-07
fix(storybook): Resetting static dirs when building - FRONT-3789

planctus:
This removes this error in the build:
```
(node:6414) UnhandledPromiseRejectionWarning: Error: EEXIST: file already exists, mkdir '/home/runner/work/europa-component-library/europa-component-library/src/playground/ec/build/styles'
(Use `node --trace-warnings ...` to show where the warning was created)
(node:6414) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:6414) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:6414) UnhandledPromiseRejectionWarning: Error: EEXIST: file already exists, mkdir '/home/runner/work/europa-component-library/europa-component-library/src/playground/ec/build/styles'
(node:6414) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
```
We are not asking anymore storybook to copy files from the static folders we need when building, basically.


emeryro (changes requested):
I tried to do a full build after cleaning my folder, and it seems that resources are missing for storybook (running yarn start:eu)
image

planctus:
wow, thank god you tested this.. :)

emeryro (approved):

PR #2686, opened 2022-11-15
feat(range): add Range component - FRONT-3756

emeryro:


src/implementations/vanilla/components/range/range.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Display value when changed
+   */
+  handleChange() {
+    // Update value
+    this.currentValue.innerHTML = this.rangeInput.value;
+
+    return this;
```
planctus: why do we return the instance here..?

src/implementations/vanilla/components/range/range.js
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Initialise component.
+   */
+  init() {
+    this.rangeInput = queryOne(this.rangeInputSelector, this.element);
+    this.currentValue = queryOne(this.currentValueSelector, this.element);
+
+    if (this.rangeInput && this.currentValue) {
+      // Display default value
+      this.currentValue.innerHTML = this.rangeInput.value;
```
planctus, line 65: Reading the specs the "correct" way of retrieving the value of a range input type is "valueAsNumber", but it's not really relevant here

src/implementations/vanilla/components/range/_range.scss
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+.ecl-range[disabled],
+.ecl-range--disabled,
+.ecl-range[readonly],
+.ecl-range--readonly {
+  background-color: $_disabled-color;
+
+  &::-webkit-slider-thumb {
```
planctus: i see that we adding vendor prefixes here, we could use https://www.npmjs.com/package/postcss-input-range instead

planctus (commented):
Very nice, i just think we could improve the css letting postcss add the vendor prefixes like we do in all the other components.

planctus (approved):

PR #2687, opened 2022-11-15
fix(rating field): use dynamic id and name - FRONT-3796

emeryro:
Remove hardcoded id and name for the rating field


src/implementations/twig/components/rating-field/README.md
```diff
@@ -8,34 +8,25 @@ npm install --save @ecl/twig-component-rating-field
...
-  - "value" (string) (default: '')
-  - "disabled" (boolean) (default: false)
-  - "checked" (boolean) (default: false)
-  - "required" (boolean) (default: false)
-  - "helper_id" (string) (default: '')
-  - "helper_text" (block) (default: '')
-  - "label" (block) (default: '')
+- **"id"** (string) (default: 'rating'): unique id for the rating field
+- **"items"** (array) (default: []): array of radio_button
 - **"name"** (string) (default: '')
```
planctus, line 13: we are setting a default  "rating" for this parameter, i wonder whether we really need to get two separated parameter for setting ids and names of the inputs, couldn't we just use one and use it as such for the name of the input group and for the ids with an appended string..?

src/implementations/twig/components/rating-field/rating-field.html.twig
```diff
@@ -106,9 +108,10 @@
   {% if _items is not empty and _items is iterable %}
   <div class="ecl-rating-field">
     {%- for _item in _items|reverse %}
+      {% set _item_id = _item.id is not empty ? _item.id : _id ~ '-' ~ _item.value %}
```
planctus: i think we should add an escape for the value here. `item.value|e('html_attr')`, the value is a string but we don't know what it contains..

planctus (changes requested):
Small thing in the doc.. ;)

src/implementations/twig/components/rating-field/README.md
emeryro: I'd rather keep both name and id separated, as they are not exactly the same thing. Also that's how we defined the checkboxes/radio.
But I'll simplify a bit the logic

planctus (approved):

PR #2699, opened 2022-11-28
feat(rating-field): add read only

emeryro:
Add a parameter to display the rating field as read only


planctus (commented):
I have a question...
In html we have "disabled" but also "readonly", now it seems we are mixing them up since we say that a disabled form element/group is a readonly one.
The difference is that the readonly input still submits the value, disabled doesn't and here it really seems that what we want is a "readonly" field more than a disabled one, since we set a value for it, i would expect that to be part of the submitted form values.

emeryro:
That's a valid question, but I'd say that it will depend on the needs.
For now, the reported need is from eUi, to display the value of the rating in a listing, oustside of any form. So that correspond to what we have here.

planctus (approved):

PR #2700, opened 2022-11-28
feat(file): Adding detailed meta in default variant - FRONT-3813

planctus:


src/specs/components/file/demo/data--without-translation.js
```diff
@@ -4,6 +4,7 @@ const exampleLink = `${publicUrl}/example`;
 module.exports = {
   title: 'State of the Union 2018 brochure',
   aria_label: 'Download file State of the Union 2018 brochure',
+  detail_meta: ['Resource type', 'Publication date'],
```
emeryro, line 7: before we recieve feedback, I'd suggest to put only the date here, as it is on figma

emeryro (changes requested):
It seems that on figma, the date is displayed bold on EC, to be clarified.

Later on, we should take time to cleanup a little this twig file, it is quite hard to read...

src/specs/components/file/demo/data--without-translation.js
planctus: also in figma you have the meta + date, this is also demoing the right "spacing" between the items, i think we can leave it like this, also because for the mess we make in the story this gets currently replaced everytime with the same value coming from another data file.

planctus:
True, the font weight in EC is inconsistent in the different examples, this is the one the "implementation" is based on:
image

planctus:
About the template, yes, it's a crap, the thing is that the thumbnail variant messes things up, but i guess that if we decide to "refactor" it this would aprobably involve some design change.

emeryro (dismissed):

src/implementations/vanilla/components/file/_file.scss
```diff
@@ -62,6 +62,7 @@ $_image-width: null !default;
   flex-basis: 50%;
   flex-grow: 1;
   flex-wrap: wrap;
+  flex-direction: column;
```
emeryro: with this change, the file type and size go under the name (instead of inline before)
Is that something requested by designers?
image

planctus: uff, no it's not..thanks for noticing this.. ;)

emeryro (approved):

PR #2704, opened 2022-12-05
fix(select): fix multi-select with long labels - FRONT-3817

emeryro:
Add a wrapper around the label text (in javacript), to fix display issue with long label

⚠️  Markup change: markup generated by js has been changed


planctus (commented):
If we are about to do this i guess we should mark the pr as a breaking change, or we could introduce a new "category" to mark the prs in the changelog that require changes in the markup, eUi doesn't use our js and this change would not be transparently available for them.

emeryro:
@planctus  I think that using the breaking change tag would be enough.
Yes, the entry would be duplicated in the change log, but that would just highlight more what is breaking or not. Or do you foresee other issue with having it twice?

emeryro:
Another option would be to just put it in the PR description

PR #2709, opened 2022-12-07
feat(menu): improve keyboard navigation - FRONT-3798

emeryro:
Improve keyboard validation on small screen, by looping in the menu items with "tab" key


planctus (commented):
I find just a bit incosistent the fact that the only way to navigate the items in the mobile menu is by using tab, while we support also arrow up and down in desktop.
There is also a small improvement we could do, when you have a long list of sub-items you would maybe expect, after reaching the last, to restart from the first or in any case not find yourself stuck there having to find your way back to the menu :)

emeryro:
I'm a little surprised here, what you describe as possible improvement is precisely what has been added in this PR.
If I'm not wrong, it is already possible to use the arrows to navigate on mobile, and it loops back to the top of the menu after the last sub item

planctus:
mmhh..ok, it is possible also in mobile to use arrows to navigate the list of subitems, true, but i had to discover how to do it, what i mean is that you can only use Tab to access the first item and then you can browse the elements with the arrows (the four of them), but since you reached the list using TAB it comes natural to keep tabbing, more than switching on the arrows... 
When you reach the last item, only arrow up/left work and that gives focus on the close button, which you wouldn't expect and this breaks the continuity of the user experience because on the close button arrows are not working. Shouldn't the user be able to traverse back the list still using arrows, instead..?

I did not find this navigation with the keyboard particularly consistent, personally, but if this is the expected behavior, fine.

emeryro:
I agree that the arrow navigation is not optimal on mobile.
For now, this ticket was about the use of the tab key, so we can consider it done, and see later to improve key navigation

planctus (approved):

PR #2712, opened 2022-12-08
feat(select): Grouping capabilities for simple and multiple select - FRONT-3542

planctus:
- this is only adding the support for optgroups in the default and multiple select
- in the mocks there are now styles for the default select that were not foreseen and cannot be reproduced simply with some css, these styles have not been "reproduced".
- It is pretty limited as a functionality, we do support groups in the multiselect but as simple strings, no functionality is associated to it (like selecting all the options in the group)


emeryro (dismissed):

emeryro:
The hover area of item in groups is quite strange
image
I guess that it is a mistake, as the click area goes on the full width

emeryro:
Also the cursor does not seem consitent. For instance title have a hand cursor while they are not clickable, and hover area has a default cursor
image

emeryro (dismissed):

PR #2714, opened 2022-12-12
feat(spinner): fix center alignment - FRONT-3821

emeryro:


planctus (commented):
I'd say that the fix makes sense, but looking at the preview is still a bit off, less than before but if you measure the left and right portion of the viewport, is not the same..
I guess it depends on the flex style of the element, it's probably the center alignment causing this, but it's not clear the reason why.

planctus (approved):

PR #2716, opened 2022-12-15
feat(site-header): use popover for language list - FRONT-3761

emeryro:
Use a new language switcher in the site header. This new element is directly integrated in the site header component.
Previous component "language switcher" is now deprecated


planctus (commented):
image

I think we could handle this use case where the label of the list might be wider than the list itself when it goes to a single column, to prevent the misalignment between the two lists simply adding a `width: max-content` to the  `.ecl-site-header__language-category` selector

PR #2732, opened 2023-01-12
feat(shadow): harmonize shadows - FRONT-3828

emeryro:
Add sass variables and utilities for shadows
Update display of components using shadows


planctus (dismissed):

planctus:
The size check is complaining about the EU css.. :)

PR #2735, opened 2023-01-23
feat(social media share): add popover - FRONT-3830

emeryro:
Provide a working example of popover in the social media share and social media follow components.
Also slightly update display of popover component.

Note: data structure of social media links has been updated to fully use ECL Link component, but previous structure is still suported


src/implementations/twig/components/social-media-share/social-media-share.html.twig
```diff
@@ -67,21 +51,51 @@
...
+          icon: link.icon|default({})|merge({
+            extra_classes: _icon_extra_classes,
+          }),
           extra_classes: _link_extra_classes,
         }) only %}
       </li>
-    {% endif %}
     {% endfor %}
+
+    {% if _popover is not empty %}
```
planctus, line 76: I understand that the popover expects its own, additional data, compared to a simple link, but i wonder, do we really need to duplicate the processing of the item when it's a popover..? couldn't this be handled by the same code block, in the end the popover is a link plus an array of links:
`...
{ 
  link: {
    ...
 },
  icon: {
    ...
  },
 {
  popover: true,
  link: {
    ...
  },
  icon: {
    ...
  },
  links: [
    ...
],
`
If link.popover is set we include the popover instead of the link template, with the link becoming the popover "toggle"... what do you think..?

src/implementations/twig/components/social-media-follow/social-media-follow.html.twig
```diff
@@ -63,26 +48,56 @@
...
+      <li class="ecl-social-media-follow__item">
+        {% set _link_extra_classes = 'ecl-social-media-follow__link' %}
+        {% if link.extra_classes is not empty %}
+          {% set _link_extra_classes = _link_extra_classes ~ ' ' ~ link.extra_classes %}
+        {% endif %}
+        {% set _icon_extra_classes = 'ecl-social-media-follow__icon' %}
+        {% if link.icon is not empty and link.icon.extra_classes is not empty %}
+          {% set _icon_extra_classes = _icon_extra_classes ~ ' ' ~ link.icon.extra_classes %}
+        {% endif %}
+        {% include '@ecl/link/link.html.twig' with link|merge({
```
planctus: isn't this simply:
```
        {% include '@ecl/link/link.html.twig' with {
          link: link.link|default(link)|merge({
            type: 'standalone',
          }),
```

src/implementations/twig/components/social-media-follow/social-media-follow.story.js
```diff
@@ -27,26 +34,17 @@ const getArgTypes = () => ({
...
-    table: {
-      category: 'Content',
-    },
-  },
 });
 
 const prepareData = (data, args) => {
-  const isItThere = data.links.slice(-1)[0].label === 'Other social networks';
+  correctPaths(data);
+  const clone = JSON.parse(JSON.stringify(data));
```
planctus, line 41: maybe spread syntax..? ` const clone = { ...data };`

planctus (changes requested):
It seems we can improve a bit the syntax of this backward compatible twig merge we do in this template.
I would add also a test about the backward compatibilty.

src/implementations/twig/components/social-media-share/social-media-share.html.twig
emeryro: I guess that it could be the case, indeed. But that would mean that the popover could be located on every link (not only the last one), based on the data passed. There could also be multiple popover. This is not really the expected behaviour.
In the end, this twig template will most likely never be used (that's still a webtools widget), do I think that we shouldn't worrk to much about its optimisation.

src/implementations/twig/components/social-media-follow/social-media-follow.html.twig
emeryro: not exactly. We also add a default class in extra_classes, while keeping the possiblity to have custom ones coming from the data (lines 54 and 60)

src/implementations/twig/components/social-media-follow/social-media-follow.story.js
emeryro: I tried this already, and if I'm not wrong that won't work correctly. The spread syntax does not create a real copy, just a reference, so when deleting of editing the "cloned" data set, it would also be applied to the original one

src/implementations/twig/components/social-media-follow/social-media-follow.html.twig
planctus: sure, that was only a "partial", the full include would be:
```
        {% include '@ecl/link/link.html.twig' with {
          link: link.link|default(link)|merge({
            type: 'standalone',
          }),
          icon: link.icon|default({})|merge({
            extra_classes: _icon_extra_classes,
          }),
          extra_classes: _link_extra_classes,
        } only %}
```
currently there is a parent merge, that is what i'm removing 
`{% include '@ecl/link/link.html.twig' with link|merge({`

src/implementations/twig/components/social-media-follow/social-media-follow.story.js
planctus: forget it, sorry.. like if i never experienced this.. :D spread on the object works as a clone, but not for nested objects and we have plenty in our data..

planctus (approved):

PR #2736, opened 2023-01-23
feat(message): make close button optional - FRONT-3837

emeryro:
Update twig template to make close button optional.
Button is now displayed only if the "close" parameter is passed.


src/implementations/twig/components/message/message.html.twig
```diff
@@ -52,19 +52,19 @@
...
-      extra_classes: 'ecl-message__close',
-      extra_attributes: [{
-        name: 'data-ecl-message-close',
-      }],
-    }) only %}
+    {% if _close is not empty %}
+      {% include '@ecl/button/button.html.twig' with _close|merge({
+        type: 'button',
+        variant: 'ghost',
+        icon: _close.icon|merge({ name: 'close-filled' })|default({}),
```
planctus: this line seems problematic to me:
the default should probably be set for the _close.icon, 
`_close.icon|default({})|merge({ name: 'close-filled' }),`
which we don't know whether is set, but in any case passing an icon object without checking for the icon_path is not ensuring that the icon will be rendered.

planctus (changes requested):
Just a remark about the way we build the object we pass for the close button

planctus (approved):

PR #2742, opened 2023-01-30
feat(carousel): add different sizes - FRONT-3842

emeryro:
Add 3 possible sizes for the carousel, based on the sizes used in the new banner


src/implementations/twig/components/carousel/carousel.story.js
```diff
@@ -21,7 +22,7 @@ const getArgTypes = () => {
       name: 'width',
       type: 'select',
       description: `The media container extends to the whole viewport by default when outside the grid,
-        if it's inside it can still be extended by adding class .ecl-page-banner--full-width`,
+        if it's inside it can still be extended by adding class .ecl-banner--full-width`,
```
planctus, line 25: there is a default type "radio" in this control, considering that this is not related to a param in the template i would simply remove the table with the defaults

planctus (commented):
image

It seems to me that we might be experiencing again issues related to the init/destroy/re-init of our components in storybook, this happens when selecting the large size with the outside of the container, if then you open the devtools, for instance, the slides get resized..

PR #2747, opened 2023-02-01
feat(banner): Rename variants - FRONT-3845

planctus:


emeryro (changes requested):
Carousel data should be updated too, as it uses banner
image

emeryro (approved):

PR #2761, opened 2023-02-09
fix(file): Using safer approach for label markup - FRONT-3854

planctus:


emeryro (approved):
Not sure why our code didn't work on php, but fine

PR #2768, opened 2023-02-14
chore(flags):  Reducing icon sizes

planctus:


emeryro (commented):
It seems good, I just wonder why the new icons are defined with a size of 150x150 instead of 40x40
image

planctus:
Right, i noticed that in the new svgs width and height were not set, i guess that 150x150 is a sort of a default in github for svg, anyway, i added the dimension manually in the svgs. Will tell this to the designers.

emeryro:
Another remark: it seems that the color of some flags are different (much brighter). I don't know if that was on purpose
image
If it's fine for the designers, so it is for me too...

emeryro (approved):

PR #2776, opened 2023-02-20
feat(modal): add modal component - FRONT-3858

emeryro:


src/implementations/twig/components/modal/modal.html.twig
```diff
@@ -0,0 +1,114 @@
...
+{% endif %}
+
+{# Print the result #}
+
+<dialog id="{{ _id }}" class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+  <div class="ecl-modal__container ecl-container">
+    <div class="ecl-modal__content ecl-col-12 ecl-col-m-8 ecl-offset-m-2 ecl-col-l-6 ecl-offset-l-3">
+      <header class="ecl-modal__header">
+        {% if _header_icon is not empty and _header_icon.icon is not empty %}
+          {% set _icon_extra_classes = 'ecl-modal__header-icon' %}
```
planctus: i think this should simply be:
```
          {% include '@ecl/icon/icon.html.twig' with _header_icon|merge({
            icon: _header_icon.icon|merge({
              path: _header_icon.icon.path|default(_icon_path),
              size: 'm',
            }),
            extra_classes: 'ecl-modal__header-icon ' ~ _header_icon.extra_classes ?: '',
          }) only %}
```
I assume that we want to fix the size of the icon and provide a fallback for the path in case that is not set in the icon object. So we merge our data into the user data, otherwise it works the other way around..

src/implementations/twig/components/modal/modal.html.twig
```diff
@@ -0,0 +1,114 @@
...
+              path: _header_icon.icon.path|default(_icon_path),
+              size: 'm',
+            }|merge(_header_icon.icon),
+            extra_classes: _icon_extra_classes,
+          }) only %}
+        {% endif %}
+
+        {% if _header is not empty -%}
+          <div class="ecl-modal__header-content">
+            {%- block header _header -%}
```
planctus: not sure the reason why we keep using blocks like this, there is no difference whatsoever with a standard parameter, normally a block would be overridable and that is their main use case, but we use it just as a variable.
In any case, since we do this in many other templates, let's keep doing it..

src/implementations/twig/components/modal/modal.html.twig
```diff
@@ -0,0 +1,114 @@
...
+          extra_attributes: [{
+            name: 'data-ecl-modal-close',
+          }],
+        } only -%}
+      </header>
+      {% if _body is not empty %}
+      <div class="ecl-modal__body">{%- block body _body -%}</div>
+      {% endif %}
+      {% if _buttons is not empty %}
+      <footer class="ecl-modal__footer">
```
planctus, line 109: why is this footer hardcoded instead..? This is what a block is meant for, theoretically we could define this markup but make it overridable. It looks arbitrary, though, the use this footer for action buttons expecting user data to build this specific elements. To me this could be simply a twig block and for the default example we would pass this markup with buttons..

src/implementations/vanilla/components/modal/modal.js
```diff
@@ -0,0 +1,198 @@
+import { queryAll } from '@ecl/dom-utils';
```
planctus, line 1: we are adding very similar scripts with all these new js driven components...it feels that we would benefit from defining a common base class, like for popover, modal, at least, and extend it.
Not saying to do it right now, maybe, but we are duplicating code and increasing the js size, this way.

src/website/src/pages/ec/components/modal/docs/code.mdx
```diff
@@ -0,0 +1,19 @@
...
+  system="ec"
+  selectedKind="components-modal"
+  selectedStory="default"
+  showFrame
+  frameHeight="400"
+>
+  <Html markup={modal} />
+</Playground>
+
+Note: you can have any button in the modal act as a close button. To do so, simply add `data-ecl-modal-close` to it.
```
planctus, line 19: About this note, from what i see in the template we are harcoding the close button, so this is true for the vanilla component but not for the twig implementation.

src/implementations/vanilla/components/modal/modal.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    // Bind click event on toggle
+    if (this.attachClickListener) {
+      this.element.addEventListener('click', this.handleClickOnToggle);
+    }
+
+    // Bind click event on close buttons
+    if (this.close && this.attachClickListener) {
+      this.close.forEach((close) => {
```
planctus, line 108: but why do we query for multiple close buttons..? i mean, i assume that this came as a "requirement" sort of, but still, do we really expect multiple close buttons in the same dialog?

src/implementations/vanilla/components/modal/modal.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    this.openModal();
+  }
+
+  /**
+   * Open the modal.
+   */
+  openModal() {
+    if (this.isDialogSupported) {
```
planctus, line 162: nice, but is this really enough..? showModal() is not only revealing the modal, so i guess that our "polyfill" here is only capable of showing and hiding the modal, not the backdrop, (but i see we are not using it, we set the background on the ecl-mdoal) as well as not blocking interaction outside the dialog, right..?

src/implementations/vanilla/components/modal/modal-print.scss
```diff
@@ -0,0 +1,11 @@
+/*
+ * Modal
+ * @define modal
+ */
+
+@use 'sass:map';
+@use '@ecl/theme-dev/theme';
+
+.ecl-modal {
```
planctus, line 13: Not sure you've got requirements for the print css, but maybe we could do something more here, like hiding the close button and applying basic styles to the elements.

src/implementations/vanilla/components/modal/_modal.scss
```diff
@@ -0,0 +1,116 @@
...
+  align-items: center;
+  border-bottom: 1px solid $_separator-color;
+  display: flex;
+  font: map.get(theme.$font-prolonged, 'l');
+  padding: map.get(theme.$spacing, 'm');
+}
+
+.ecl-modal__header-icon {
+  flex-shrink: 0;
+  margin-right: map.get(theme.$spacing, 's');
```
planctus: this should be margin-inline-end, right..?

src/implementations/vanilla/components/modal/_modal.scss
```diff
@@ -0,0 +1,116 @@
...
+  margin-inline-start: map.get(theme.$spacing, 'm');
+
+  &:last-of-type {
+    margin-inline-start: 0;
+  }
+}
+
+.ecl-modal__close {
+  padding: map.get(theme.$spacing, '2xs') map.get(theme.$spacing, 's');
+  margin-right: -#{map.get(theme.$spacing, 's')};
```
planctus: also here, margin-inline-end..

src/implementations/vanilla/components/modal/_modal.scss
```diff
@@ -0,0 +1,116 @@
...
+.ecl-modal__content {
+  background-color: map.get(theme.$color, 'white-100');
+  border-radius: $_border-radius;
+  box-shadow: map.get(theme.$shadow, '4');
+  font: map.get(theme.$font-prolonged, 'm');
+  padding: 0;
+}
+
+.ecl-modal__header {
+  align-items: center;
```
planctus: it feels it might work better with an "flex-start" value, in mobile:
image

src/implementations/vanilla/components/modal/modal.js
```diff
@@ -0,0 +1,198 @@
...
+   */
+  static autoInit(root, { MODAL: defaultOptions = {} } = {}) {
+    const modal = new Modal(root, defaultOptions);
+    modal.init();
+    root.ECLModal = modal;
+    return modal;
+  }
+
+  constructor(
+    element,
```
planctus, line 29: it feels a bit odd, our "modal" element is actually a trigger button that only takes care of opening our modal if it finds that with a query on the dom.
This means that the only possible way to initalise a modal is by having an element that identifies the modal via an aria-controls attribute.
I think this should be one of the ways to launch a modal, by calling a method in its class that shows the Element of the class (this) and this should be also possible without clicking anywhere.
This is probably why this script is so similar to the popover one, even though the two components are different: it makes sense to "attach" a popover to an element that will reveal it, since it's a sort of a tooltip, indeed, but a modal is something different...

src/website/src/pages/ec/components/modal/index.md
```diff
@@ -0,0 +1,8 @@
+---
```
planctus, line 1: in the thumbnail i see the buttons (to me this is a particular use case, it shouldn't be the default for a modal component), aligned to the left, but they are right aligned in the implementation, actually it's the flex direction, row-reverse, doing it, but this is not intuitive, users will pass one button after the other (again, to me this button thingy should simply not be there) and be probably suprised to see them reversed in the order...

planctus (changes requested):
Some minor things and a question regarding the overall approach in the definition of this component.

src/implementations/twig/components/modal/modal.html.twig
emeryro: that's right.
the only "drawback" I see with your proposal is that when there is no extra classes, there is an additional space
image

src/implementations/twig/components/modal/modal.html.twig
emeryro: that's what I did initially, a simple block in the footer. But according to the designer, the footer is meant to contain only buttons.
I guess that I could keep an empty block, just in case, and have the buttons parameter available too, to have them quickly

src/website/src/pages/ec/components/modal/docs/code.mdx
emeryro: It is supposed to be true for both.
There is a hardcoded button in the header, but you can also have other close button wherever needed. It the example, the secondary button is also a close button

src/implementations/vanilla/components/modal/modal.js
emeryro: yes, there could be another one in the footer, according to design

src/implementations/vanilla/components/modal/modal.js
emeryro: I tested it on some old browsers who supposedly do not support dialog, and it seems to work just fine.
If we recieve complains we coudl investigate more, but I'd say that for now this should be enough

src/implementations/vanilla/components/modal/modal.js
emeryro: I'm not sure I follow you here. For the modal, in any case, we have to have a trigger (button, link or whatever), saying "when you click on me it will open this modal". That's the puropse of the aria-controls.
Or maybe you were considering the other way around (creating a modal and "attaching" it to an element)?

src/implementations/twig/components/modal/modal.html.twig
planctus: yeah, this was more about simplifying and fixing the merge, you can build this string differently to avoid the issue

src/implementations/twig/components/modal/modal.html.twig
planctus: i would prefer wrapping the current code to include the buttons into a twig block, if the users want, they can override the block with something else in a child template or passing markup. The result is not that different, in the end, but it would be maybe the first time we use a twig block for a reason :)
If we have to stick to the requirements, it seems this is not needed, though, so if designers want buttons only, twig users will get only those..

src/implementations/twig/components/modal/modal.html.twig
emeryro: in the end it seems that the user is not free to use whatever icons he wants, and it is more some variant of the modal, similar to message.
So I will refactor this block and add a variant parameter instead...

src/implementations/vanilla/components/modal/modal.js
planctus: yep, the component is a modal, so a dialog on top of a backdrop thingy . You could potentially want to show it without any previous interaction by the user or element controlling it the same way you could have multiple triggers for the same modal.
In any case, we can ship it like this and see the reactions, it is true that in most cases the way to interact with a modal would be a single element in the page.

emeryro: well, that's possible indeed. So the javascript would be tied to the modal itself (with the auto-init, etc.), and the "target" would be the trigger.
This means refactoring a little the javascript to make it work the other way around (this.element would be the modal and this.target would be the trigger).
If we wan tto do it, it should be done now, before we release the first version of the component

src/implementations/twig/components/modal/modal.html.twig
emeryro: I updated the structure of these blocks. Let me know if this is more correct this way

src/implementations/vanilla/components/modal/modal.js
planctus: i would go for that, it seems the right way

src/website/src/pages/ec/components/modal/index.md
planctus: shouldn't we ask for an updated thumbnail with these buttons aligned on the right..?

src/implementations/twig/components/modal/modal.html.twig
```diff
@@ -0,0 +1,131 @@
...
+{# Print the result #}
+
+<dialog data-ecl-auto-init="Modal" id="{{ _id }}" data-ecl-modal-toggle="{{ _toggle_id }}" class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+  <div class="ecl-modal__container ecl-container">
+    <div class="ecl-modal__content ecl-col-12 ecl-col-m-8 ecl-offset-m-2 ecl-col-l-6 ecl-offset-l-3">
+      <header class="ecl-modal__header">
+        {% if _header_icon is not empty %}
+          {% include '@ecl/icon/icon.html.twig' with {
+            icon: {
+              path: _icon_path|default(''),
```
planctus: we define this variable, no need for the default here, what i would do is to add a condition about the _icon_path not being empty before including the icon

src/implementations/vanilla/components/modal/modal-print.scss
planctus: Now it looks much better, but there is no margin between the two buttons:
image

This only happens when enabling also the "default-print" css

src/implementations/twig/components/modal/modal.html.twig
```diff
@@ -0,0 +1,131 @@
...
+            extra_classes: 'ecl-modal__icon',
+          } only %}
+        {% endif %}
+
+        {% if _header is not empty or block('header') is not empty %}
+          <div class="ecl-modal__header-content">
+            {%- block header %}{{ _header }}{% endblock -%}
+          </div>
+        {% endif %}
+
```
planctus, line 82: also here i would add a check for the icon_path not being empty to determine whether to pass also the icon or just the button.

planctus (changes requested):
Looking good, only small remarks ;)

emeryro:
For the buttons, I'm not sure that it is a real case. The ecl-print-default css is not supposed to be used when not printing (so with the "normal" ECL css), and if you have the ecl-print css it is displayed correctly (even with the ecl-print-default)

planctus (approved):

PR #2784, opened 2023-02-28
feat(blockquote): use picture tag - FRONT-3873

emeryro:
Use picture tag instead of background image

Notes:
- data structure has been updated, but a backward compatibility is provided


src/implementations/twig/components/picture/README.md
```diff
@@ -0,0 +1,69 @@
+# ECL Blockquote component
```
planctus: this is the blockquote's README :)
I would suggest to separate the prs, in any case, we are introducing a new "component", even though mostly transparent for the users, but still it feels it deserves its own enty in the changelog :)

planctus (commented):
As in the inline comment, i would create a pr first for the introduction of the picture template and a separate one for the adoption in the blockquote

planctus:
I will also add a test in the blockquote for the backward compatibility, i think this should always be the case

planctus (commented):
Question:
I wanted to raise this in the picture pr, but then i convinced myself not to, now that i look at the blockquote it seems relevant again.
Why do we nest the img data in the picture object in the picture template?
If that is for the easyness of passing a single object, we have to consider that most of the replacements in the different components will need this deprecation while if we keep the img data separated from the picture data we are basically only extending the existing components by extending the needed data.

emeryro:
I find it cleaner that way: we pass the picture object, which contains everything needed to render the image. It also mimic the markup structure.
Currently there are not two component using the same data structure for images (sometimes it requires a 'path', sometimes a 'src', sometimes just a string, ...). So in any case we would have to update the structure of existing components, and having element in a dedicated structure would make it easier for the backward compatibility (we don't have to adapt it to every existing data structure)

emeryro:
We could flatten the structure in the picture component, having directly sources and image at root level, if that would make things better.

planctus (approved):
It seems we have lost the show image control functionality, the image is shown by default and the control does nothing

PR #2785, opened 2023-03-01
chore(print): Global font-family set in default-print css - FRONT-3861

planctus:
Not sure this is really needed, the truth is that we should load the print css in the website, this is where the problem that originated this request was identified, this generic rule is not really going to be "fired", given ecl markup while it would affect "alien" markup, which is out of the scope of our stylesheets.


emeryro:
I think that we could put this rule in both the print and default-print css.
As you said, if the css is not loaded, it won't change anything. But at least we do what we can on ECL side...

emeryro (approved):

PR #2787, opened 2023-03-02
chore(content-block): Removing examples with tags - FRONT-3844

planctus:
I've only removed the examples, templates were not touched not to introduce a breaking change, should we add a "deprecation" message in the doc blocks of those templates..?


emeryro:
@planctus I think that we should mark the tags as deprecated indeed.
In the twig template, and the readme file

emeryro (approved):

PR #2789, opened 2023-03-06
feat(picture): add picture component - FRONT-3868

emeryro:
Add component to handle the display of images by using the html picture tag

### DIFF
--- a/src/implementations/twig/components/picture/picture.html.twig
+++ b/src/implementations/twig/components/picture/picture.html.twig
@@ -14,19 +14,39 @@
         },
         ...
       ]
-    - "picture_class" (optional) (string) (default: '')
-    - "image_class" (optional) (string) (default: '')
+    - "extra_classes" (optional) (string) (default: ''): Extra css classes, added to the root picture tag
+    - "extra_image_classes" (optional) (string) (default: ''): Extra css classes, added to to the img tag
+    - "extra_attributes" (optional) (array) (default: [])
+      - "name" (string) Attribute name, eg. 'data-test'
+      - "value" (optional) (string) Attribute value, eg: 'data-test-1'
 #}
 
 {# Internal properties #}
 
 {% set _picture = picture|default({}) %}
-{% set _picture_class = picture_class|default('') %}
-{% set _image_class = image_class|default('') %}
+{% set _extra_image_classes = extra_image_classes|default('') %}
+{% set _css_class = 'ecl-picture' %}
+{% set _extra_attributes = '' %}
+
+{# Internal logic - Process properties #}
+
+{% if extra_classes is defined and extra_classes is not empty %}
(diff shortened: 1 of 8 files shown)


src/implementations/twig/components/picture/picture.html.twig
```diff
@@ -0,0 +1,47 @@
...
+      - "sources" (array) (default: []): format: [
+        {
+          "src" (string) (default: ''): Path to the source image
+          "media" (string) (default: ''): Media condition to use this source. Could be a breakpoint ('s', 'm', 'l', 'xl') or a free string.
+          "type" (string) (default: ''): Type of this source
+        },
+        ...
+      ]
+    - "picture_class" (optional) (string) (default: '')
+    - "image_class" (optional) (string) (default: '')
```
planctus: our templates have extra_classes and extra_attributes by "default" forthe root element, why aren't we using them here..?

src/implementations/twig/components/picture/picture.html.twig
```diff
@@ -0,0 +1,47 @@
...
+        {% set _source = _source|merge({'media': '(min-width: 768px)'}) %}
+      {% elseif _source.media == 'l' %}
+        {% set _source = _source|merge({'media': '(min-width: 996px)'}) %}
+      {% elseif _source.media == 'xl' %}
+        {% set _source = _source|merge({'media': '(min-width: 1140px)'}) %}
+      {% endif %}
+      <source srcset="{{ _source.src|default('') }}" media="{{ _source.media|default('') }}" type="{{ _source.type|default('') }}">
+    {% endfor %}
+  {% endif %}
+  <img class="{{ _image_class }}" src="{{ _picture.img.src|default('') }}" alt="{{ _picture.img.alt|default('') }}" />
```
planctus: It seems we should add some checks here before rendering stuff, all these attributes might not have any value and we would render them empty.
It also feels that this data structure gets too structured in order to have a single object, the img object is nested into the picture one, but ok let's keep it like this.

src/specs/components/picture/demo/data.js
```diff
@@ -0,0 +1,27 @@
...
+        src: 'https://inno-ecl.s3.amazonaws.com/media/examples/example-image4.jpg',
+        media: 'l',
+      },
+      {
+        src: 'https://inno-ecl.s3.amazonaws.com/media/examples/example-image3.jpg',
+        media: 'm',
+      },
+      {
+        src: 'https://inno-ecl.s3.amazonaws.com/media/examples/example-image2.jpg',
+        media: 's',
```
planctus, line 23: weren't you defining a source with a custom media query before..? I think we should have that in this data we use the test the picture template

planctus (changes requested):

planctus (approved):

planctus (approved):

PR #2792, opened 2023-03-08
feat(tag): deprecate display tag - FRONT-3885

emeryro:
Display tag is deprecated.
Supported tags are link or removable


planctus (changes requested):
In the README there is:
(can be 'link', 'removable', 'display', 'button' (deprecated))
for the rest, all good

planctus (approved):

PR #2793, opened 2023-03-08
feat(gallery): Component updates, new video and audio icons - FRONT-3887

planctus:


src/implementations/twig/components/gallery/README.md
```diff
@@ -16,6 +16,7 @@ npm install --save @ecl/twig-component-gallery
   - "download" (object) (default: {}): object of type link
   - "share" (object) (default: {}): object of type link
 - **"items"** (array) (default: [])
+  - "playable" (boolean) (default: false) Activate the play icon on the thumbnail
```
emeryro: can't we detect this is parameter video or embedded_video is filled?

src/resources/icons-ec/src/all/audio.svg
emeryro, line 1: this change of icon may affect other components or pages
we should at least highlight it in the pr description

emeryro (commented):
Well done here!
Just a few remarks:

src/implementations/twig/components/gallery/README.md
planctus: this is the current check we have:
`{% if (_item.playable is not empty or _item.video is not empty or _item.embedded_video is not empty) `

The parameter is meant to give the chance to manually "trigger" the play icon, could be removed in case we don't expect any use case where this would be used

emeryro: I don't really see a case where we would want to have this play icon if this is not a video.
We could keep it, but I fear that it may be confusing for users (we should at least make it clear in the comment/readme)

planctus: it is foreseen for the audio files as well, but in the end this would still be an "embedded_video" so i think we can safely remove this parameter...

emeryro (approved):

PR #2795, opened 2023-03-09
feat(language-list): accessibility improvements - FRONT-3890

emeryro:
- improve keyboard navigation (language list opens with space or enter)
- add role button to the language switcher
- update aria label to specify current language
Note: this change has been done in the ECL example, but it would require work on implementation to have dynamic content
- add lang and hreflang information to each language link
- fix contrast issue


planctus (dismissed):

planctus:
I think we can merge this, @emeryro, we just need to raise the limit for the js size ;)

PR #2797, opened 2023-03-13
fix(banner): fix right to left

emeryro:
Fix display of full width banner when using right to left


planctus (commented):
I noticed this bug while reaching the playground from the website, nothing related to this pull request but still worth attention:
image

I guess it depends on the complex way the story markup and the data are processed, i would move the correctPaths() call to the prepareData() function, separating the data processing from the rendering of the markup. To be tested whether this would fix the current issue, though.

emeryro:
I created another ticket for the issue you reported

planctus (approved):

PR #2798, opened 2023-03-14
feat(menu): advanced overflow management - FRONT-3889

emeryro:
Add an overflow management for the menu.
It consist in 2 elements:

- Menu items can be resized automatically, to be displayed in "n" lines (2 lines by default, configurable). This ensures a more harmonized display
- If there is not enough space to display all the items (after the modification done before), an overflow button is diplayed to split the items in 2 "pages"

Notes:
- data structure of menu items hasn't changed (only new buttons to manage the overflow)
- quite heavy javascript updates...


src/implementations/vanilla/components/menu/menu.js
```diff
@@ -367,13 +473,165 @@ export class Menu {
   }
 
   /**
-   * Check for a specific menu item how to display the mega menu
+   * Check how to display menu horizontally and manage overflow
+   * @param {Node} menuItem
```
planctus: there is no parameter for this method

src/implementations/vanilla/components/menu/_menu.scss
```diff
@@ -521,6 +529,46 @@ $_title-font-weight: null !default;
       background-color: transparent;
     }
 
+    // Previous / Next items
+    .ecl-menu__items-previous {
```
planctus, line 533: it all looks incredibly complex, so this might look as an irrelevant thing, but i wonder if we couldn't avoid this situation where the item "button" gets kind of "attached" to the next items button:
image

src/implementations/vanilla/components/menu/menu.js
```diff
@@ -610,7 +873,79 @@ export class Menu {
   }
 
   /**
-   * Click on a menu item
+   * Click on the previous items button
+   */
+  handleClickOnPreviousItems() {
```
planctus, line 877: It seems we introduced an issue with the keyboard navigation of the menu when items are hidden:
you can move to the hidden elements using both tab or arrow right but there is no way to go back, or better, using left arrow navigates the list backward but visually there is no change and only the previously hidden items are visible.

src/specs/components/menu/demo/data--fr.js
```diff
@@ -7,11 +7,13 @@ module.exports = {
...
   back: 'Retour',
   menu_link: exampleLink,
+  button_previous_label: 'Previous items',
+  button_next_label: 'Next items',
   icon_path: '/icons.svg',
   items: [
     { label: 'Accueil', path: exampleLink },
     {
-      label: 'Item 2 titre',
+      label: 'Réglementation, surveillance financière',
```
planctus, line 16: wouldn't it be better to have plausible labels also for the "default" demo..? Also because with the current labels we have we don't demo the new features of the menu

src/implementations/vanilla/components/menu/_menu.scss
```diff
@@ -127,6 +128,10 @@ $_title-font-weight: null !default;
       z-index: map.get(theme.$z-index, 'modal');
     }
 
+    .ecl-menu__inner--has-overflow {
+      overflow-x: clip;
```
planctus, line 132: We do have a vertical scrollbar, now...
image

planctus (changes requested):
Wow, this is pure madness, we are handling the number of lines of the menu items exposing that as an option for the user :)
It is all so complicated that it's really hard even to look at the code, reviewing it for real would take weeks.
I noticed a few things, but also tested the functionalities introduced which seem effective, thanks for the effort, i just hope this thing won't explode in our hands at some point with a thousand of lines of js to handle all the requirements..

src/specs/components/menu/demo/data--fr.js
emeryro: My idea was to have the default menu fitting correctly in the screen, to showcase a "good" example, and then a longer one with overflow
I kept the "default" and "translated", maybe I can just rename the stories

src/implementations/vanilla/components/menu/_menu.scss
emeryro: I don't really see what we could do to avoid that.
The next/previous buttons are displayed on top of the menu (when needed), so in some cases the last item would be partially hidden below it, sometimes it would go as on your screenshot where most of it is displayed. But that's not really something that we can control

src/implementations/vanilla/components/menu/_menu.scss
emeryro: That's already the case in the prod version, if there is not enough vertical space to display the hidden sub menu (I assume that you had the inspector open or the storybook control positioned on the bottom).
image

src/implementations/vanilla/components/menu/menu.js
emeryro: that's true, it is supposed to move the menu back to the left when going backward. I'll check

src/specs/components/menu/demo/data--fr.js
emeryro: Stories renamed. I kept the spec file name as is, as they are also in other component (including deprecated ones)

planctus:
there is small issue introduced in the keyboard navigation, when accessing the dropdown it is possible to browse the items using arrow down but the "default" is not prevented and there is a scroll in the body

planctus:
i understand that we have this since the previous versions, but it's the delay of the css transition when hovering on the items really needed..? To me that looks like a bug, you see the link getting underlined and after a bit everything gets the white background..not really sure why are doing that.

src/implementations/vanilla/components/menu/menu.js
```diff
@@ -610,7 +872,79 @@ export class Menu {
...
+    if (!this.itemsList || !this.btnNext) return;
+
+    this.offsetLeft = 0;
+    if (this.direction === 'rtl') {
+      this.itemsList.style.right = '0';
+      this.itemsList.style.left = 'auto';
```
planctus, line 883: why don't we move the styles in the css and we work adding/removing selectors in the js, instead? I think it would be cleaner, both for the "vanilla" component per se but also for the implementation when our vanilla js is not being used, it's easy to see a selector being added/removed and you don't have to be bothered about setting inline styles

src/playground/ec/.storybook/preview-head.html
```diff
@@ -8,6 +8,11 @@
   integrity="sha512-qTXRIMyZIFb8iQcfjXWCO8+M5Tbc38Qi5WzdPOYZHIlZpzBHG3L3by84BBBOiRGiEb7KKtAOAs5qYdUiZiQNNQ=="
   crossorigin="anonymous"
 ></script>
+<script>
```
planctus: why this..? Or, i mean, it's true that we were not handling the js fallback in storybook but as long as we don't add the no-js class to the html tag in the storybook iframe we are just adding an unused class "has-js" when the script is run...

planctus (commented):
To me it's pretty much ok, we went beyond a level of reasonability in terms of the amount of code we need to fullfill all the requirements and i don't want to imagine the reactions by those who will have to implement this incredibly complex component without using the resources we provide, but that's it..
I left a few comments to your evaluation, none of them i would consider it "blocking", anyhow.. ;)

src/implementations/vanilla/components/menu/menu.js
emeryro: that would work for cases where we reset values to 0 or auto, but for all the dynamically calculated values we would still have to add them via js.
Having a mix between classes and inline values would be even harder to handle I think

src/playground/ec/.storybook/preview-head.html
emeryro: that's what is defined in our guidelines (https://ec.europa.eu/component-library/ec/getting-started/), so I thought it would be nice to be able to test the no-js behavior when needed

planctus: ok, but we need to add the no-js class to the html tag in the storybook iframe for this to happen, in this script we are removing a class that we never set, that's what i'm saying..

planctus: i'll try to expain this better:
Our components have, sometimes, styles based on the no-js class.
What we do in the website, and what we "recommend" the user to do, is to add this class to the html tag of their pages for then replacing that (but it's irrelevant the replacement, it could be simply removed, since we only use the no-js class..) once the dom is ready.

If we want to reproduce this in storybook we would at least need to have the no-js class set by default on the html tag in the storybook iframe, but at the same time storybook would not work with javascript disabled, so in the real world we should provide ourselves a way to only disable the ecl js in storybook for giving the users the chance to test the "no js" behaviour of our components.
(If we had an opt-in approach for the autoInit in our templates this would be easier, it would be enough to add the no-js class on the #root wrapper when the user disables the auto-init, but it's not the case..)

My proposal is to simply remove this, for the moment, it is completely useless, if we want we can think about how to offer this option to the users in the future.

emeryro: Ok I removed it for now

planctus (approved):

PR #2800, opened 2023-03-15
feat(form-elements): Using generic sibling selector to allow slightly different markup - FRONT-3857

planctus:


emeryro (commented):
It seems to work fine with "~"
Shouldn't we consider updating alos other css still using "+"? For instance the select and radio

planctus:
Ok, let's update all the form elements with this loose approach, it was requested for the checkbox (in swift they get something in between the label and the input) but from our point of view the same approach can and should be used everywhere

emeryro (approved):

PR #2803, opened 2023-03-20
chore: Upgrading mdx to v2, React 18

planctus:
There is one drawback, line breaks inside table cells do not work and it doesn't seem there is an alternative syntax that would allow it.


emeryro (changes requested):
It seems to work fine. I just noticed a few things:
- First paragraph missing in float utilities
image

- Wrong display on guidelines spacing
image

emeryro (approved):

PR #2804, opened 2023-03-20
feat(list-illustration): Using picture tag - FRONT-3877

planctus:
Some checks have been added about the "emptyness" of the two objects picture and image, the twig "empty" returns false when testing an empty object, indeed


emeryro (changes requested):
The squared images are deformed. I guess that we should use `object-fit: cover` in this situation

emeryro (approved):

PR #2807, opened 2023-03-20
feat(file): Using picture - FRONT-3875

planctus:


src/implementations/twig/components/file/file.story.js
```diff
@@ -210,10 +210,11 @@ const prepareData = (data, args) => {
     data.description = '';
   }
 
-  if (args.show_image && data.image) {
-    data.image.src = args.image;
-  } else if (!args.show_image && data.image) {
-    data.image = {};
+  if (args.show_image && data.picture) {
+    data.picture.img = {};
```
emeryro: that would remove every other parameter for the image, no? (the alt text mostly)
So either we handle it manually, or we could remove the picture parameter when not used, as on https://github.com/ec-europa/europa-component-library/blob/v3.8.0-dev/src/implementations/twig/components/blockquote/blockquote.story.js#L52

emeryro (changes requested):

src/implementations/twig/components/file/file.story.js
planctus: you are right, i fixed this but at the same time i find it crazy the way we handle this data in order to offer controls for everything, it adds complexity where it would not really be needed, we should step out from this approach, in my opinion, for ECL v4

emeryro (approved):

PR #2810, opened 2023-03-23
feat(site-header): Using picture, keeping data structure unchanged - FRONT-3882

planctus:
Use picture tag to handle logo

Note: for EU, the change of logo between mobile and desktop is now handled with the picture source

### DIFF
--- a/src/implementations/twig/components/site-header/__snapshots__/site-header.test.js.snap
+++ b/src/implementations/twig/components/site-header/__snapshots__/site-header.test.js.snap
@@ -4763,10 +4763,6 @@ exports[`Site Header EU renders correctly 1`] = `
                 media="(min-width: 996px)"
                 srcset="/standard-version/positive/logo-eu--en.svg"
               />
-              <source
-                media="(min-width: 480px)"
-                srcset="/condensed-version/positive/logo-eu--en.svg"
-              />
               <img
                 alt="European Union flag"
                 class="ecl-site-header__logo-image"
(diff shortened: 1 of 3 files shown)


emeryro (commented):
Coulnd't we take advantage of the picture sources to group the mobile and desktop logo in the same tag?

planctus:
Fascinating, i did this on my own and then i read your comment.. :)

src/implementations/twig/components/site-header/__snapshots__/site-header.test.js.snap
```diff
@@ -4723,18 +4755,24 @@ exports[`Site Header EU renders correctly 1`] = `
...
+            <picture
+              class="ecl-picture ecl-site-header__picture"
               title="European Union"
-            />
+            >
+              <source
+                media="(min-width: 996px)"
+                srcset="/standard-version/positive/logo-eu--en.svg"
+              />
+              <source
```
emeryro: this second source is not needed. By default it will take the image in the img tag (mobile) and switch to desktop

src/implementations/vanilla/components/site-header/site-header-eu.scss
```diff
@@ -78,16 +78,12 @@ $_search-width-xl: 31.5rem;
   flex-grow: 1;
 }
 
-.ecl-site-header__logo-image-mobile {
```
emeryro, line 81: if we remove the previous css class definition, that would be a breaking change
So we should either keep them (deprecated) or flag the pr as breaking

emeryro (changes requested):

emeryro (approved):

PR #2814, opened 2023-03-28
feat(media-container): Using picture - FRONT-3878

planctus:
Use HTML picture tag


src/implementations/twig/components/media-container/media-container.test.js
emeryro, line 1: we should add a test with the previous data structure

emeryro (changes requested):
Just some test update

src/implementations/twig/components/media-container/media-container.html.twig
```diff
@@ -1,8 +1,8 @@
 {% apply spaceless %}
 {#
   Parameters:
-  - "alt" (string) (default: ''),
   - "image" (string) (default: ''),
```
emeryro: having both image and picture is quite confusing.
I think we should at least document it properly in the comment (maybe having a separate section for video)

src/implementations/twig/components/media-container/README.md
```diff
@@ -20,13 +20,18 @@ npm install --save @ecl/twig-component-media-container
...
 - **"ratio"** (string) ('') Ratio of the embedded media, if empty the ratio will be set by the js
 - **"extra_classes"** (optional) (string) (default: '') Extra classes (space separated)
 - **"extra_attributes"** (optional) (array) (default: []) Extra attributes
+
   - "name" (string) Attribute name, eg. 'data-test'
   - "value" (string) Attribute value, eg: 'data-test-1'.
 
+  Deprecated:
+
+  - **"alt"** replaced by card.picture.img.alt
```
emeryro:
```suggestion
  - **"alt"** replaced by picture.img.alt
```

emeryro (approved):

PR #2817, opened 2023-03-29
feat(card): Using picture - FRONT-3874

planctus:
Use picture tag instead of background image


src/implementations/twig/components/card/__snapshots__/card.test.js.snap
emeryro, line 1: we should add the usual test with previous data structure

emeryro (changes requested):

src/implementations/twig/components/card/__snapshots__/card.test.js.snap
planctus: we have that already, @emeryro 
```
  describe('Backward compatibility', () => {
    test('renders correctly with deprecated data', () => {
      expect.assertions(1);
      return expect(render(dataCardBc)).resolves.toMatchSnapshot();
    });
  });
```

emeryro (approved):

PR #2818, opened 2023-03-30
feat(page header): use picture tag - FRONT-3880

emeryro:
Replace thumbnail and background images tag by a picture


planctus (commented):
Not related to this pull request, but i see in the usage page of the component that we always use and specify a 1/1 ratio for the thumbnail, but this was never enforced, apparently, i wonder whether it's the implementation that needs to be updated or the usage pages..

src/implementations/twig/components/page-header/page-header.html.twig
```diff
@@ -65,6 +70,17 @@
   {% endfor %}
 {% endif %}
 
+{# Backward compatibility #}
+
+{% if _picture.img is empty and _thumbnail.src is not empty %}
+  {% set _picture = {
+    'img': {
+      'src': _thumbnail.src|default(''),
```
planctus: you check for the emptyness of the thumbnail.src already, why setting its default here..?
I see that in the picture template we check for for an empty "alt" before rendering the attribute, so we can keep this structure as it is, although we might be passing an empty value for the alt property.

emeryro:
We will check with COMM for the image ratio

planctus (commented):
Why aren't we replacing also the background image in the page header with the picture @emeryro ..? It seems a much more relevant use case than the thumbnail itself for using it.

src/implementations/vanilla/components/page-header/page-header-print.scss
```diff
@@ -47,3 +47,9 @@
   margin-bottom: 0;
   margin-top: map.get(theme.$spacing-print, 'm');
 }
+
+.ecl-page-header--image {
```
planctus, line 51: but wasn't the possibility to have these images in print one of the reasons to adopt the picture instead of a background image..?

src/implementations/vanilla/components/page-header/page-header-ec.scss
```diff
@@ -138,14 +138,15 @@
...
+  }
 
   .ecl-page-header__background {
-    background-position: 50% 0;
-    background-size: cover;
     display: block;
     height: 10.875rem;
-    position: relative;
+    object-fit: cover;
+    object-position: 50% 0;
```
planctus, line 149: mmh..i understand we want to reproduce the current "look and feel" with the image used in the demo, but this property seems only meaningful in this specific use case, while there is no clear "requirement" or guildeline about the image to be uploaded and the portion of it that will be shown by the browser.

src/implementations/vanilla/components/page-header/page-header-ec.scss
```diff
@@ -169,7 +170,7 @@
 }
 
 .ecl-page-header--overlay-dark {
-  .ecl-page-header__background::before {
+  .ecl-page-header__background-container::before {
     background-color: rgba(map.get(theme.$color, 'black'), 0.5);
```
planctus, line 174: It seems we could set this overlay effect with a single css property, mix-blend-mode, https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode, instead of using a pseudo element absolutely positioned, it's not supported in internet explorer, though, in case we are still bothered by that.
Also relevant for the banner pr, this property has been used for the "image-overlay" variant

emeryro: I tried it, but without success here.
As there is a blue background on the whole element, the overlay ended up being blue too. I'll double check

emeryro: @planctus I confirm that mix-blend-mode does not work as we would like to.
On the banner too, the result is not correct. It takes into consideration all the background set on the element, so the end result is something odd and depending on the background colors (this is clearly visible on the EU banner in your PR)

src/implementations/vanilla/components/page-header/page-header-ec.scss
emeryro: If I remember well, that was requested at some point to center the image vertically. 
Maybe in the future we could offer a paramter to change the image position, but for now I think that the safest move is to keep what we have currently (visually). Otherwise that may considered as a breaking change

src/implementations/vanilla/components/page-header/page-header-ec.scss
planctus: ok, i will change this in the banner as well..

planctus (approved):

PR #2819, opened 2023-03-30
feat(banner): Using picture instead of the background image - FRONT-3872

planctus:


src/implementations/vanilla/components/banner/_banner.scss
```diff
@@ -107,28 +112,28 @@ $_font-weight: null !default;
...
-  .ecl-banner::before,
-  .ecl-banner--m::before {
+  .ecl-banner,
+  .ecl-banner--m {
+    .ecl-banner__image {
+      aspect-ratio: 4/1;
+    }
+  }
+
+  .ecl-banner--plain-background::before {
```
emeryro: do we still need this? I think that it was for the ratio

src/implementations/vanilla/components/banner/_banner.scss
```diff
@@ -184,18 +189,12 @@ $_font-weight: null !default;
 .ecl-banner--text-box {
   background: map.get($_background, 'text-box');
 
+  // manage ratio 4:1
   .ecl-banner__image {
+    aspect-ratio: 4/1;
```
emeryro, line 147: this aspect ratio overrides the one from the banner size, so changing size does not work anymore

emeryro (changes requested):
Image is not cropped the same way as before (both desktop and mobile)
image
image

src/implementations/vanilla/components/banner/_banner.scss
planctus: But nothing has changed regarding the plain-background variant, there is no image so the aspect-ratio is still set by this css hack.

planctus:
@emeryro , we replaced a background image with an img tag, this change in the way the image is rendered is "intended", now we use object-fit, before we had a background-size, they do not behave the same.
I've added a object-position to reproduce the previous behaviour, but i'm not really confident this is what we should do for every image.

emeryro (approved):

PR #2820, opened 2023-03-31
feat(gallery): Using picture - FRONT-3876

planctus:
Use picture tag for image gallery and overlay

### DIFF
--- a/src/implementations/twig/components/gallery/__snapshots__/gallery.test.js.snap
+++ b/src/implementations/twig/components/gallery/__snapshots__/gallery.test.js.snap
@@ -843,17 +843,24 @@ exports[`Gallery Default renders correctly with deprecated data 1`] = `
         class="ecl-gallery__item"
       >
         <a
+          aria-label="Image 2"
           class="ecl-gallery__item-link"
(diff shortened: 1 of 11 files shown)


src/implementations/twig/components/gallery/__snapshots__/gallery.test.js.snap
```diff
@@ -102,19 +843,16 @@ exports[`Gallery Default renders correctly 1`] = `
         class="ecl-gallery__item"
       >
         <a
-          aria-label="Image 2"
           class="ecl-gallery__item-link"
           data-ecl-gallery-item=""
           data-ecl-gallery-item-share="/share#example-image2.jpg"
-          href="https://inno-ecl.s3.amazonaws.com/media/examples/example-image2.jpg"
+          href=""
```
emeryro: I don't think that this href has been emptied on purpose

emeryro (changes requested):
A few remarks:

- the documentation in gallery.html.twig should be updated (item structure)
- the gallery overlay is still using single img. Shouldn't we use the picture here too?

emeryro (approved):

PR #2821, opened 2023-03-31
feat(description-list): Adding option and logic to hide/show items in  lists - FRONT-3895

planctus:
as requested standaone links, inline links, taxonomy list items can be limited by setting a common parameter when including the template.
.
A control has been added in the story as a range, by default the vertical example is having collapsed lists while the horizontal doesn't.


src/implementations/vanilla/components/description-list/description-list.js
```diff
@@ -0,0 +1,136 @@
...
+    this.moreItemLabel = this.element.getAttribute(this.moreItemLabelSelector);
+    this.visibleItems = this.element.getAttribute(this.visibleItemsSelector);
+    this.lists = queryAll(this.listsSelector, this.element);
+
+    // Add see more button in each list and bind click event on it
+    if (this.lists[0] && this.visibleItems > 0 && this.moreItemLabel) {
+      this.lists.forEach((list) => {
+        if (list.children && list.children.length > this.visibleItems) {
+          this.showHide(list.children);
+          const button = document.createElement('a');
```
emeryro, line 69: wouldn't it be better to use a real button instead of a link (without href)?

src/implementations/twig/components/description-list/description-list.html.twig
```diff
@@ -67,7 +80,12 @@
 
           {# taxonomy #}
           {% if _item.type is not empty and _item.type == 'taxonomy' %}
-            <dd class="ecl-description-list__definition ecl-description-list__definition--taxonomy">
+            <dd
+              class="ecl-description-list__definition ecl-description-list__definition--taxonomy"
+            {%- if _visible_items > 0 -%}
+              data-ecl-description-list-collapsible
```
emeryro, line 86: does it count as markup change?

src/implementations/twig/components/description-list/description-list.html.twig
```diff
@@ -11,6 +11,8 @@
         ...
       ]
     - "variant" (string) (default: ''): global variant of the list (horizontal, vertical)
+    - "visible_items": (integer) (default: 0)
+    - "more_label": (string) (default: '')
```
emeryro:
```suggestion
    - "more_label": (string) (default: '') Label for the more button, if requested
```

src/implementations/twig/components/description-list/description-list.html.twig
```diff
@@ -11,6 +11,8 @@
         ...
       ]
     - "variant" (string) (default: ''): global variant of the list (horizontal, vertical)
+    - "visible_items": (integer) (default: 0)
```
emeryro:
```suggestion
    - "visible_items": (integer) (default: 0) Number of visible items (0 to display all items)
```

src/implementations/vanilla/components/description-list/description-list.js
planctus: i changed this in the last commit, it is because of the styles that the buttons have, that were a bit problematic and would sure requiremore iterations on the design side to tweak it.

src/implementations/twig/components/description-list/description-list.html.twig
planctus: not really, to be honest, we call "markup change" something that needs to be done at the markup level, here eUI will entirely rebuild this functionality, not using our attributes since they don't use our js..
We can still add the label to raise attention on a functionality that they will have to "implement", though

src/implementations/vanilla/components/description-list/description-list.js
emeryro: Ok, to be seen if that doesn't raise accessibility issues later (having a link without href is usually a bad practice)
Also the css has to be updated in this case, to have the hand cursor on hover (it isn't added if there is no href)

planctus: i added the href.. ;) a link is still valid without an href but without it we would not be able to make the keyboard navigation work as expected without additional coding, so i simply added it

emeryro:
Now that we can focus the link, I noticed that it is displayed wrongly, with the click area going outside of the the text.
image

planctus:
thanks @emeryro, the focusable area should be ok now

emeryro (dismissed):

emeryro (approved):

PR #2840, opened 2023-04-20
feat(list with illustration): add value, image size and alignment - FRONT-3929

emeryro:
- Optional value, to display metrics
- Different image sizes for square image and icons (regular images are not affected)
- Optional centered alignment for horizontal list with illustration


planctus (commented):
Looks great, but i'm not sure about the print version, it seems that before when we had columns we were ignoring that when printing, now that we introduced images we might want to keep the "layout", but it's a bit risky considering that we don't know the size of the printable surface. I see that it's possible to use viewport units to try to detect the size of the medium in a media query for print, but it's maybe just safer to do what we were doing before.

planctus (dismissed):

PR #2846, opened 2023-04-26
feat(styles): add ECL4 styles - FRONT-3932

emeryro:
Apply new styles for ECL EC. There is still improvements and cleanup to be done at a later stage, but this should allows us to start working on components.

Notes:

- theme structure has been slightly update, to have sass maps defined in their own files
- utilities have been updated, to match new possible values
- website documentation hasn't been changed. To be seen with COMM later what we want to display or not
- a documentation has been started to provide help and highlights for the migration to ECL 4


src/themes/ec/maps/shape.scss
```diff
@@ -0,0 +1,18 @@
+// For some reasons, it does not work to redefine this color into the "with"
+$shadow-color: #26324b;
+@forward '@ecl/theme-dev/maps/shape' with (
+  $shadow: (
```
planctus, line 4: why do we have this different approach in EC? it's not really intuitive the naming of the resulting utilities, with values like ecl-u-shadow-16, especially in comparison with the more "reasonable" scale we have in EU.
Furthermore, it seems that only these "shadow" should be defined for EC while the way we are currently doing it makes all the unused values in EC generate utilities.
Wouldn't it be better to have the same "levels" in EC and EU? And shouldn't we only generate the needed utilities in the two systems or demo then the same way in the two systems, otherwise?

src/themes/dev/maps/color.scss
```diff
@@ -1,5 +1,5 @@
-// EC colours are used by default,
-// EU colours are set in corresponding themes
+// EC and EU colours are set in corresponding themes
+// TODO: Colors here should be removed once update is finish
```
planctus, line 2: From what i understand we could already remove this file and the relative forwarding and usage by the themes, no..? It is confusing to have the "old" and "new" approach together, but i'm not really convinced by this additional forwarding we have in the themes, we are creating files and an additional layer doing it. I think it would be beneficial only in case we are about to remove the "dev" theme as a common ground between the two systems and then we would define these maps in the themes directly.

.size-limit.json
```diff
@@ -1,7 +1,7 @@
 [
   {
     "path": "dist/packages/ec/styles/ecl-ec.css",
-    "limit": "351 KB",
+    "limit": "377 KB",
```
planctus, line 4: but in the end is this increased size justified? Where does it come from..?

src/themes/ec/maps/color.scss
```diff
@@ -0,0 +1,101 @@
...
+    'accent30': #887de8,
+
+    // Other colors
+    'branding': #254acb,
+    'black': #000,
+    'white': #fff,
+
+    // TODO: Old EC colors; to be removed once components are updated
+    // semantic colours
+    'tertiary': #e3e3e3,
```
planctus, line 57: i think we miss the "info" here, which is used in message and modal..

planctus (commented):
Left a few comments around..

src/themes/ec/maps/shape.scss
emeryro: That's because the designers wanted to have the possibility to add other intermediate level later if needed
image
And indeed, currently there are unneeded utilities generated, but this will be cleaned-up when the EC components are updated.
In the end, I assume that the same approach will be applied to EU later (after 4.0...)

src/themes/dev/maps/color.scss
emeryro: I agree that we are in an "in-between" currently.
I tried to remove part of the dev theme already, but it used also in the website, and I didn't want to go too far for the first step. The idea was to have something usable quickly, to start working on the components, and go back to the styles later.

So yes, later we could remove this dev theme, or maybe have a "neutral" one, similar to what has been done one OEL

.size-limit.json
emeryro: I guess that it comes from the utilities.
We introduced new values (and so new utilities), but we are not able to remove previous values yet.
The size should be reduced once we can clean up the styles

src/themes/ec/maps/color.scss
emeryro: that's true. Info color is not defined in the new color palette, so I don't know how the designers expect the messages to be displayed, but I'll add it in the old colors

PR #2860, opened 2023-05-09
fix(select-multiple): Fixing unwanted submit on click on buttons  - FRONT-3948

planctus:


emeryro (changes requested):
It seems that the issue is still there.
I tested by adding a form wrapper in the story: 
``export const Multiple = (args) => `<form action="example">${selectBox(prepareData(dataMultiple, args))}<input type="submit"/></form>`;`` 

When I click on "close" for instance, the form is sent
select

planctus:
Question: did you run yarn before making this test..? :)

emeryro:
Ok, it's working fine :)

emeryro (approved):

PR #2861, opened 2023-05-10
fix(menu): Sub items not displayed - FRONT-3949

emeryro:


planctus:
For the sake of this bug we introduced, this is fixing it but my gut feeling is that this javascript should be entirely refactored, we strongly rely on an html attribute value, repeating over and over the same checks, it's out of control and with a certain suprise i've jst realised that this menu won't simply work without javascript:
we do make an attempt, with our "no-js" class that we are not setting ourselves, but that's not going to work. I was still thinking that the opening of the dropdown was driven by css but we are actually only relying on javascript and only on mouse events, not touch events, so this would not work on an ipad, like..

planctus (approved):

PR #2864, opened 2023-05-11
feat(link): ECL4 - link - FRONT-3934

emeryro:
Apply new styles to links and provide variables for customization


planctus (commented):
do we have to go that deep with custom properties...?
I mean, what is that we want to achieve, making everything customisable? Do we need that? And then i wonder, if we do this for colors, rounded corners, outlines, borders, then i wonder why not doing it for spacing as well. But the reality is that it feels too much and potentially dangerous, or questionable, the fact that we define dozens of custom properties for this component but we are actually exposing them in the root of the document, so nothing gets really scoped into the link.

It also feels that this might be something we want to have the time to evaluate, but if we start like this we are already fully into something that will then grow and grow indefenitely, i agree in trying to do something more than just a "revamp" of EC, and custom properties were something "on the table", but i didn't really think we would do everything with them.

I like the way you organised all the different properties in the different variants with  a map, but the resulting syntax it's very very complex, it is already complex because of the way to retrieve values from a map (and i never liked the fact that we are retrieving the values from the maps directly), but nesting keys into a single map object makes it even harder. Could we turn these into flat maps under a key already identifying them as the "variant" where to be applied?

I don't necessarily have strong opinion about the usage of custom properties, but to me this should be limited to a set of properties that you could use to alter the "theme", more than each element of any component.

Anyway, just a few thoughts, we both miss experience using this approach, that is also why i would probably not dive into that suddenly, but i feel that we could introduce "some" custom properties, and evaluating which ones, but not any kind of color we use needs to be customisable element by element, really..

emeryro:
I agree with you: there are lots of css variables here.
But unfortunately, that's the one we need already on our side to differentiate EC and EU. So we need to have some variables for that. We may not convert every one of them to css variable, and keep some scss variable for that, but we need them as variable anyways.

The main pain point is the fact that we have to pass the sass variable to the component (see [here for instance](https://github.com/ec-europa/europa-component-library/pull/2864/files#diff-6026893efac2ee522cbfcb31284dd41f4a251619d43356504d7396c2b94148d9R9)). At first I used flat variable, but then I had to list all of them in the different files, that's why I opted for the nested approach. 
If we want to keep some scss variables for links, it would probably be better to group them in a nested structure too. So unless we find another way, I don't think that we could really simplify the map structure.

I'll do some test to reduce the amount of exposed variables, and use directly the $link map

emeryro:
I removed the css variables for now.
Once we have the global theme parameter available, we will see what should be introduced or not at component level

PR #2866, opened 2023-05-16
feat(gallery): Overlay iteration - FRONT-3918

planctus:


emeryro (commented):
Not sure if this is related to this update, but there are some strange display and behavior on my Firefox

blur issue:
image

close issue:
gallery

planctus:
the blur, although i find it's an ugly thing, is there because requested by the designers, for the close button the issue is..? that the outline it's not fully visible..?
image
This is how it looks on firefox for linux..

emeryro:
For the close, each time you close the overlay (on FF), the gallery grid is modified (the previously selected image takes full width).

planctus:
Oh, true..it was in the picture pr where i guess this was introduced, fixed.

emeryro (approved):

PR #2868, opened 2023-05-17
fix(media-container): Fixing control, adapting labels - FRONT-3956

planctus:


emeryro (dismissed):
This control works fine now.
I have a question here: the "ratio" control only works for iframe, I don't recall if that was on purpose or not, but as it is now it is misleading (the control is displayed in the other stories but does nothing)

planctus:
I guess that it is only meant to affect the embedded media, the ratio param is only taken into account there:
```
{% if _embedded_media is not empty %}
  {% set _embed_class = _ratio ? "ecl-media-container__media ecl-media-container__media--ratio-" ~ _ratio : "ecl-media-container__media" %}
```
I moved the control to the story where it is supposed to be used, then.. ;)

emeryro (approved):

PR #2869, opened 2023-05-22
feat(site-header): change icon - FRONT-3950

emeryro:
Replace icon, and deprecate language code display


planctus:
image
there is an alignment issue on EU

planctus (approved):

PR #2876, opened 2023-05-24
fix(select): Randomising generation of default id for select all checkbox - FRONT-3966

planctus:


emeryro (commented):
Using a random number should work indeed, but don't we already have a unique id for the select field, to be reused here?

planctus:
i'd say that nothing is  really "unique" here, not even with this random string appended.. :)
what we currently do for all these checkboxes is to create an id starting from 'select-multiple' by default and for the standard ones we add the value of the option so it becomes select-multiple-{value}, we don't set any id for the select by ourselves, instead.

By the way this id we start from is an option of the js, so potentially the users could fix on their side, but i think that autoInit is always used for everything on the implementation side, so better doing this on our side.

emeryro (approved):

PR #2877, opened 2023-05-24
fix(file download): fix description display - FRONT-3961

emeryro:


src/implementations/vanilla/components/file/_file.scss
```diff
@@ -123,11 +123,8 @@ $_image-width: null !default;
 
 .ecl-file__description {
   color: $_file-description-color;
-  display: inline-flex;
```
planctus, line 126: ok this "fixes the bug", why removing also the other properties? For the max width that still seems relevant in case of no image and for the white-space handling, i don't know what it was meant for, but i saw that it was introduced when the rtl was implemented, didn't notice any problem without it but there was probably a reason why it was put in place.

planctus (approved):

PR #2884, opened 2023-05-30
feat(button): ECL4 button - FRONT-3945

emeryro:
Apply new style and variable structure for buttons


src/themes/ec/_variables.scss
```diff
@@ -4,6 +4,80 @@
 // Use this file to assign default values to component variables
 // Note: if you override css variables it will take over on these values
```
planctus, line 6: the question is: why defining all these variables at the theme level? The variable are kind of scoped for each component, then why do we define them outside of the component definition? I know we did this for the link already, but we might re-think this a bit: 
from what i understand here we are trying to reduce to a single file the scss for a component, but we are just moving it in the theme this "logic" and i don't think it's the right place, i would consider having two variables files in the component that we would collect then in a "global" variables index to be used by the two presets, instead. 

This way we are storing the variables in a single file, so imagine building a single component css (what we do in webcomponents), in this case we would need to "use" all the variables at once for compiling a single scss file, furthermore this file is going to grow indefinitely if it has to store all the variables for all the components and maintaining it it's going to be pretty unhandy.

emeryro: My idea was to centralize the variables used to customize the components in the theme, so users do not have to update files in each components. All the customization, variable assignment etc could then be kept in a single folder instead of having them spread everywhere. That way the components would be theme agnostic.

Also as there will most probably be some custom themes apart of EC and EU (for EWPP extended), having to create a file in each component would be quite heavy.

I agree that this variables file may grow too much, but if that's the case we could split it in several files (still in the theme folder for me)

PR #2888, opened 2023-06-01
feat(button): improve display of button with icon only - FRONT-3976

emeryro:
add new variant .ecl-button--icon-only to display square button with only the icon


planctus (approved):
Nice the removal of the before/after classes!

docs/Migrating-v4.md
```diff
@@ -57,16 +57,19 @@ Spacing scale has been enriched for EC, now going from 2XS to 6XL (previously 2X
 
 ### Button
 
+- Classes `ecl-button__icon--before` and `ecl-button__icon--after` are no longer needed for the icon; position is detected using CSS.
 - To keep consistency, CSS class for call to action button has been renamed from `ecl-button--call` to `ecl-button--cta`
 - `ghost`button is now called like that everywhere (it was sometimes called `text` previously). CSS classes haven't been modified
+- New variant added to handle button with icon only: `ecl-button--icon-only`. By using it, it is no longer needed to add utility classes on the button label. If you are using the twig template, this variant rely on the already existing parameter `hide-label`
```
planctus:
```suggestion
- New variant added to handle button with icon only: `ecl-button--icon-only`. By using it, it is no longer needed to add utility classes on the button label. If you are using the twig template, this variant relies on the already existing parameter `hide_label`
```

src/website/src/pages/ec/components/button/docs/code.mdx
```diff
@@ -46,3 +47,17 @@ import {
 <Playground system="ec" selectedKind="components-button" selectedStory="ghost">
   <Html markup={buttonGhost} />
 </Playground>
+
+## Button with icon only
+
+Button label can be hidden by adding class `ecl-u-sr-only` to it. The label has to be filled in any case, to provide context to screen readers.
```
planctus: Isn't this information outdated?

src/website/src/pages/ec/components/button/docs/code.mdx
```diff
@@ -46,3 +47,17 @@ import {
...
+
+Spacing of icon only button is adapted to make it square, but to have it work correctly, icon should be placed after the label.
```
planctus: Is this true? I've tried appending the label to the button, so after the icon and it still shows correctly

src/website/src/pages/eu/components/button/docs/code.mdx
```diff
@@ -46,3 +47,15 @@ import {
 <Playground system="eu" selectedKind="components-button" selectedStory="ghost">
   <Html markup={buttonGhost} />
 </Playground>
+
+## Button with icon only
+
+Button label can be hidden by adding class `ecl-u-sr-only` to it. The label has to be filled in any case, to provide context to screen readers.
```
planctus: Same question about the EC page..

src/website/src/pages/eu/components/button/docs/code.mdx
```diff
@@ -46,3 +47,15 @@ import {
...
+
+<Playground
+  system="eu"
+  selectedKind="components-button"
+  selectedStory="secondary"
```
planctus, line 62: shouldn't we add:
selectedArgs="icon_name:logged-in;hide_label:!true" to demo the icon only use case?

src/implementations/twig/components/button/button.html.twig
```diff
@@ -50,7 +50,9 @@
   {% endfor %}
 {% endif %}
 
-{% set _label_css_class = _hide_label ? 'ecl-u-sr-only' : 'ecl-button__label' %}
+{% if _hide_label %}
```
planctus: It feels it's a bit unsafe what we are doing, i mean that you specify in the docs that the icon has to be present for the hide_label param to work properly, but to me we should kind of ignore it, then, in case there is no icon, because otherwise you get the square layout and the label still visible, and we could prevent this from happening

src/implementations/twig/components/site-header/__snapshots__/site-header.test.js.snap
```diff
@@ -107,22 +107,22 @@ exports[`Site Header EC renders correctly 1`] = `
                     Select your language
                   </div>
                   <button
```
planctus, line 109: Probably this is not relevant, the language toggle is a link, not a button, but it looks like this currently:
image
In Eu:
image

I do have concerns regarding this color palette for EC, by the way, how can this blue be the primary color? Everything looks like it's disabled and the "disabled" becomes almost unreadable.

planctus (changes requested):
Just a question about the template and a few others about the docs. We can improve the link to playground to demo the icon only use case.

planctus:
I removed the markup changes label from the pr, this is for ECL 4, right..? I feel that we might have too many prs involving markup changes for this label to be relevant, and it's a major version, anyhow,

src/website/src/pages/eu/components/button/docs/code.mdx
emeryro: unfortunately, passing arguments set in a select list in storybook do not work (at least I never found a way to make it work).
So we could pass the boolean to hide the label indeed, but not the icon name

src/implementations/twig/components/site-header/__snapshots__/site-header.test.js.snap
emeryro: I noticed it too. 
The ticket for the language switcher is ready, so we'll handle that in it.

And for the primary color, that's how the designer defined it... It doesn't look so bad on my side, maybe it's related to your screen

planctus (approved):

PR #2894, opened 2023-06-06
feat(label): ECL4 Label - FRONT-3962

emeryro:


planctus:
The playground link in the website needs an update, it's linking to https://ecl-preview-2894--europa-component-library.netlify.app/playground/ec/?path=/story/components-label--default which doesn't exist, this is already the case in production so i will also open a ticket to fix it in v3

planctus (approved):

PR #2896, opened 2023-06-06
feat(footer): ECL4 footer - FRONT-3970

emeryro:


planctus:
is this duplicated title expected in EU? 
image

As i mentioned in the other pull request, i would not use the markup changes label for ECL 4 prs.

emeryro:
That's how it is currently for the EU footer, so I guess that it is alright. That's because the long list of institution  is grouped into a single element on mobile.
And ok for the markup change tag

planctus (approved):

PR #2901, opened 2023-06-08
feat(text-input): ECL4 text input and form group - FRONT-3983

emeryro:


src/implementations/vanilla/components/form/_form.scss
```diff
@@ -3,19 +3,10 @@
  */
 
 @use 'sass:map';
-@use '@ecl/theme-dev/theme';
-
```
planctus, line 7: i wish we could be able to provide a form-group template, one of the most questionable things in ECL about the twig templates is the duplication in all the form elements of the related code. This is how is done in webcomponents, for instance, but in twig it would be different. Still we could handle all the rendering in the form group template and use the single form element templates only to pass params to it. The form group would then have dozens of params, probably, but that complexity can be completely handled on our side, keeping the same "experience" in the implementation for the users..

src/implementations/vanilla/components/form/_form.scss
```diff
@@ -3,19 +3,10 @@
...
-$_font-invalid: null !default;
-$_font-optional: null !default;
-$_form-element-margin: null !default;
-$_margin-invalid: null !default;
+
+// Exposed variables
+$theme: null !default;
+$form-group: null !default;
 
 .ecl-form-group {
```
planctus, line 11: this css is critical, we've been adding code here to handle all the exceptions, but here we use lots of those selectors that we actually shouldn't use to decouple a bit the html structure from the styles.
Not sure we can get rid of this approach, but to a certain extent it feels this could be simplified, when really needed exceptions could be added at the component level, instead of here in the form group.

planctus (commented):
I would love to find a better solution for handling both the styles and the "markup" of the form elements. Ok to keep all these separated templates, although most of them are rendering the same html tag, but the lack of a form group template forces us to duplicate a lot of markup.
It also feels that v4 could be the chance to investigate a different approach about the styles of the form group elements, ideally the form css shouldn't handle all the use cases itself, delegating customsation to the form input components.

src/implementations/vanilla/components/form/_form.scss
emeryro: this could be a good opportunity to do it, indeed.
But I guess that we should discuss it with EWPP, to see how we could provide usable template. Form elements are usually quite tricky to handle

emeryro: I created the ticket https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-3995 to see how to handle that

src/themes/ec/_variables.scss
```diff
@@ -181,6 +181,42 @@ $link: (
...
 
+// Form
+$form-group: (
+  color-label: var(--ecl-color-dark),
+  color-help: var(--ecl-color-dark),
+  color-invalid: var(--ecl-color-error),
+  color-optional: var(--ecl-color-dark),
+  color-required: var(--ecl-color-error),
+  font-label: map.get($font, 'm'),
+  font-weight-label: map.get($font-weight, 'regular'),
```
planctus, line 192: but why did they remove the bold from the label, i wonder..

planctus (approved):

PR #2908, opened 2023-06-14
fix(storybook): Fixing args in url when the value is from a select or a radio control - FRONT-3996

planctus:
The solution was to add a mapping object in all the controls.
It can be tested on the featured-item, we currently have one link to the story where we try to set the image on the right and it doesn't work in production, it works in this pr, instead.
Other args can be tested by selecting a value, getting the url, load it in a different tab and see if the query string is still in place.

There is one single control where this cannot be applied, because we set the full url of images, this is not going to be in the query string for security reason.

### DIFF
--- a/src/implementations/twig/components/banner/banner.story.js
+++ b/src/implementations/twig/components/banner/banner.story.js
@@ -74,6 +74,11 @@ const getArgTypes = (data) => {
           l: 'large',
         },
       },
+      mapping: {
+        s: 'small',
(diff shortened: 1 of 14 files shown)


emeryro (dismissed):
nice!
We will have to apply this update on ECL4 too

planctus:
yep, but i still have to fix the icon story, i noticed.. ;)

emeryro (approved):

PR #2909, opened 2023-06-15
fix(list-illustration): fix display issue - FRONT-3997

emeryro:
fix width overlay issue on desktop

### DIFF
--- a/src/implementations/vanilla/components/list-illustration/_list-illustration.scss
+++ b/src/implementations/vanilla/components/list-illustration/_list-illustration.scss
@@ -185,7 +185,7 @@ $_icon_color: null !default;
   }
 }
 
-@include breakpoints.up('m') {
+@include breakpoints.up('s') {
   .ecl-list-illustration--zebra {
     left: auto;
     margin-inline-start: 0;
--- a/src/implementations/vanilla/components/list-illustration/_list-illustration.scss
+++ b/src/implementations/vanilla/components/list-illustration/_list-illustration.scss
@@ -185,6 +185,14 @@ $_icon_color: null !default;
   }
 }
 
+@include breakpoints.up('m') {
+  .ecl-list-illustration--zebra {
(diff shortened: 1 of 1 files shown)


src/implementations/vanilla/components/list-illustration/_list-illustration.scss
```diff
@@ -185,6 +185,14 @@ $_icon_color: null !default;
   }
 }
 
+@include breakpoints.up('m') {
```
planctus: but why do we have this css?

```
.ecl-list-illustration--zebra {
    -webkit-margin-start: -50vw;
    left: 50%;
    margin-inline-start: -50vw;
    position: relative;
    width: 100vw;
}
```
what is this for? The proposed change fixes the main issue with our demos, but still it feels we shouldn't force the element to take full width even in mobile and tablets, potentially we could be doing this still in a layout with columns, for instance, and in any case it's not really clear what is that we want to achieve with that.

emeryro: this was a request from COMM, to have the zebra list look less like cards when on mobile...
image

emeryro: but if you think that it would be too dangerous to keep it, we can put it away, at least for now

planctus: did not really get the "comparison" with cards, but anyway, like this is surely safer, if we limit it to the mobile breakpoint.
Just to understand, this is done to ensure that the grey background of the zebra would cover the full width of the viewport regardless of a container that could set a padding, basically.

emeryro: yes basically this is to have no white padding around the grey area, so it looks more like a vertical list of items

planctus (approved):

PR #2918, opened 2023-06-21
feat(form): ECL 4 - Adding form group template, adapting form elements - FRONT-3995

planctus:
One of those pull request that needs some guidance to recognize the changes made:

1. Creation of a form group template that renders the form elements and includes the form input template(s)
2. Single story replacing the individual story files for form elements
3. The data structure has not changed much, for clarity the input props are expected in a input named object when using the form group
4. Spec files have been kept for how they were, including the form group props, so that they can be used for the new demos as well as the single tests for the input elements
5. Styles have been simplified for the group, they set generic styles applied to all the form elements the same way, it is still possible to introduce exceptions in the input scss files but it would be great if we could just keep the same spacing everywhere.


src/implementations/twig/components/checkbox/checkbox-group.html.twig
```diff
@@ -14,28 +14,17 @@
     - "name" (string) Attribute name, eg. 'data-test'
     - "value" (optional) (string) Attribute value, eg: 'data-test-1'
   Blocks:
-  - "label"
-  - "helper_text"
   - "invalid_text"
```
emeryro, line 17: why do we we still have an invalid_text parameter here? Can't we rely on the one in form group?

src/implementations/twig/components/checkbox/README.md
```diff
@@ -8,7 +8,6 @@ npm install --save @ecl/twig-component-checkbox
 
 ### Parameters:
 
-- **"label_id"** (string) (default: '')
 - **"items"** (array) (default: [])
   - "id" (string) (default: '')
```
emeryro, line 12: shouldn't we put the data structure of the checkbox group here also? That's the one used in the example below

src/implementations/twig/components/text-input/text-input.html.twig
```diff
@@ -76,43 +76,6 @@
 
 {# Print the result #}
 
-<div class="{{ _css_group_class }}">
```
emeryro, line 79: corresponding parameters have not been removed from the top comment

src/implementations/vanilla/components/form/_form.scss
```diff
@@ -15,63 +15,15 @@ $form-group: null !default;
   display: block;
   margin: 0;
```
emeryro, line 16: shoudln't we name this file "form-group", to be consistent?

src/implementations/vanilla/components/checkbox/_checkbox.scss
```diff
@@ -29,6 +29,10 @@ $_label-color: null !default;
   position: relative;
 }
 
+.ecl-checkbox ~ .ecl-checkbox {
```
emeryro, line 32: this causes extra spacing in the multi select component:
image

src/implementations/twig/components/form-group/form-group.story.js
```diff
@@ -0,0 +1,150 @@
...
+    show_label: true,
+    show_helper: true,
+    show_error: true,
+    invalid: data.invalid || false,
+    disabled: data.disabled || false,
+    required: data.required || false,
+    label: data.label || '',
+    helper_text: data.helper_text,
+    invalid_text: data.invalid_text,
+    input: {
```
emeryro, line 36: having the full input exposed in a control makes it harder to play with; this is no description for the editable fields, or the list of possible values. Some fields are also not meant to be edited (icon path), or redondent with form group (required).

image

image

Won't it be possible to expose only a few fields, like placeholder and width?

emeryro (changes requested):
Thanks for the big cleanup of the forms!

That's quite a big change at template level, we should document this in the migration guide

src/implementations/twig/components/checkbox/checkbox-group.html.twig
planctus: well, this is a kind of unique situation, i guess that because of a design request, when we have a single checkbox we render the form error under the field, i don't think this makes sense, but i kept it as it was, with the checkbox group taking over the rendering of the form error block.
Now, what i will do is to pass the invalid_text and _invalid_icon to the checkbox-group template from the form group one but it feels they should still be listed in the checkbox-group, you could potentially decide to include this directly in certain situations, even without the form-group.

src/implementations/twig/components/checkbox/README.md
planctus: and isn't that the case already? I notice now that we pass the name from the checkbox group to the single items, so i'm removing it from the items and i see that _invalid_icon is not listed, will add that but it seems you noticed something else, i don't know..

src/implementations/vanilla/components/form/_form.scss
planctus: well, it was the case already, i mean, the main selector was and still is ecl-form-group, but ok, let's do it now.

src/implementations/twig/components/form-group/form-group.story.js
planctus: mmh..i don't think that would be enough. Those are not common properties of all the form elements, so this would mean re-introducing conditions and complexity where i tried to get rid of both. If we would have to add relevant data belonging to the input and create controls for them we would probably decide to go back to indivdual stories and separate definitions of controls, while here we try to get rid of all this redundancy, but we can still decide differently, eh.

The way the object control works it makes it possible to show the data structure while being able, if really needed, to change something in there to play with the component. 
It is true that some of those value are not meant to be edited but still it is relevant to show their presence, maybe.
Furthermore having the whole input as a single control let us create a seamless way to merge the data and the controls, a single  Object.assign(args.input, data.input); is used to pass to the controls the props and the values from the data.

To me the important is to remove all the unnedeed duplicates, redundancy, logic, rationalize in other terms how we handle the form elements.
About the user experience, it feels like not much has changed, but didn't really expose many controls relative to the inputs, in v3, so this might feel as an improvement as well.

src/implementations/vanilla/components/form/_form.scss
emeryro: it should be updated also in the package.json of the form elements

emeryro:
Almost good, just a remark left.

also, the order of stories is kind of chaotic, we could make it more readable
image

emeryro (approved):

PR #2921, opened 2023-06-26
chore(twing): ECL4 Upgrading twing and twing-loader, tests, stories, website pages and components adaptations

planctus:
In this pull request:

- twing and twing-loader upgrade to the latest stable version
- stories adapted to the new twing render method, it returns a promise
- showcase pages in the website adapted, also the html source code shown there is a twing rendering, so a promise
- test have been adapted as well to the new rendering
- the result should not present any relevant difference compared to the current one.

There is one issue that came out with the iframe resizer, with the menu, what you see in those pages reproduces what we currently have, but it does it by manually setting the height of the iframe and preventing the autoresizing, for some reasons there is a loop otherwise that would continuosly resize the iframe.


src/implementations/twig/components/table/__snapshots__/table.test.js.snap
```diff
@@ -744,8 +742,6 @@ exports[`Table Default renders correctly with row extra attributes 1`] = `
         </tr>
         <tr
           class="ecl-table__row"
-          data-test=""
```
emeryro, line 747: extra attributes are no longer tested here

src/playground/ec/.storybook/main.js
```diff
@@ -33,6 +33,8 @@ if (isProd) {
 
 const webpackFinal = (config) => {
   // Trick "babel-loader", force it to transpile @ecl addons
+  config.module.rules[0].exclude = /node_modules\/(?!@ecl\/).*/;
```
emeryro, line 36: do we need to have this line duplicated?

emeryro (changes requested):

src/implementations/twig/components/table/__snapshots__/table.test.js.snap
planctus: wow, this looked weird, since nothing was touched in the template or the data..but:
it turned out that the current template is actually expecting an impossible data structure, the rows array contains an array of objects with the cell data, we did expect an extra_class and extra_attributes props with a value, which is not valid json or js, it was working because we were adding props to these objects dynamically in the test, but you could not reproduce that data structure. 
This is the reason why with the new twing, these attributes and classes where not rendered anymore.
Furthermore the extra_attributes as a string cannot be supported on our side, they would be escaped and we cannot use |raw, so i've used for those the same approach we have for the main attributes.
The overall result is that now we would have the row cells inside a "data" named object, and then we have extra_classes as a string and extra_attributes as an array of objects

emeryro (approved):

PR #2922, opened 2023-06-26
feat(notification): ECL4 notification - FRONT-3985

emeryro:
Previously called "Message"


src/implementations/twig/components/notification/notification.story.js
```diff
@@ -67,6 +84,9 @@ const prepareData = (data, args) => {
   const clone = JSON.parse(JSON.stringify(data));
 
   // Optional elements
+  if (!args.show_title) {
+    delete clone.title;
```
planctus, line 77: it seems this control is not working, doesn't hide the title, could it be a conflict that i solved not in the right way..?

src/implementations/twig/components/notification/notification.html.twig
```diff
@@ -47,31 +47,34 @@
...
     icon: _icon,
-    extra_classes: 'ecl-message__icon',
+    extra_classes: 'ecl-notification__icon',
   } only %}
   
-  <div class="ecl-message__content">
+  <div class="ecl-notification__content">
     {% if _close is not empty %}
-      {% set _close_icon = {} %}
+      {% set _close_icon = { name: 'close-filled' } %}
```
planctus: this is really confusing, we first set the _close_icon as an object with a name property, then if the icon is present in the data we just override that with what is coming from the data.
I'm not sure what is the desired behaviour, should the users be able to use a different icon if they want? In that case this approach can work but i would still expect a merge, so that if the name is not present in the data we would complete that missing info, but the icon we set here is different than the one requested in the data and used in the demo, here we have close-filled and we demo close, instead.
If the icon is not supposed to be overridden, then the merge should happen the other way around, and we would force the name of the icon that way. To be honest i would expect also the size of the icon to be set by the template, in other words we only  need a path to an icon sprite, most likely, and we could even think about only asking for that and building the icon object ourselves.
Right after wiriting this i see that we have a different icon in the two systems (didn't we say that this should have been avoided from the beginning?) and even different sizes for the icon depending on the variant..? Are they out of their mind..? i mean..

planctus (commented):
Mmh..i'd love to see things going the other way around, reducing exeptions, misalignment in the two systems, and instead we are increasing the complexity, forcing then the users to be fully aware of what they are doing (end enforcing logic on their side) in order to properly implement a single (theoretically simple..) component.
I see that the print was not re-done, guess that was on purpose, but i do have a certain level of confusion about what we are doing..

src/implementations/twig/components/notification/notification.html.twig
emeryro: yes, I was not sure how to handle this... Having different icon for both systems would mean relying on the user to provide the correct data. I put a line for that in the documentation, we'll see if it comes as an issue during the first implementation (that could force the designers to review their work)
What I could do for now is having different data files for EC and EU, so we don't have to do this kind of weird thing in the template or story.

PR #2924, opened 2023-06-27
feat(link): simplify link icon css - FRONT-3981

emeryro:


planctus (commented):
I do see the menu still using the ecl-link--icon-after selector, is this on purpose or something to still act on?

emeryro:
Good catch!
This is hardcoded in the menu template. I had a quick look at it, and I think that we will have to cleanup this part anyways when updating the menu (the links are not using the link template for some reason, we should see if we could fix that)
So I'd suggest handeling that in another ticket

PR #2926, opened 2023-06-28
feat(spinner): ECL4 spinner - FRONT-3991

emeryro:


src/implementations/twig/components/spinner/spinner.story.js
```diff
@@ -93,20 +94,32 @@ const prepareData = (data, args, story) => {
   return data;
 };
 
+const renderStory = (data, args) => {
+  let story = spinner(prepareData(data, args));
+  if (args.overlay) {
+    story = `
+        ${story}
+        <p class="ecl-u-type-paragraph ecl-u-mt-none">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer facilisis scelerisque massa, tempor scelerisque neque pharetra et. Duis sed orci in tellus gravida volutpat. Aliquam volutpat efficitur turpis. Cras non tortor fermentum ipsum viverra varius. Quisque congue viverra lacus, a pellentesque sapien congue nec. Duis at vulputate ligula. Sed ac elementum erat, ac pretium mauris. Sed vehicula arcu neque, in consequat velit porta ac. Mauris ullamcorper ante a urna blandit, et ultricies lacus efficitur. Vivamus dignissim felis a nunc cursus lacinia eu sit amet diam. Etiam vestibulum libero ac ultricies molestie. Vestibulum mattis quis nulla ac pharetra. Donec at velit at orci ornare euismod. Curabitur sed malesuada sapien, ac ultrices quam. Pellentesque a tristique libero, et fringilla odio. Sed tincidunt vehicula eros nec mattis.</p>
```
planctus: can we use import { loremIpsum } from 'lorem-ipsum'; instead of this hardcoded text? The package is already there so you can  already import it.
An example usage of this lib is in the inpage story

planctus:
i would maybe change the default viewport used in storybook, the thing is that otherwise the "overlay" is not really "effective" we get the text all on the left, like this is just an overhead in our story without any clear improvement in the user experience.

planctus (approved):

PR #2933, opened 2023-07-04
feat(language switcher): add resize event - FRONT-4013

emeryro:


planctus (approved):
The 200ms delay creates a bit of fun in seeing this arrow moving around, but it's good to me, this is something we are doing more for not being bothered by a tester than for the users, i'd say.. :)

PR #2935, opened 2023-07-04
feat(breadcrumb): ECL4 breadcrumb - FRONT-4003

emeryro:


src/implementations/twig/components/breadcrumb/breadcrumb.html.twig
```diff
@@ -26,11 +25,10 @@
 
 {# Internal properties #}
 
-{% set _variant = variant|default('') %}
 {% set _breadcrumb_size = links|length %}
 {% set _links = links|default([]) %}
 {% set _icon_file_path = icon_file_path|default('') %}
-{% set _icon_size = icon_size|default('2xs') %}
+{% set _icon_size = icon_size|default('fluid') %}
```
planctus: small thing, but why do we keep "exposing" this parameter..? Is the size supposed to be altered by the users..?
Mind also the fact that we still have "2xs" in the README file.

src/themes/eu/_variables.scss
```diff
@@ -4,6 +4,26 @@
 // Use this file to assign default values to component variables
 // Note: if you override css variables it will take over on these values
 
+// Breadcrumb
+// stylelint-disable length-zero-no-unit
```
planctus, line 8: but why do we disable the rule instead of simply setting 0 or none..?

emeryro: that's because the value is used in a css calc( ), and it does not work without unit

planctus (commented):
i noticed and fixed an issue with the escape in the html source code in the website and also in storybook. You just need to run prettier on a couple of files, it has been upgraded.

PR #2936, opened 2023-07-05
feat(gallery): Handling iframes size via js - FRONT-4016

planctus:
The demo has been filled with different videos from different providers, this can be reverted once this pr gets approved.


src/implementations/vanilla/components/gallery/gallery.js
```diff
@@ -354,6 +355,37 @@ export class Gallery {
...
+      const width = window.innerWidth;
+      const height =
+        this.overlay.clientHeight -
+        this.overlayHeader.clientHeight -
+        this.overlayFooter.clientHeight;
+
+      if (width > height) {
+        iframe.setAttribute('height', `${height}px`);
+
+        if ((height * 16) / 9 > width) {
```
emeryro: I don't really understand why we are still using predefined ratio here (16/9 and 4/3).
If I'm not wrong, we have the available height and width, but why assuming that the video is using one ratio or the other?

emeryro (changes requested):
It works fine in most cases, but the vimeo video seems to be still cropped
image

src/implementations/vanilla/components/gallery/gallery.js
planctus: yes, i was fixing that while you reviewed this, should be ok now. 
About the ratio, before we were always forcing 16/9, the idea here is to keep that format as long as it possible without cropping the video in landscape viewports and switching to 4/3 in portrait.
Like this we can mantain a certain consistency also about the ratios, but yes we could also forget about it and always try to fit the container..

emeryro (dismissed):

emeryro (approved):

PR #2943, opened 2023-07-06
feat(media-container): add expandable - FRONT-4017

emeryro:
Main component css classes have been put a level higher, in a container, to be able to add an optional expandable block.

also fix some missing dependencies for picture component


planctus:
Question: where is "Infography" coming from..? :) Is this an english word..? Shouldn't it be infographic..?

src/implementations/twig/components/media-container/media-container.html.twig
```diff
@@ -81,6 +83,10 @@
   {% set _css_class = _css_class ~ ' ' ~ extra_classes %}
 {% endif %}
 
+{% if _expandable is not empty and _expandable.id is defined %}
```
planctus, line 93: why not setting this as an item of extra_attributes instead  of building the string manually?

src/implementations/twig/components/media-container/media-container.html.twig
```diff
@@ -93,6 +99,10 @@
 
 {# Print the result #}
 
+{% if _expandable is not empty %}
+<div class="ecl-media-container__container">
+{% endif %}
+
 <figure class="{{ _css_class }}"{{ _extra_attributes|raw }}>
 {% if _embedded_media is not empty %}
   {% set _embed_class = _ratio ? "ecl-media-container__media ecl-media-container__media--ratio-" ~ _ratio : "ecl-media-container__media" %}
```
planctus: nothing has changed here, but couldn't we set the common class ecl-media-container__media even at the top of the template and then add the ratio selector conditionally..?

src/implementations/twig/components/media-container/media-container.html.twig
```diff
@@ -93,6 +99,10 @@
 
 {# Print the result #}
 
+{% if _expandable is not empty %}
```
planctus: can we keep the html indentation here? this shouldn't be indented basically, same for the closure tag.
The solution is not that clean, adding a wrapper conditionally as an element for the whole component, but i see that it was somehow needed in order to accomplish this.. 
But what if we add it anyhow, what if we set the root classes (or part of them) to this div instead of the figure for all the variants? Aslo because it has zero styles associated, currently...

planctus (commented):
few comments left to your evaluation.. ;;)

emeryro:
thanks for the feedback @planctus 
Could you have a final look, as the markup and css have been updated?

src/implementations/twig/components/media-container/media-container.html.twig
```diff
@@ -93,44 +106,51 @@
...
+  {% if _description is not empty %}
+    <figcaption class="ecl-media-container__caption">
+      {{- _description -}}
+    </figcaption>
+  {% endif %}
+  </figure>
+
+  {% if _expandable is not empty %}
+    {% include '@ecl/expandable/expandable.html.twig' with _expandable|merge({
+      extra_classes: _expandable_class,
```
planctus, line 156: doing this would override existing extra_classes in the data, i think we should merge this class with the others, if present.

src/implementations/twig/components/banner/banner.story.js
```diff
@@ -223,7 +224,10 @@ const prepareData = (data, args) => {
 
 const renderStory = (data, args) => {
   let story = banner(prepareData(data, args));
-  if (args.width === 'container' || args.width === 'inside') {
+  if (
+    args.width === 'inside the grid container' ||
```
planctus, line 228: it felt there was something strange since you are receiving the labels and not the values in the args object, this is due to a reverse mapping, it should be:
```
    mapping: {
      'outside the ecl-container': 'outside',
      'inside the ecl-container': 'container',
      'inside the ecl-container, with fullwidth class': 'inside',
    },
```
and then you would deal with the values here, the same in the url ;)

planctus (changes requested):
i'm not sure this thing makes really much sense, not it's better since the button is part of the standard markup of the component, but still it feels like this is a composition sort of...anyway, i've just noticed something odd in the story, tha we can fix

planctus (approved):

PR #2946, opened 2023-07-07
feat(card): Removing height: 100% - FRONT-4004

planctus:
I don't see the reason why setting height:100%, not that it has an impact, normally, but it breaks when used in webcomponents, into a column layout.


emeryro:
There is a reason in fact: the height of the card is harmonized when having this 100% height:

with 100% height
image

without
image

planctus:
oook, true, thanks for spotting this, in the meantime i have patched the card scss in the webcomponents, so let's keep it thw way it is in ECL.

PR #2948, opened 2023-07-11
feat(lists): ECL4 - Unordered and ordered list - FRONT-4007

planctus:
This is only including changes on ordered and unordered list, description-list has been kept outside of this pr.

- same variants have been added in the ordered list as well
- variant no-bullet has been renamed in no-marker since it handles different symbols


src/implementations/twig/components/ordered-list/ordered-list.html.twig
```diff
@@ -48,21 +54,28 @@
...
-          <li class="ecl-ordered-list__item">
-            {{- _item.label -}}
-            {% if _item.nested is not empty and _item.nested is iterable %}
-              {{- list.ordered_list(_item.nested) -}}
-            {% endif -%}
-          </li>
+        {% set _item_class = 'ecl-ordered-list__item' %}
+        
+        {% if not nested and loop.first %}
+          {% set _item_class = _item_class ~ ' ecl-ordered-list__item--first' %}
```
emeryro: can't we use css selectors like first-of-type instead of having specific classes?

emeryro (commented):
the new style of unordered list is... orginal, let's say

src/implementations/twig/components/ordered-list/ordered-list.html.twig
planctus: The thing is that this will also trigger the first and last items of the nested lists, but ok, still better to override a css rule than requesting additional classes..

emeryro (approved):

PR #2950, opened 2023-07-12
feat(description-list): Revamp for v4 - FRONT-4024

planctus:


emeryro (changes requested):
Font does not seem to be correct on print (it should be set in the $description-list-print variable)
image

We could also point out in the documentation that the link are now standalone (mostly for people not using twig)

Other than that, that's fine

emeryro (approved):

PR #2951, opened 2023-07-12
feat(styles): Apply new structure to EU - FRONT-4008

emeryro:
add css variables for EU colors, based on already updated component, and pending ones

Notes:

- there are way more color tint used in EU, so more variables
- pages "colours" on the website will be updated in another ticket


src/themes/eu/_variables.scss
```diff
@@ -222,21 +222,21 @@ $form-group: (
...
-  border-color-invalid-focus: map.get($color, 'red-100'),
-  color-input: map.get($color, 'grey-140'),
-  color-placeholder: map.get($color, 'grey-40'),
+  border-color: var(--ecl-color-dark-10),
+  border-color-hover: var(--ecl-color-dark-10),
+  border-color-focus: var(--ecl-color-primary),
+  border-color-invalid: var(--ecl-color-error),
+  border-color-invalid-hover: var(--ecl-color-error-10),
+  border-color-invalid-focus: var(--ecl-color-error-10),
+  color-input: var(--ecl-color-dark20),
```
planctus, line 238:
```suggestion
  color-input: var(--ecl-color-dark-20),
```

planctus, line 238: wow, at the beginning i thought this was a typo, instead we have both dark-20 and dark20, as well as dark-10 and dark10, guess there is reason for this but it's super confusing at the moment.. I read now the explanation added in the migrating docs and i'm even more confused.. :D

emeryro: Initially I wanted to use dark-10, dark, dark+10, ... but the "+" is not allowed in variable names...
So "dark-20" should be read "dark minus 20" and "dark20" is "dark plus 20" 

If you think of a better way to name them, I'm open

planctus: well, we are not really the first one dealing with colour palette generation and naming, i think the standard is to define a scale, starting from / ending to and then the itermediate steps are consistently going from the lightest to the darkest variant.
With this example, the ecl-color-dark could be an alias for, i don't know, ecl-color-dark-50 and then the darker one would get ecl-color-dark-60 / 70 etc.. and the lighter ecl-color-40 / 30 etc..
I'll add here a link to a very well known design system, but more or less the same approach is used in many others.
https://m2.material.io/design/color/the-color-system.html#color-usage-and-palettes

emeryro: what we did on ECL is what had also been done on OEL https://oelibrary.netlify.app/default/?path=/story/resources-colors--default
Not saying that it is the best solution, for sure, but at least that's consistent
Some systems uses "primary-light", or "primary-darker" to define the colors, here it is just converted to numerals.
In the end, the color palette is the same, instead of having 
```
primary-10
primary-20
primary-30 (=primary)
primary-40
primary-50
```
we have
```
primary-20
primary-10
primary
primary+10
primary+20
```
We can check with designers if they could still change the way colors are handled (if that won't break anything on Figma)
We can also check with ECL users if this approach would trigger issue, or would be harder to understand.

planctus: the difference between the two, for me, is that the first makes sense and second doesn't :)
Or better, if it was somehow possible to define these variables as such, that would be acceptable, although still not preferrable, for me, because it would still be prone to errors.

in reality, while the first is a logical ordering in a scale of color tints, the second creates two mirrored scales based on the "pure" color, where there is only the presence of a character changing (an hyphen...). I might not be the model for this evaluation, but to me it looked like a typo as soon as i saw it, which to me means that is not intuitive, but it's only my opinion.

I don't see how this would involve changing things on the design side, it's a naming convention, if we cannot adopt the same we won't , but it's not that we are forced to follow something we cannot even reproduce as such...

emeryro: well... We could put this system in place for EU, where we have the full palette defined.
But for EC, there may be new color added later, if the designers identify a new need in a component (that was one of the reason they wanted to keep some flexibility). So the question would be where to put the middle point (primary-500? primary-450?). 
As far as I know there is no real logic in the palette definition, they just chose some lighter and darker tint of blue for primary.
But in any case we have to use the same naming convention on design and implementation side, otherwise we would have to maintain some kind of manual mapping on our side to translate their name to ours

planctus: not sure i really understand here, isn't just the same..? it's a name associated to a value, in any case it would need an "indentifier", in both naming conventions..
But anyway, think i've argued enough, if you feel we are ok using this naming convention because we are forced to by the questionable  approach used by our designers, ok, let's close this discussion.

PR #2961, opened 2023-07-18
fix(footer): fix accessibility and cookie links - FRONT-4031

emeryro:


planctus (commented):
I noticed that we have two duplicated links to https://european-union.europa.eu/privacy-policy_en in "privacy policy" and in "legal notice", not sure about this, but maybe this could be fixed as well in this pr.
Ah, there is also the link to the european commission that points to https://ec.europa.eu/commission/index_en
shouldn't this be updated to https://commission.europa.eu/index_en ?

planctus (approved):

PR #2962, opened 2023-07-19
feat(page header): ECL4 page header - FRONT-4021

emeryro:


planctus:
we are masters in doing things we don't understand ourselves, maybe it's me but i don't see any difference between the different families, there is no selector defining any of these "variants" which result in just a different name associated to the same thing and lots of different demos that are just reproducing the same component.
Maybe i'm wrong but i don't really see why keep the families categorisation, it doesn't make any sense, it just increases the complexity in the way we present this stuff and for the users it feels it's only confusing, i'm confused myself..

emeryro:
Indeed, now the page header is the same for all families. We could ask COMM about that, but the only reason we keep this difference is for business reasons

PR #2966, opened 2023-07-24
feat(language-switcher): ECL4 language switcher - FRONT-3963

emeryro:


docs/Migrating-v4.md
```diff
@@ -125,6 +125,8 @@ EX: `{% include '@ecl/form-group/form-group.html.twig' with {
 
 - Twig parameter `message` has been renamed to `notification`
 - CSS class `ecl-site-header__message` has been renamed to `ecl-site-header__notification`
+- Icon for the close button is now different between EC and EU: `close` for EC, `close-filled` for EU
```
planctus: too bad... :)

src/implementations/twig/components/site-header/site-header-language-switcher.html.twig
```diff
@@ -82,18 +82,15 @@
...
-          size: 's'
-        },
-        hide_label: true,
-        extra_classes: 'ecl-site-header__language-close',
-        extra_attributes: [{ name: 'data-ecl-language-list-close' }],
-      } only %}
+      {% if _language_selector.overlay.close is not empty %}
+        {% include '@ecl/button/button.html.twig' with _language_selector.overlay.close|merge({
+          variant: 'ghost',
+          extra_classes: 'ecl-site-header__language-close',
```
planctus, line 88: i think it's ok like this, i mean that should we expect extra classes or extra attributes for this button? It would be better, in theory, to always merge the data coming from the user with our internal definition of an include.
Here you merge, but then you set extra_classes and extra_attributes, if these are present in the user data they would be overridden.

src/specs/components/site-header/demo/data--ec.js
```diff
@@ -28,6 +28,16 @@ module.exports = {
     code: 'en',
     overlay: {
       title: 'Select your language',
+      close: {
+        variant: 'ghost',
```
planctus: not so important, but since we correctly hardcode this parameter in the template i would remove it from here.

src/specs/components/site-header/demo/data--ec.js
```diff
@@ -28,6 +28,16 @@ module.exports = {
...
+        label: 'Close',
+        icon: {
+          path: '/icons.svg',
+          name: 'close',
+          size: 'm',
```
planctus, line 36: the size should rely on user data? I mean, are they supposed to change it if they want, otherwise this would be probably better defined in the template

src/specs/components/site-header/demo/data--eu.js
```diff
@@ -29,6 +29,16 @@ module.exports = {
     code: 'en',
     overlay: {
       title: 'Select your language',
+      close: {
+        variant: 'ghost',
+        label: 'Close',
+        icon: {
+          path: '/icons.svg',
+          name: 'close-filled',
```
planctus, line 36: tell me that you protested about this, (the different icon in EC and EU) this shouldn't happen, it has a minimal business value, ok, but an high risk of mistakes by the users.

src/implementations/vanilla/components/site-header/_site-header-language-switcher.scss
```diff
@@ -191,6 +199,12 @@ $_active-item-background: null !default;
   }
 
   .ecl-site-header__language-link {
-    width: calc(188px - (2 * #{map.get(theme.$spacing, 's')}));
+    width: calc(
```
planctus, line 204: oh gosh...

src/website/src/pages/ec/components/site-wide/site-header/docs/code.mdx
```diff
@@ -18,6 +18,7 @@ import {
   selectedKind="components-site-wide-site-header"
   selectedStory="core"
   showFrame
+  frameHeight="700"
```
planctus, line 22: i guess that you added this to fix some issue you had without it, but the experience in the website is pretty bad, closing the popover generates a messy and super quick collapse and expand, generally we disable the autoResize when we pre-define the size of the iframe, like this:
```
frameHeight="700"
disableAutoResize
```

planctus (commented):
A few commnents for you to evaluate them, the experience in the website pages should be improved, though.

src/implementations/twig/components/site-header/site-header-language-switcher.html.twig
emeryro: I agree that in theory it should be the case for each include we do, in every component. But that would make the twig template way bigger.
So if we have an easy way to do so, without having to write 4 or 5 lines each times, we could do it

src/specs/components/site-header/demo/data--ec.js
emeryro: unfortunately the size is different between EC and EU, so we cannot really put it in the twig template (and putting it in the story won't help).
I should add that in the doc

src/specs/components/site-header/demo/data--eu.js
emeryro: I raised it when working on the notification, but I guess that it is worth raising it again

src/website/src/pages/ec/components/site-wide/site-header/docs/code.mdx
emeryro: thanks, that's indeed better

src/implementations/twig/components/site-header/site-header-language-switcher.html.twig
planctus: it's twig, might not look beatiful but it's not even that we have to do something that complex to achieve this, in this case it would be like:
```
        {% include '@ecl/button/button.html.twig' with _language_selector.overlay.close|merge({
          variant: 'ghost',
          extra_classes: _language_selector.overlay.close.extra_classes|default('') ~ ' ecl-site-header__language-close',
          extra_attributes: _language_selector.overlay.close.extra_attributes|default([])|merge([{
            name: 'data-ecl-language-list-close',
          }]),
        }) only %}
```
another approach (we have already many situations handled like this) would be to process the parameters we want to alter separately to combine the existing values with the ones we want to add and then use them in the include, but this is a bit more verbose.

PR #2968, opened 2023-07-25
fix(carousel): Preventing focus trap - FRONT-4037

planctus:


emeryro (changes requested):
This seem to fix the issue when "tabbing" forward, but now when tabbing backward (using shift + tab), we get stuck in the carousel dots

planctus:
Well, the goal here was to avoid the trap. Even in the current state you are not trapped anymore, you can move forward, but ok, it makes probably sense to handle here also the shift+tab which is something we never did, i'd say.
Now it should be possible to always go to the next element or to go back and be able to traverse the carousel focusable elements back.

emeryro (approved):

PR #2969, opened 2023-07-26
feat(site-header): update non-eu language list - FRONT-4039

emeryro:


src/implementations/twig/components/site-header/ec-site-header.story.js
```diff
@@ -58,7 +58,8 @@ const getArgs = (data) => {
   }
   if (data.language_selector) {
     defaultArgs.languages_eu = data.language_selector.overlay.items.length;
-    defaultArgs.languages_non_eu = 5;
+    defaultArgs.languages_non_eu =
+      data.language_selector.overlay.non_eu_items.length;
```
planctus, line 62: mmh..but why do we "expose" a default value for this..? It's not anything related to a parameter, right..? we use it to showcase the component with any kind of amounf of items, so the "table" should simply be empty, there is no default value,

planctus (commented):
I would simply remove this default values we set for the items number. ;)

planctus (approved):

PR #2971, opened 2023-07-27
feat(blockquote): ECL4 blockquote - FRONT-4035

emeryro:


.size-limit.json
```diff
@@ -7,7 +7,7 @@
   },
   {
     "path": "dist/packages/eu/styles/ecl-eu.css",
-    "limit": "327 KB",
+    "limit": "328 KB",
```
planctus: i suggest, especially here in v4, to set the limit calculating a percentage, let's say 5% above the current value, otherwise we are continuosly forced to change this value in concurrent pull requests

planctus (approved):

planctus:
Do we want to take the chance to update the quote in the examples? :)
We are still "quoting" Juncker, maybe we can find something belonging to the new president..

PR #2972, opened 2023-07-28
feat(accordion): EC Revamp for v4 - FRONT-4030

planctus:


src/implementations/twig/components/accordion/accordion.html.twig
```diff
@@ -72,12 +72,12 @@
                 <span class="ecl-accordion__toggle-label">{{ _label_collapsed }}</span>
                 {% include '@ecl/icon/icon.html.twig' with {
                   icon: {
-                    path: icon.path,
-                    name: icon.name,
-                    size: "m"
+                    path: icon.path|default(''),
+                    name: icon.name|default(''),
+                    size: icon.size|default('xs'),
```
emeryro, line 77: we could mention in the doc that the icon size now has to be set manually depending on the system EC/EU

src/implementations/vanilla/components/accordion/accordion.js
```diff
@@ -152,7 +164,13 @@ export class Accordion {
       }
     }
 
-    return this;
+    this.toggles.forEach((item) =>
+      item.classList.remove('ecl-accordion__toggle--active'),
```
emeryro, line 168: one question here: we already have the aria-expended toggled when opening/closing an item. 
couldn't we use that to apply the style instead of adding a new class?

emeryro (changes requested):
one small thing for the print: there is an extra "Open/Close" label displayed
image

src/implementations/vanilla/components/accordion/accordion.js
planctus: i'd say no, because it's another thing, it's not about opened/closed accordion item, but the "latest" interaction by the user, regardless of the fact that the item is expanded or not.
Furthermore, we also need a way to distinguish between an action performed with the mouse and one performed with the keyboard, so...

planctus:
thanks for reviewing this, i removed the toggle label from the print, but this is also in v3, in production, maybe we want to hide that also there?

PR #2981, opened 2023-08-03
fix(site-header): fix close button - FRONT-4018

emeryro:
Second attempt to fix language switcher close button when there is no EU languages


src/implementations/vanilla/components/site-header/site-header.js
```diff
@@ -255,17 +255,29 @@ export class SiteHeader {
     }
 
     // Check total width, and change display if needed
-    this.languageListEu.parentNode.classList.remove(
-      'ecl-site-header__language-content--stack',
-    );
+    if (this.languageListEu) {
```
planctus, line 258: not easy to understand the relation between this selector and the close button, guess that it was not working because a js error thrown by these "lines", and i don't see any error regardless of the number, or the presence of the two lists, so...good as it is! :)

planctus (approved):

PR #2985, opened 2023-08-10
feat(menu): ECL4 menu - FRONT-4032

emeryro:


src/implementations/twig/components/link/link.html.twig
```diff
@@ -105,6 +107,10 @@
   {% set _css_class = _css_class ~ ' ' ~ extra_classes %}
 {% endif %}
 
+{% if _link.hide_label and _icon is not empty %}
```
planctus, line 110: if we are introducing this we could quicky show this behavior in the link component as well, no..?

planctus (approved):

PR #2986, opened 2023-08-10
feat(site-header): EC site header,  v4

planctus:


src/implementations/vanilla/components/site-header/site-header-ec.scss
```diff
@@ -6,6 +6,7 @@
 @use 'sass:map';
 @use '@ecl/theme-dev/theme';
```
emeryro: I guess that we can remove this use

emeryro (changes requested):
This is fine for EC site header (except a small comment below).
For EU, there is a small misalignment, and I think that the css should be updated too, to benefit from the css variables
image

src/implementations/vanilla/components/site-header/site-header.js
```diff
@@ -212,6 +217,30 @@ export class SiteHeader {
     }
   }
 
+  /**
+   * Add title tag in svgs.
+   */
+  improveAccessibility() {
```
emeryro: this works fine, but I'm wondering why we don't put the title directly in the markup, instead of using javascript?

planctus: mmh..i did it this way thinking that we couldn't do it otherwise, but now that i see the icon template it seems we can pass as_image and extra_accessibility['title'] to get it..so yes, i will do it like that

emeryro (approved):

PR #2991, opened 2023-08-17
feat(content-block): V4 revamp for EC - FRONT-4055

planctus:


emeryro (changes requested):
EC is fine, but variables seems wrong on EU
image

planctus:
Ehm..yes, it seems that secondary it's not like dark.. :D

emeryro (approved):

PR #2992, opened 2023-08-18
feat(date-block): EC revamp for v4 - FRONT-4060

planctus:


src/specs/components/date-block/demo/data.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
-  date_time: '2019-09-26',
-  day: '26',
-  month: 'Sep',
-  month_full: 'September',
-  year: '2019',
+  date_time: '2023-06-01',
+  day: '01-02',
+  month: 'Sept-Jun',
```
emeryro: that's a detail, but I find the demo data quite confusing, having "Sept" before "Jun", and no correspondance with the date_time field.
Not sure what the use case is, but I would expect the multiple month to be used mostly when there is an event going on different months, for instance 30 Sept - 02 Oct

emeryro (approved):
Looks good for me.
Just a small remark below

src/specs/components/date-block/demo/data.js
planctus: i perfectly agree, to be honest i wanted to have the same as in figma, but it's surely confusing, i will restore the previous spec file, single date which is the "default".

PR #2998, opened 2023-08-23
feat(checkbox): EC revamp for V4 - FRONT-4069

planctus:


src/implementations/twig/components/checkbox/checkbox-item.html.twig
```diff
@@ -36,6 +37,7 @@
 {% set _helper_text = helper_text|default('') %}
 {% set _label = label|default('') %}
 {% set _css_class = 'ecl-checkbox' %}
+{% set _required_text = required_text|default('') %}
```
emeryro, line 40: Shouldn't it be '*' by default?
Also we could document this new parameter in the migration guide

src/implementations/vanilla/components/checkbox/_checkbox.scss
```diff
@@ -130,52 +118,70 @@ $_label-color: null !default;
...
-      outline: 2px solid map.get(theme.$color, 'blue-100');
-      outline-offset: 2px;
+      @if map-has-key($checkbox, 'hover-circle') {
+        &::before {
+          @each $property, $value in map.get($checkbox, 'hover-circle') {
+            #{$property}: $value;
+          }
+        }
+      } @else {
+        outline: 2px solid map.get($theme, 'color' 'blue-100');
```
emeryro: We're not supposed to use color blue-100 anymore (or any non semantic colors). I assume that it should be --ecl-color-primary

src/implementations/vanilla/components/checkbox/_checkbox.scss
```diff
@@ -130,52 +118,70 @@ $_label-color: null !default;
...
     }
   }
 
   &:checked {
     ~ .ecl-checkbox__label {
       .ecl-checkbox__box:not(.ecl-checkbox__box--disabled) {
-        outline: 2px solid map.get(theme.$color, 'blue-100');
-        outline-offset: 2px;
+        @if not map-has-key($checkbox, 'hover-circle') {
+          outline: 2px solid map.get($theme, 'color' 'blue-100');
```
emeryro: Same here

emeryro (changes requested):

src/implementations/twig/components/checkbox/checkbox-item.html.twig
planctus: Changed a bit the logic, we never set a default for the required text so also in this case we define it as an empty string and we expect a `required_text` prop to be defined in the checkbox object when there is a single, required, checkbox.
An item has been added in the migration doc.

emeryro (approved):

PR #3002, opened 2023-08-28
feat(card): ECL4 Card - FRONT-4057

emeryro:


planctus:
It feels like we could have only one story in storybook, the inclusion of a section of the component is not defining a variant for it, i find this only confusing.
We can still maintain the two examples in the website if that is what comm expects, but in storybook it feels it could just be one with the "list" disabled by default.
What about then using a different viewport as the default one for this component? Demoing full width makes it huge and requires lots of scrolling to see the whole..

planctus (approved):

PR #3008, opened 2023-08-31
feat(expandable): ECL4 expandable - FRONT-4071

emeryro:


planctus (commented):
All good for me, but i wonder: wouldn't it be the case to "minimally" stlye the "dropdown"? We keep having this utility class used in order to give some spacing to the "collapsed" content, but it feels like this could be done via our css instead, otherwise it's always needed to add a certain utility class in order to do something that looks pretty "standard".

PR #3009, opened 2023-09-01
feat(rating-field): EC Revamp for v4 - FRONT-4072

planctus:


src/implementations/vanilla/components/rating-field/_rating-field.scss
```diff
@@ -6,37 +6,56 @@
...
 
 @use 'sass:map';
-@use '@ecl/theme-dev/theme';
 @use '@ecl/vanilla-layout-grid/mixins/breakpoints';
 
+// Exposed variables
+$theme: null !default;
+$rating-field: null !default;
+
 .ecl-rating-field {
```
emeryro, line 15: Not sure if this is just for this element, but when tere is no helper text, the stars are very close to the title
image

src/themes/ec/_variables.scss
```diff
@@ -642,6 +642,16 @@ $range: (
...
 /* stylelint-enable scss/no-global-function-names */
 
+// Rating field
+$rating-field: (
+  checked-color: var(--ecl-color-secondary-100),
+  stroke-color: var(--ecl-color-dark-100),
+  hover-stroke-color: var(--ecl-color-primary-80),
+  checked-hover-color: var(--ecl-color-secondary-140),
+  checked-focus-stroke: var(--ecl-color-dark-100),
+  focus-circle: true,
```
emeryro: I guess that `focus-circle` is no longer used

src/themes/ec/maps/color.scss
```diff
@@ -12,14 +12,14 @@
     'primary-40': #b1c0f8,
     'primary-20': #d8e0fb,
     'secondary': #ffad33,
```
emeryro: `secondary` color should have the same value as `secondary-100`

emeryro (changes requested):
I didn't follow all the discussion about the focus, but the fact that there is no indication of focus on the element would be reported as an accessibility issue.
Here is a proposal: put a focus-within on the parent, so we could at least have a background/border or anything else when there is a focus
image

src/implementations/vanilla/components/rating-field/_rating-field.scss
planctus: oh ,wow...this was coming from the range component..

src/themes/ec/_variables.scss
planctus: it is used, not in the way it was, but still to distinguish the two systems and apply the outline on focus in EU. I agree with you that removing any style on focus might be questionable, but since they rejected their own request of a focus style for "design" reasons, i guess that applying a style to the whole element would make it even worst.

emeryro (approved):

PR #3011, opened 2023-09-04
feat(navigation-list): EC Revamp for v4 - FRONT-4058

planctus:


src/implementations/vanilla/components/navigation-list/_navigation-list.scss
```diff
@@ -71,11 +69,29 @@ $_border: null !default;
...
+    border: 1px solid map.get($navigation-list, 'image-border-color');
   }
 
   .ecl-navigation-list__content-block {
     padding-inline-start: 0;
     padding-inline-end: 0;
   }
 }
+
+@if map-has-key($navigation-list, 'yellow-bar') {
```
emeryro, line 81: nicely done, we should apply this logic to other components too (card for instance), and remove custom rules in theme files

emeryro (commented):
The default display on the website is without border, while on the playground we have border by default. 
Maybe we could have the same display everywhere

emeryro (approved):

planctus:
Ok, i did it but:
are we sure that this is the logic we want to have for this component?
border is param that needs to be passed to each item, which doesn't seem the right way, but more than this, the default is false so unless you explicitely set the border to true in each item you don't get bordered items.
It feels to me the default might actually be true, for the border and in any case it should be a parameter of the parent template and then passed to the child.

emeryro:
I agree that there should be a parent parameter. I don't really see a case where we could have one item with a border and not the others.
Now to set it as default or not, that's probably just a question for the designers: what should be the "default" display they want to advertise

planctus:
But we just aligned the website demo using the bordered version, so if dgcomm wants to advertise the default as the non bordered we should do the opposite, instead, and align storybook.

PR #3012, opened 2023-09-05
feat(banner): EC Revamp for  v4 - FRONT-4073

planctus:


docs/Migrating-v4.md
```diff
@@ -64,6 +64,11 @@ Spacing scale has been enriched for EC, now going from 2XS to 6XL (previously 2X
   overridden by the data provided when working with the EU theme to `m`.
 - A selector `ecl-accordion__item--active` is now added via our vanilla js to mark the latest item the user interacted with.
 
+### Banner
+
+- The picture and the banner credit are now wrapper in a container `ecl-banner__picture-container` to handle the new position of the credit text.
+- Three variants only: `plain-background`, `text-overlay` (replacing text-hightlight) and 'text-box', the `image-overlay` one doesn't esxist anymore.
```
emeryro:
```suggestion
- Three variants only: `plain-background`, `text-overlay` (replacing text-hightlight) and `text-box`, the `image-overlay` one doesn't exist anymore.
```

src/implementations/vanilla/components/banner/_banner.scss
```diff
@@ -4,24 +4,17 @@
...
-$_font-weight: null !default;
+// Exposed variables
+$theme: null !default;
+$banner: null !default;
 
 .ecl-banner {
   align-items: center;
-  border-radius: map.get($_border-radius, 'banner');
-  box-shadow: $_box-shadow;
+  background: var(--ecl-color-branding);
```
emeryro, line 15: not sure why we are using the 'branding' color here. As far as I know, it was meant to be used only in the header/footer.
I guess that it comes from the specs. I'll ask the designers...

src/implementations/vanilla/components/banner/banner.js
```diff
@@ -0,0 +1,172 @@
...
+    const banner = new Banner(root, defaultOptions);
+    banner.init();
+    root.ECLBanner = banner;
+    return banner;
+  }
+
+  constructor(
+    element,
+    {
+      bannerContainer = '.ecl-banner__container',
```
emeryro: I think that we should use the same kind of selector as in the other components (data-ecl-something).
That way we keep a better independence between the js and css

src/implementations/vanilla/components/banner/banner.js
```diff
@@ -0,0 +1,172 @@
...
+    this.checkViewport();
+    this.element.setAttribute('data-ecl-auto-initialized', 'true');
+  }
+
+  waitForAspectRatioToBeDefined() {
+    const aspectRatio = getComputedStyle(this.bannerImg).getPropertyValue(
+      '--css-aspect-ratio',
+    );
+    if (typeof aspectRatio === 'undefined' || aspectRatio === '') {
+      setTimeout(this.waitForAspectRatioToBeDefined.bind(this), 100);
```
emeryro: couldn't it end in an infinite loop, if for some reason this aspect ratio never get set?
maybe we could have a safety net, to give up after x iteration

src/implementations/vanilla/components/banner/banner.js
```diff
@@ -0,0 +1,172 @@
...
+      this.bannerImg.style.aspectRatio =
+        computedStyle.getPropertyValue('--css-aspect-ratio');
+    }
+
+    this.element.style.height = 'auto';
+  }
+
+  /**
+   * Trigger events on resize
+   * Uses a debounce, for performance
```
emeryro: wrong copy/paste for the comment :)

emeryro (changes requested):
with this new banner, there is a display issue on the carousel (EC and EU)
image

also on mobile the copyright is not fully visible
image

src/implementations/vanilla/components/banner/_banner.scss
planctus: yep, this is what we have in figma: background: var(--others-branding, #254ACB);

emeryro (approved):

PR #3023, opened 2023-09-08
feat(expandable): Tweaking expandable content, adding divider - FRONT-4080

planctus:


src/implementations/vanilla/components/expandable/_expandable.scss
```diff
@@ -16,11 +16,30 @@ $_expandable-content-color: null !default;
...
+  height: 1px;
+  margin: 0;
+}
+
 .ecl-expandable__content {
   color: $_expandable-content-color;
-  font-family: #{map.get(theme.$font-family, 'default')};
+  font: map.get(theme.$font-prolonged, 'm');
+  margin-top: map.get(theme.$spacing, 'm');
+  max-width: 80ch;
```
emeryro: I'm a little worried by this max width. It will work fine when there is a block of text in the expandable, but if there is something else (table, image, ...) it should probably take the full width

src/implementations/vanilla/components/expandable/_expandable.scss
```diff
@@ -16,11 +16,30 @@ $_expandable-content-color: null !default;
...
+  display: none;
+  height: 1px;
+  margin: 0;
+}
+
 .ecl-expandable__content {
   color: $_expandable-content-color;
-  font-family: #{map.get(theme.$font-family, 'default')};
+  font: map.get(theme.$font-prolonged, 'm');
+  margin-top: map.get(theme.$spacing, 'm');
```
emeryro, line 23: To limit the risks of double margin, we could remove the margin top of the first element. That's what has been done on ECL4
https://github.com/ec-europa/europa-component-library/pull/3008/files#diff-a5765bf8fb7136b374a2c627867a3c0463f88b0b4cb4504a4b405c1c35cbd92aR33

src/implementations/vanilla/components/expandable/_expandable.scss
emeryro: Also, couldn't we set the max width on the child elements rather than on the container, and so use a margin-bottom instead of an additional hr?

planctus: mmh..yep, although this is also valid for the left indentation, the truth is that on design side they always "deal" with text only inside this hidden container, it all originated from this..
So, i followed your suggestion, but it feels weak in any case, if we remove the max-width then we have to rely on a parapgraph tag wrapping the content, which is what has been done.
The main advantage of this is that we can set the border directly on the content wrapper, so no markup change in this pr.

src/implementations/vanilla/components/expandable/_expandable.scss
planctus: i did it, but it's only going to work as long as the margins are not set through an utility class. I considered the case of multiple <p> so resetting margin-top of the first and margin-bottom of the last, if there is only one both margins will be reset

emeryro (approved):

emeryro (approved):

PR #3024, opened 2023-09-11
fix(carousel): fix rtl full screen display - FRONT-4081

emeryro:


planctus (commented):
mmmh..ok, but i think we can take the chance to fix some other issue with this story:

- size and demo grid content need a mapping or a fix to their mapping, they are not kept in permalinks

emeryro:
I added the mapping for the size, but for the boolean I'm not sure what to do.
I checked other stories, sometimes the arg is kept in the url, sometimes not, but I didn't find any pattern to make it work every time...

planctus:
I know it feels silly, and i also think this is a regression in storybook, booleans used to work even without an explicit mapping, in any case it's like this: 
```
      mapping: {
        0: false,
        1: true,
      },
```
:D

planctus:
Oh gosh... now that these mapping work try to get the link with the "grid content" and/or a size set, the carousel misbehaves..but ok, maybe this is for another investigation.. 
image

planctus (approved):

PR #3028, opened 2023-09-11
feat(modal): add scrollable modal - FRONT-3994

emeryro:
- Make the body of the modal scrollable
- Add a new section below the body for fixed content
- Add 2 different sizes for the modal


planctus (commented):
Ehm..not sure what we are doing here:
we made this following a request to get the modal scrollable, not just a "block" inside of it. There is still the chance that the modal will "cut" content that would not be reachable by the user, which i think was the main scope of this operation.

Here we keep basing on specific examples defined by the designers, but the use cases are way more than the ones described by them. Now we have this fixed body with just a checkbox (i'm not really sure i understand why they keep placing form elements inside the modal, which is surely not the standard or default use of it, nothing that we handle on our side, furthermore), if that i used for anything else, then the "scrollable" content might shrink to its miminum size and part of this fixed body could be hidden as in the previous version of the component.

To me this component had a limited sense before and it keeps being something else than a "generic" modal component, something kind of tailored to a certain use case, which we don't even know the existence of.

So, i don't know how to review this, let's say that if the designers are happy so we are..? :)

PR #3029, opened 2023-09-12
feat(website): update documentation - FRONT-4045

emeryro:


src/website/src/pages/ec/utilities/spacing/docs/usage.mdx
```diff
@@ -16,13 +16,13 @@ Different clases can be used to set the margin around an element. All spacing to
...
 Available values are:
 
-- `ecl-u-ma-*` (ecl-u-ma-2xs to ecl-u-ma-4xl): all around margins
-- `ecl-u-mv-*` (ecl-u-mv-2xs to ecl-u-mv-4xl): vertical margins
-- `ecl-u-mh-*` (ecl-u-mh-2xs to ecl-u-mh-4xl): horizontal margins
-- `ecl-u-mt-*` (ecl-u-mt-2xs to ecl-u-mt-4xl): top margins
-- `ecl-u-mr-*` (ecl-u-mr-2xs to ecl-u-mr-4xl): right margins
-- `ecl-u-mb-*` (ecl-u-mb-2xs to ecl-u-mb-4xl): bottom margins
-- `ecl-u-ml-*` (ecl-u-ml-2xs to ecl-u-ml-4xl): left margins
+- `ecl-u-ma-*` (ecl-u-ma-2xs to ecl-u-ma-6xl): all around margins
```
planctus, line 19: reading this docs i went then checking the utilities we generate, thinking that maybe we were still using left or right there, but instead we do use inline-start and inline-end, so why don't we advertise this anywhere? I think we can still use the same "labels" like "left margins" for clarity, but on top we can specify that our spacing utilities are RTL compliant.

src/website/src/pages/ec/utilities/background/docs/code.mdx
```diff
@@ -10,13 +10,19 @@ import { Playground } from '@ecl/website-components';
...
-  <div className="ecl-u-type-m ecl-u-bg-white">.ecl-u-bg-white</div>
-  <div className="ecl-u-type-m ecl-u-bg-blue ecl-u-color-white">
-    .ecl-u-bg-blue
-  </div>
-  <div className="ecl-u-type-m ecl-u-bg-yellow">.ecl-u-bg-yellow</div>
-  <div className="ecl-u-type-m ecl-u-bg-grey ecl-u-color-white">
-    .ecl-u-bg-grey
-  </div>
-  <div className="ecl-u-type-m ecl-u-bg-grey-25">.ecl-u-bg-grey-25</div>
+  <p className="ecl-u-type-m ecl-u-bg-white">
```
planctus, line 13: I'm just a bif confused by the misalignment with the storybook demo we have for this, there we show also "error" and "success", shouldn't we put in both also the other semantic colors? warning and info

src/website/src/pages/ec/utilities/shadow/docs/usage.mdx
```diff
@@ -17,24 +17,6 @@ Available values are:
...
-
-By using classes `ecl-u-shadow-negative-*` and `ecl-u-shadow-negative-inner-*` you can manage negative shadow of an element (for dark background).
-Available values are:
-
-- ecl-u-shadow-negative-1 (elevation 1)
-- ecl-u-shadow-negative-2 (elevation 2)
-- ecl-u-shadow-negative-3 (elevation 3)
-- ecl-u-shadow-negative-inner-1 (depth 1)
-- ecl-u-shadow-negative-inner-2 (depth 2)
+- ecl-u-shadow-6 (elevation 6)
```
planctus, line 20: this is a personal consideration, i don't understand this "scale", why is not simply 1-2-3-4, it feels a bit randomish to pull an existing level of shadow like this..

src/website/src/pages/ec/guidelines/spacing/index.mdx
```diff
@@ -12,53 +12,36 @@ import { Paragraph, RemToPixels } from '@ecl/website-components';
...
-
 ## Use of space
 
-Space is used to create hierarchy and visual relationships. By using the space tokens, you can achieve visual grouping and other associations both between and within the components. Along with the typography system, they contribute to an overall visual hierarchy, providing users with a structured way of organising information.
+For spacing tokens to be able to conform in the baseline grid and contribute to the vertical rhythm, all sizes must be multiples of <RemToPixels rem="0.25" />px. A vertical rhythm starts from a baseline and it's the basis for everything from the font sizes, line height and image sizes. The spacing tokens always referring to fixed values and thus they are not responsive. All spacing tokens can be used as values for both margins and paddings.
```
planctus: I'm not sure the reason why here we use the world "token" to mean our utility classes, not that is necessarily wrong per se, but we only do this here in the spacing and so it suprises me.

planctus (commented):
Only a couple of doubts about these utilities pages/demos, nothing crucial.

src/website/src/pages/ec/utilities/shadow/docs/usage.mdx
emeryro: I think we discussed this already, but that's how the shadows are defined at design level: the "elevation" is kind of the z-index for them. And they wanted to keep the possibility to add other shadows in-between
image
Now, we could use 1, 2, 3, 4 for our utilities, but then they won't be in sync with the terminolgy they use, or with the sass maps

src/website/src/pages/ec/guidelines/spacing/index.mdx
emeryro: this should be some remains of the time where we used design tokens. I'll rephrase it

src/website/src/pages/ec/utilities/background/docs/code.mdx
planctus: the control in storybook has some misalignment between the options and the labels, but it also need a mapping if we want the paramters to generate "permalinks".

planctus (approved):

PR #3038, opened 2023-09-18
feat(file): Ec revamp for v4 - FRONT-4089

planctus:


src/implementations/twig/components/file/file.html.twig
```diff
@@ -4,7 +4,7 @@
   Parameters:
   - "variant" (string) (default: 'default') Name of the variant [default, thumbnail]
```
emeryro, line 5: we are still passing the `variant` parameter to the other twig files, but I think that it is no longer used there

src/implementations/vanilla/components/file/file-print.scss
```diff
@@ -85,27 +94,27 @@ $_image-width: 13.125rem;
 .ecl-file__image {
```
emeryro, line 98: on print the image margin seems to be on the wrong side
image

emeryro (changes requested):
I spotted a small display issue on focus, for EU. I guess that it missing the align-self css
image

Also, we should add a line in the documentation for the new title structure

emeryro (approved):

PR #3043, opened 2023-09-25
feat(list): add indentation - FRONT-4059

emeryro:


planctus (commented):
We are removing our padding that was limiting the indentation of the list, so we'll let the browser set its own, which might or not be consistent between differnt ones, i see that chrome and firefox are adding 40px, wouldn't it be better to set this value in our css so that in all the browsers lists would look the same?

planctus (approved):

PR #3053, opened 2023-10-03
feat(file upload): ECL4 file upload - FRONT-4102

emeryro:


planctus (changes requested):
We need to fix the link to playground from the usage pages in EC and EU

PR #3054, opened 2023-10-03
feat(datepicker): ECL4 datepicker - FRONT-4107

emeryro:


planctus (commented):
It feels to me that we should do something to change the approach we used in v3, moment.js is only an optional dependency of pikaday and it is used only in certain cases, in our scenario it becomes a mandatory dependency because we set the format option when we create the pikaday object.

`his.picker = new Pikaday({
      field: this.element,
      format: this.format,
      yearRange: this.yearRange,
      firstDay: 1,
      i18n: this.i18n,
      theme: this.theme,
      reposition: this.reposition,
      ...`
An easy win could be to only pass this option to pikaday when the user customised it (i guess this is rarely done), if we do this we would make moment optional also in ECL, not forcing the users to include it wherever they use ECL.
Since this approach would not be harming current implementations, we could even apply this to v3.
I've also made an investigation regarding the moment.js versions, but it seems that the last version that works with pikaday is the one we currently use.

emeryro:
I would gladly ease the dependendy we have to moment.js, as it triggered lots of support already.
But I don't fully understand what you propose as quick win. We change the format ourselves, to set it to 'DD-MM-YYYY' by default. This already requires to have moment.js, no?

planctus:
Oh yes, you are right, i wrongly assumed we were using the "default" format.
Might not be the "cleanest" way, but looking around i've found out that this format can be set without involving moment.js, like this:
`      toString: date => {
        const day = ('0' + date.getDate()).slice(-2);
        const month = ('0' + (date.getMonth() + 1)).slice(-2);
        const year = date.getFullYear();
        return day + '-' + month + '-' + year;
      },`

and it seems to work fine for our default format. So we could use the format option in case a custom value is passed by the user, or this one in case is not.

emeryro:
I'd suggest doing this update in a follow-up task, so we don't delay the component itself.
It would also give more visibility to this change...

planctus (approved):

PR #3060, opened 2023-10-12
feat(css): add css variables - FRONT-4105

emeryro:
Add sets of css variables for
- typography
- spacing
- shadow

Note: this increased the size of the css files quite a lot...


planctus:
At first glance i didn't understand the reason why of the increased size of the compiled css, but considering the names of our custom properties then it became understandable.
Why don't we create aliases for those custom props and then we use the alias everywhere?
EX:--ecl-spacing-xl could be aliased in s-xl
doing this sistematically would limit the size of the compiled css while still giving the same level of flexibility, with the only downside of seeing in the "inspector" the aliased names.

PR #3061, opened 2023-10-13
feat(select): EC Revamp for V4 - FRONT-4106

planctus:


docs/Migrating-v4.md
```diff
@@ -182,6 +182,13 @@ EX: `{% include '@ecl/form-group/form-group.html.twig' with {
 - Search button is now using existing variant of the button; it was previously using class `ecl-button--search`, which has no definition.
   Please note that EC search form uses ghost button, while EU button uses primary button
 
+### Select
+
+- Both "variants" are now using javascript. For the single an multiple select a selector `ecl-select--active` is toggled in order to apply a transformation to the toggle icon.
```
emeryro:
```suggestion
- Both "variants" are now using javascript. For the single and multiple select a selector `ecl-select--active` is toggled in order to apply a transformation to the toggle icon.
```

emeryro (changes requested):
There is an issue on the single select: when trying to open it with the mouse it immediately closes

Also the keyboard navigation ha changed for select multiple: before, the user put the focus on the select, and decided to open it or no, now it is open immediately when the focus is on it.
I think that this will be reported as an issue for accessibility (being forced to go through all the select options)

I do fear that taking over the default select behaviour would add quite a lot of complexity in terms of maintenance...

planctus:
Was the issue you've found really about the single select..? There was a misbehaviour in the multiselect where two operations were executed in a sequence doing the opposite, this has been fixed.

There are some many listeners that it's easy to see them overlapping in certain cases, for the default select, though, we don't really do much, we basically handle the selector we use to style the icon, for the rest is pretty much the standard experience.

The fact that the multiselect dropdown was opening with a simple Tab was on purpose, at least this seemed the expected behaviour on the design side, but i reverted it and now it's just like it used to be, just that now we react to both Enter and Space, other behaviours related to the navigation with the keyboard have been adapted.

emeryro:
This is fine for the multi select
For the single select, I still have some unconsistent behaviour, mostly for the icon display
2023-10-17-select
Also when using the keyboard, the icon is flipped even if the dropdown is not open, but I guess that we cannot do much about that.

These are not blockers, but to be checked with the designers if we want to keep this approach

planctus:
Right, the behavior of the single select was initially implemented basing on the focus event, being the click not reliable enough, now i changed this into a combination of mousedown and click plus the keyboard so that we can have a consistent behaviour even when navigating with the keyboard:
you first focus on the select, then when you open it you get the flipping icon, when you close it the icon goes back to its original position.

emeryro (approved):

PR #3063, opened 2023-10-16
feat(datepicker): Removing dependency from moment.js, removing pikaday from the bundled js - FRONT-4111

planctus:


src/implementations/vanilla/components/datepicker/datepicker.js
```diff
@@ -87,11 +84,16 @@ export class Datepicker {
    * Initialise component.
    */
   init() {
+    if (typeof window.Pikaday === 'undefined') {
```
emeryro, line 87: shouldn't we exit this function if pikaday is not available?
Otherwise it will still break the ecl js

README.md
```diff
@@ -90,10 +90,16 @@ The ECL is bundled in various [presets](docs/presets.md) in order to accomodate
...
 
-ECL uses [Pikaday](https://github.com/Pikaday/Pikaday) which requires [moment.js](https://momentjs.com/) and this library is not bundled by ECL.
-Therefore **moment.js needs to be loaded or bundled by the application or website using ECL**, depending on the needs.
+ECL uses [Pikaday](https://github.com/Pikaday/Pikaday) and this library is not bundled anymore by ECL.
+Therefore **pikaday needs to be loaded or bundled by the application or website using ECL**, depending on the needs, it is only required when
+a datepicker instance is present in a webpage.
+Additionally, when customising the date format used by the datepicker, in order to get a consistent output [moment.js](https://momentjs.com/) is also needed, the order of the scripts in this case should be:
+
+- moment.js
+- pikaday
```
emeryro: we could add a link to the cdn for pikaday (and moment.js), to make things easier for users

src/implementations/vanilla/components/datepicker/datepicker.js
planctus: there is a `throw new TypeError` inside this if ;)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw

emeryro:
Ok that's fine for me.
We should just add a paragraph or a link in the v4 migration guide, as it could be a breaking change

emeryro (approved):

PR #3067, opened 2023-10-17
feat(list-illustration):  EC revamp for v4 - FRONT-4112

planctus:


src/implementations/vanilla/components/list-illustration/_list-illustration.scss
```diff
@@ -24,8 +18,8 @@ $_icon_color: null !default;
 .ecl-list-illustration__item {
   display: flex;
   flex-direction: column;
-  padding-bottom: map.get(theme.$spacing, 'm');
-  padding-top: map.get(theme.$spacing, 'm');
+  padding-bottom: map.get($theme, 'spacing', 'm');
```
emeryro: we should use the aliases

src/themes/ec/_variables.scss
```diff
@@ -573,6 +573,29 @@ $list: (
...
+  background-color: var(--c-n-40),
+  value-font-mobile: var(--f-2xl),
+  title-font: var(--f-l),
+  value-font-weight-mobile: map.get($font-weight, 'regular'),
+  value-font-desktop: var(--f-3xl),
+  value-font-weight-desktop: map.get($font-weight, 'regular'),
+  icon-color: var(--c-d),
+);
+$list-illustration-print: (
+  title-color: var(--c-d),
```
emeryro: just a remak: if the colors are the same on print, we could use the variables defined for the screen version, no need to have a specific variable here.
Usually, the only variable that I define for print are for spacing and typography as they are different from screen display

emeryro (changes requested):

emeryro (approved):

PR #3068, opened 2023-10-18
feat(ui): Making pictures clickable when needed - FRONT-4115

planctus:


src/implementations/vanilla/components/content-block/content-block.js
```diff
@@ -0,0 +1,152 @@
+import { queryOne } from '@ecl/dom-utils';
+
+/**
+ * @param {HTMLElement} element DOM element for component instantiation and scope
+ * @param {Object} options
+ * @param {Boolean} options.attachClickListener Whether or not to bind click events
+ * @param (String) options.dataSelector The selector of the element where to attach the click listener
+ * @param (String) options.dataTitle  The selector of the element containing the link
```
emeryro: just a remark: the name of these paramter is not really reflecting their usage (no real data involved)
maybe we could use names like `targetSelector` and `linkSelector` or `titleSelector`

src/implementations/vanilla/components/content-block/content-block.js
```diff
@@ -0,0 +1,152 @@
...
+    const contentBlock = new ContentBlock(root, defaultOptions);
+    contentBlock.init();
+    root.ECLContentBlock = contentBlock;
+    return contentBlock;
+  }
+
+  constructor(
+    element,
+    {
+      dataSelector = '.ecl-picture',
```
emeryro: we could limit the risk of having bad element selected, by using a specific data attributes for pictures which could benefit from this.
For instance, adding `data-ecl-picture-click` only in the card and content item component

src/implementations/vanilla/components/content-block/content-block.js
```diff
@@ -0,0 +1,152 @@
...
+      return;
+    }
+
+    this.title = queryOne(this.dataTitle, this.element);
+    this.linkEl = this.title ? queryOne('a', this.title) : false;
+    if (this.linkEl) {
+      this.picture.style.cursor = 'pointer';
+      const img = queryOne('img', this.picture);
+      if (img) {
+        img.title = this.constructor.convertToFullURL(
```
emeryro, line 80: I think that this may trigger some accessibility remarks. The title (and so the url) will be read as is by screen readers.

emeryro (commented):
I added a few remarks.
Overall it works fine, but it is still to be decided where to put this code. 
Having it in the content block covers the current needs, but if tomorrow we have to add this kind of link to the banner or featured item for instance, we would have an issue

src/implementations/vanilla/components/content-block/content-block.js
planctus: It came out as a visual thing more than anything else, it felt weird to see the pointer but not seeing any tooltip.
Technically speaking, as long as these images are not focusable, the screen readers would simply skip them, as far as i know, with or without the alt and the title attributes.

In any case for the moment i disabled this title, kept the code in place as an opt-in thingy, though.

planctus:
About the overall choice regarding what component to attach this to, i think we mainly have two choices:

1. the content block
2. the picture

i wouldn't consider applying this from the individual components that might use the functionality, if those are not already using javascript .
Moving it now to the picture would assume that this functionality will be extended to all, or at least most, of the components using it, and the implementation would have to change in order to take into account all the different use cases, so for me let's keep it in the content-block, for now, but we already know that in the future this might change.

emeryro (approved):

PR #3071, opened 2023-10-19
feat(table): ECL4 table - FRONT-4113

emeryro:


planctus:
is this the expected alignment for the headers? 
image

emeryro:
I double checked, they are supposed to be centered in the cell now. I'll do the update

planctus (approved):

PR #3072, opened 2023-10-19
chore(controls): Adding conditions for toggable controls - FRONT-4118

planctus:


src/playground/addons/story-utils/index.js
```diff
@@ -165,6 +168,7 @@ export const getFormControls = (data, type) => {
     name: 'error message',
     type: 'boolean',
     description: `Show ${type} error message (only visible if invalid state is active)`,
+    if: { arg: 'invalid' },
```
emeryro: that doesn't seem to work. Invalid text control is always displayed

emeryro (changes requested):
That's a nice feature!
I just spotted a few cases where it could be improved

- banner: "label" is linked to the CTA button display
- featured item: "position" is only used if there is a picture
- site header: "EU languages" and "non-EU languages" can be hidden if there is no language switcher

src/playground/addons/story-utils/index.js
planctus: yep, tried a few different combinations, just using now "invalid" as a condition for both show_message and invalid_text, at least this works ;)

emeryro (approved):

PR #3074, opened 2023-10-23
feat(multi-select): improve accessiblity - FRONT-4114

emeryro:


planctus (dismissed):

src/implementations/vanilla/components/select/select.js
```diff
@@ -889,14 +902,21 @@ export class Select {
...
         this.clearAllButton.nextSibling.focus();
         break;
 
       case 'ArrowUp':
         this.optionsContainer.lastChild.querySelector('input').focus();
         break;
 
+      case 'Tab':
+        if (e.shiftKey) {
+          this.optionsContainer.lastChild.querySelector('input').focus();
```
planctus: there is one situation where this breaks: if you have selected at least one item and then you use the search and no result is found, then this container becomes the last child of the optionsContainer and it doesn't contain any input, leading to an error for the undefined input. Same happens if you use arrowUp

src/implementations/vanilla/components/select/select.js
```diff
@@ -889,14 +902,21 @@ export class Select {
         break;
 
       case 'ArrowRight':
-      case 'Tab':
         this.clearAllButton.nextSibling.focus();
```
planctus: we are assuming that a nextSibling is going to be in place here, but the close button will only be generated if a label for it is provided, so we cannot assume it is always going to be there. Better probably to put an:
`if (this.closeButton) { this.closeButton.focus() }`

src/implementations/vanilla/components/select/select.js
```diff
@@ -805,6 +805,15 @@ export class Select {
         this.search.focus();
         break;
 
+      case 'Tab':
+        e.preventDefault();
+        if (e.shiftKey) {
+          this.search.focus();
+        } else {
+          this.optionsContainer.querySelectorAll('input')[0].focus();
```
planctus: i wonder..why are we taking over the default, which would be to move the focus to the first input available?
We want to handle the shit+tab, but we are not forced to prevent the default even when only tab is used

src/implementations/vanilla/components/select/select.js
```diff
@@ -923,8 +943,11 @@ export class Select {
         break;
 
       case 'Tab':
-        e.preventDefault();
-        this.input.focus();
+        if (e.shiftKey) {
+          this.closeButton.previousSibling.focus();
```
planctus: Also in this case we cannot assume that a previousSibling is going to be present. The same we do it also with ArrowLeft.

src/implementations/vanilla/components/select/select.js
```diff
@@ -923,8 +943,11 @@ export class Select {
...
+        } else {
+          this.input.focus();
```
planctus, line 1067: a normal select would also collapse the dropdpwn when tabbing on the last element (not the same, in that case is an option), i wonder if we shouldn't do the same.

planctus (changes requested):
Nice idea to handle also the shit+tab, but we introduced some potential issue, some others were probably already in place.
There is also a styling issue, the element is taking the full width and keeps being "clickable", i am using inline-block in the pr for v4.

src/implementations/vanilla/components/select/select.js
emeryro: that's true, but  with the default tab behaviour it will "tab" out of the multi select when reaching the last item. With the custom behaviour we have it should loops back to the top.
Now that I see this, this code will trigger an error if there are no options... I'll add a test

src/implementations/vanilla/components/select/select.js
emeryro: we may keep the default behaviour for the tab, as you suggested, and keep the custom one for the arrows... I'll do a test

emeryro: There are already some differences between Chrome and Firefox here; Firefox adds some uneeded focus elements when using tab.
I didn't check with other browsers, but as we already have the logic in place to handle the focus with arrow, it would be safer to use it also for tab

emeryro:
I updated the component quite a lot, to add various accessibility improvements.
Hopefully it would be at least "good enough" now :)

planctus (dismissed):
I think we need to add the "markup changes" label in this pr...will we ever stop writing code for this component, i wonder..? :)

PR #3076, opened 2023-10-24
feat(theme): remove theme "dev" - FRONT-4103

emeryro:
Remove generic theme, to keep only EC and EU


src/implementations/vanilla/components/text-area/package.json
```diff
@@ -7,7 +7,6 @@
   "style": "text-area.scss",
   "sass": "text-area.scss",
   "dependencies": {
-    "@ecl/theme-dev": "3.11.0",
     "@ecl/vanilla-layout-grid": "3.11.0"
```
planctus: we are now using the new @ecl/vanilla-layout-form, should be listed here as well and it seems the same for all the form elements. About the new package maybe we can remove the mixins folder and access it simply as @ecl/vanilla-layout-form/mixins

src/implementations/vanilla/utilities/border/package.json
```diff
@@ -10,7 +10,6 @@
     "access": "public"
   },
   "dependencies": {
-    "@ecl/theme-dev": "3.11.0",
     "@ecl/vanilla-layout-grid": "3.11.0"
```
planctus: ok, we don't use theme-dev anymore, but now we should put theme-ec here. These applies to all the utilities package, basically.

src/implementations/vanilla/utilities/spacing/spacing.scss
```diff
@@ -1,5 +1,5 @@
 @use 'sass:map';
-@use '@ecl/theme-dev/theme';
+@use '@ecl/theme-ec/theme';
```
planctus: Mmhh..this is confusing, might be better to first finalise the other pr and integrate it here in the utilities.

planctus (commented):
Cool, it's just  a bit confusing with the components not yet "updated", i've left comments only on some dependency missing. An issue at the moment are the changes in the font-styles in the website, but i guess that the idea would be to review those later.

emeryro:
I noticed an issue here: css variables from both EC and EU are present in the main EU css. This is causing display issues, as the EC variables are called last.
I have no idea yet what is causing that...
EC css is fine

emeryro:
Issue solved for EU, I "just" had to update the component not ported to ECL4 yet, to remove dependency to theme-ec.
That's some time saved for when we will have to do them.
@planctus if you could have a final look at it with these changes, it should be good

yarn.lock
```diff
@@ -1768,6 +1768,11 @@
   resolved "https://registry.yarnpkg.com/@discoveryjs/json-ext/-/json-ext-0.5.7.tgz#1d572bfbbe14b7704e0ba0f39b74815b84870d70"
   integrity sha512-dBVuXR082gk3jsFp7Rd/JI4kytwGHecnCoTtXFb7DB6CNHp4rg5k1bhg0nWdLGLnOV71lmDzGQaLMy8iPLY0pw==
 
+"@ecl/theme-dev@3.11.1":
```
planctus, line 1771: i noticed this..this is because we didn't clean up the package.json of the preset-dev package, which still has all the deprecated components as dependecies. More than doing this, i would completely remove also the preset-dev, the css is not used, the js is still being used but i'd say we can accept to duplicate the exports in each preset and get also rid of this "legacy" preset.

planctus (commented):
Great, but it seems to me we can go even deeper and remove also the preset-dev package at this point ;)

PR #3083, opened 2023-10-27
chore(js): Storing instances in their init() in a Map object - FRONT-4130

planctus:
Pretty straightforward:

- we are storing now the instances in a Map object that gives us an easier way to retrieve them
- the way the ECL.components Map is populated has changed, instead of being an operation performed by the autoInit() module it is delegated to the single component classes, so that it works both when using autoInit() and when instantiating manually
- An update method has been added in the select js, once the instance is retrieved this can be used instead of destroying and re-initing again.

To test this you can retrieve an instance in the console and then destroy it.
For instance in the accordion story it would be:
`const accordion = document.querySelector('.ecl-accordion'); const instance = ECL.components.get(accordion); instance.destroy(); `
The result should be a "broken" accordion :)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,6 +3,9 @@ module.exports = {
   root: true,
   plugins: ['jest'],
   extends: ['airbnb-base', 'prettier'],
+  globals: {
+    ECL: 'writable',
+  },
   env: {
     node: true,
     browser: true,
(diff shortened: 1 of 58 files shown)


emeryro (changes requested):
Accordion and datepicker js are missing the `throw new TypeError(...)`

Apart from that, this is fine

emeryro (approved):

PR #3088, opened 2023-10-31
feat(utilities): use css variables - FRONT-4123

emeryro:


planctus (changes requested):
The typography showcase has examples for the text color utilities.
At the moment those examples do not work because the markup uses the "old" color utilities.
A proposal: what if we move the utilities to a separate css file? It's true that at the moment we are generating more utilities than needed for the colors, but it's like 150kb of the current size are taken by these selectors and it seems really unconvenient to force all the users to load a bigger css file with things that might not ever be used.

emeryro:
Showcase on website has been fixed.
Putting the utilities in a dedicated css is a good idea. I'd suggest to do that in another task, to make it more visible

planctus (approved):

PR #3092, opened 2023-11-06
feat(print): Finalizing iteration - FRONT-4127

planctus:
Not the easiest thing to test, what i did was to create a simple html page where i reproduced the problematic markup so to be able to test the results of the changes i made.
Affected components:

- accordion, fixed the wrong page-break-inside: avoid in the item's content
- featured item, fixed the wrong weight set via css
- fact and figures, using columns also in print
- media-container, removing max-height for the print, it was causing the overlapping of the image when used in the accordion
- link, improved the logic to add the pseudo element with the href value, it is now expected to work also for standard links when used inside an .ecl container, if the default css is in use
- link, changed styles for the exposed url


src/implementations/vanilla/components/fact-figures/fact-figures-print.scss
```diff
@@ -43,20 +35,46 @@
...
-    #{map.get(theme.$icon-print, 'm')} + #{map.get(theme.$spacing-print, 'm')}
-  );
   margin-top: map.get(theme.$spacing-print, 's');
 }
 
 .ecl-fact-figures__view-all {
   display: none;
 }
+
+@include breakpoints.up('l') {
```
emeryro: does it really make sense to use breakpoints in a print css? 
We usually just have the code for "desktop", as this is what will be displayed in a pdf/paper

src/implementations/vanilla/components/link/_link-print.scss
```diff
@@ -14,13 +14,15 @@ $_negative-color: null !default;
...
   text-decoration: none;
-}
 
-.ecl-link::after {
-  content: '(' attr(href) ')';
-  font-style: italic;
-  margin-inline-start: map.get(theme.$spacing-print, '2xs');
-  text-decoration: underline;
+  &::after {
+    color: map.get(theme.$color, 'black-100');
```
emeryro, line 19: we should set it to white for negative links
image

emeryro (changes requested):
primary and secondary links seems to be broken on print (this is already the case on prod)
image

emeryro (approved):

PR #3097, opened 2023-11-09
feat(news-ticker): ECL4 news ticker - FRONT-4138

emeryro:


planctus (commented):
There is a "glitch", a wrong alignment of the first item in storybook (only), which makes me think that this might not be a real issue.
image

emeryro:
I tried too, and I can't reproduce it outside of Storybook, and even in Storybook it works fine when displayed in full screen.
So I guess that it is indeed just a sowcase glitch

PR #3101, opened 2023-11-13
feat(gallery): EC revamp for - FRONT-4137

planctus:


src/themes/ec/_variables.scss
```diff
@@ -547,23 +547,23 @@ $file-upload: (
...
-  description-icon: map.get($icon, 'xs'),
-  description-icon-mobile: map.get($icon, 'xs'),
+  actions-margin: var(--s-xl),
+  actions-margin-desktop: var(--s-xs),
+  arrow-margin: var(--s-m),
+  arrow-margins: var(--s-xs) var(--s-2xs),
+  close-button-color: map.get($color, 'white'),
+  counter-border: 1px solid map.get($color, 'white-100'),
+  counter-font-weight: map.get($font-weight, 'regular'),
+  description-background: rgba(map.get($color, 'dark-100'), 0.7),
```
emeryro, line 554: we should use the newly introduced css variables for overlay colors (--c-ov-l and --c-ov-d)

emeryro (changes requested):

src/themes/ec/_variables.scss
planctus: mmh..these colors are only available in EC, anyway, i've used the dark color as the background, not sure how and where to use the light color instead. I've simplified the variables, the styles are pretty similar only a few of them were still used to set different styles in the two themes.

emeryro (approved):

PR #3102, opened 2023-11-13
feat(social media follow): ECL4 Social Media Follow - FRONT-4142

emeryro:


planctus (commented):
I think it's better if we disable the autoResize in EC and EU in the showcase, opening and closing the popover is currently creating issues.

planctus:
it seems there is a small glitch in the popover arrow: 
image

emeryro:
Auto resize has been removed.
For the popover, I did notice that too, but the component hasn't been updated to ECL4, so for now I just quickly added some shadow to make it visible, but we would have to handle it anyways (the same display is already used without issue in the site header)

PR #3103, opened 2023-11-13
chore(css): Separating utilities - FRONT-4134

planctus:
There is something that annoys me, because of the typography utilities we get a duplicated list of custom properties in the utilitiess css that would not be needed. I couldn't find a way to access exclusively the $typography map inside the variables file.

It seems this will reduce a lot the size of the main css, but per se the utilities css is too big, do we really need to ship responsive utilities for everything? I think we could reduce their amount significantly. 

The process has not been replicated for the print css, too many css files might just confuse the uers.


emeryro:
If the typography variables are an issue as is, we could split the variable file to put the utilities related variables alone.

Reducing the number of utilities using breakpoint would indeed greately reduce the file size, but I have no idea if these utilities are currently used or not...

And one last remark: shouldn't we put the utilities css in the "optional" folder?

planctus:
- Ok, i moved the definition of the custom properties in a separate file, so to have them only in the main css.

- I've also moved this new css into the optional folder.

- To me, i do not expect anyone to be using these utilities at that level of granularity, but even in case someone is, if we remove this responsiveness for everything they would have the chance to adapt their markup, this is a major version and i wouldn't keep carrying something heavy for all the users only for a potential, minimal subset of them. In any case let's postpone this, i mean if we do it let's do it in a separate ticket.

emeryro (approved):

PR #3109, opened 2023-11-21
feat(splash-page): Add splash page component - FRONT-4146

emeryro:
Add new component for splash language page

Also, the language list in the site header has been slightly updated to match the design


planctus:
I understand that we were not "prepared" to a reintroduction of the splash page for the language switcher, so we started building a template for a language switcher to be used only in the site header and its markup has been defined as an element of the parent component.
But now that we are re-introducing this, wouldn't it be better to re-use the same "generic" template and the same css as far as it is possible? Splash plage could still be a template on his own, if needed, but including the language-switcher.  This way we are duplicating a lot of code and it seems to me that the two "variants", splash and site header overlay are basically the same, but maybe i'm wrong..

emeryro:
I asked myself the same thing when starting of the task.
In fact, I think that it would be complex to have the same code (even css) for both. There are additional items on the splash page, and in the site header there is a whole extra layer of js/css to handle the dynamic column display and popup position. Print display is also present only for the splash page.
Knowing the designers, we may aslo have design update for one or the other only.
So we do have some code duplicates indeed, but I think that is the safest way to go for now

emeryro:
I just opened a new task for the site header language switcher (some spacing and typography to be updated).
When doing it we can evaluate how much we could put in common

src/website/src/pages/ec/components/splash-page/docs/code.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: Showcase
+order: 2
+---
+
+import { Playground, Html } from '@ecl/website-components';
+
+import { splashPage } from '../demo';
+
+## Default splah page
```
planctus:
```suggestion
## Default splash page
```

src/website/src/pages/eu/components/splash-page/docs/code.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: Showcase
+order: 2
+---
+
+import { Playground, Html } from '@ecl/website-components';
+
+import { splashPage } from '../demo';
+
+## Default splah page
```
planctus:
```suggestion
## Default splash page
```

src/website/src/pages/ec/components/splash-page/index.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Splash page
+defaultTab: usage
+status: ready
+---
```
planctus, line 5: i'm not sure the reason why we add, but only sometimes, the "tab" linking to storybook. i find it handy if you want to save a click when you know you want to reach the playground.

emeryro: I agree, we should have it for every components. I'll add a small task for that

PR #3111, opened 2023-11-23
feat(carousel): Revamp for v4 - FRONT-4149

planctus:


src/implementations/twig/components/button/button.html.twig
```diff
@@ -50,6 +54,12 @@
   {% endfor %}
 {% endif %}
 
+{% if icon.name is defined and icon.name is not empty %}
+  {% set _icons = [icon] %}
+{% elseif icon[1] is defined and icon[1] is not empty %}
```
emeryro, line 75: If I understand correctly, the button can now have an array of icons. This is worth mentioning in the doc and in the data model commented

emeryro (changes requested):
One remark: there is no indication of focus on the carousels "dots". It does work, but user does not know when the focus is on this area. This may be reported as an accessibiliy issue

planctus:
About the remark on the indicators, it was explicitely requested not to have any particular style for focus, this is due to the fact that in keyboard navigation, moving the focus is also selecting the slide, quoting here the message by the designer:

> The tab user will switch automatically from slide when navigating with the arrow which means that a dedicated focus style isn't needed and that we can used the selected slide also for the focus one.

src/implementations/twig/components/button/button.html.twig
planctus: yep, it's the same we do in the link template. I think it would not be too late, anyway, to actually move the extra_classes always inside the icon named object, or even better removing completely this object

emeryro (commented):
Is the carousel supposed to be that big by default now?
image

Also the pagination color on EU seems wrong
image

planctus:
No, but i don't know how you got that, maybe resizing..? It looks like it kept the height it had in a smaller viewport but it shouldn't happen, although it can be noticed the re-adaptation of the banner heights based on the tallest when you do that.
For EU we are waiting for requirements..

PR #3112, opened 2023-11-23
feat(menu): improve accessibility - FRONT-4126

emeryro:


src/implementations/twig/components/menu/menu.html.twig
```diff
@@ -183,10 +183,11 @@
       
       <ul class="ecl-menu__list" data-ecl-menu-list>
       {% if items is defined and items is not empty and items is iterable %}
-        {% for item in items %}
+        {% for key, item in items %}
         {% if item is defined and item is not empty %}
           {% include '@ecl/menu/menu-item.html.twig' with {
             item: item,
+            id: "ecl-menu-item-" ~ key,
```
planctus: mmh..but what happens if we have two menus in a page? I understand that this might not be recommended or expected, but i think we should play safe adding a randomized string here

src/implementations/vanilla/components/menu/menu.js
```diff
@@ -752,6 +755,7 @@ export class Menu {
         }
       }
       if (e.key === 'ArrowRight' || e.key === 'ArrowDown') {
+        e.preventDefault();
```
planctus, line 758: I think we still miss some preventDefault() in this method, arrow down and up on a sublink move the whole page, but maybe we can do it for all the cases at once if we are completely taking over the default behavior.

planctus (commented):
Just a couple of comments, dealing with generated ids is risky, better avoiding the risk of duplicated ones..

emeryro:
id has been randomized by default, with the possibilty to set it directly (needed to run the tests)
for the preventdefault, it seems that it does not work because we are using "keyup" for the event in the menu. Switching it to "keydown" does remove the screen scroll, but introduce some other issues (random "double action", focus getting out of the sub menu, ...)
So for now I would suggest to keep it as a possible improvement, if there are some remarks, but I'd rather not have to debug the whole keyboard navigation

planctus (approved):

PR #3113, opened 2023-11-23
feat(content-item): add image top variant - FRONT-4155

emeryro:


planctus:
wouldn't it be better to show these examples in a smaller viewport like we do for the card? It's really weird to see like the image moving from left to the extreme right of the viewport and it becomes huge when using top as the position.

emeryro:
To keep the possibility to have horizontal list (using the control), the viewport width should be >996px.
So it wouldn't really change the default display, but would make it harder to resize the window
Here is the viewport at 1000px
image

PR #3118, opened 2023-11-27
feat(icon): use css cutom properties - FRONT-4152

emeryro:
- use css custom properties for icon color


emeryro:
@planctus this PR was kept as draft for some reason, but the change is still relevant

planctus (approved):

PR #3120, opened 2023-11-27
chore(markup): Replacing h1 with divs - FRONT-4160

planctus:


emeryro (commented):
This is fine for the content block and all associated components.
But for the page header I'm not sure if we should remove it also. It is supposed to be a unique element at the top of the page.

planctus:
It was written generically  in the ticket "the components using h1". Will ask for confirmation.

emeryro (approved):

PR #3123, opened 2023-11-28
feat(tag): Add tag set and wrapping - FRONT-4156

emeryro:
Also improve consistency of twig pramter for icon path


src/implementations/twig/components/tag/tag-set.html.twig
```diff
@@ -0,0 +1,48 @@
+{% apply spaceless %}
+
+{#
+  Parameters:
+  - "items" (array) (default: []): array of Tag
+  - "default_icon_path" (string) Path to the icons sprite. If provided here, it will be used for every tag in the set
```
planctus: we do use icon_path everywhere, we should actually "fix" the datepicker that currently uses icons_path, i wouldn't introduce an inconsistency about this parameter.

planctus:
I'm not sure about the decision of creating a separate template for this looping thing. But if we keep it like this it might be better to:

-  Create separate story files to make it more evident that this is a component/template on his own
-  Alternatively we should create a notes file for this tag set, import that and use it in the tag set story, otherwise we are showing the "notes" for the tag template also in the tag set. Also in the migration notes it should be mentioned the fact that this is not only a different data set but also another template to be called.

src/implementations/vanilla/components/tag/_tag.scss
```diff
@@ -47,6 +47,11 @@ $tag: null !default;
   }
 }
 
+// No wrap
```
planctus, line 50: i understand that this was the request in the ticket, but have we analyzed the impact of it? I mean, why are we doing this? We know that this will produce overflowing in many situations, especially because the request came out of the fact that they might have very long string used as tags, in this case the risk is maximized. It is an "opt-in" functionality, so we could say it's a user mistake when this will happen, but it's worth nothing, i see this only as a problematic style with a very limited value in certain specific use cases, maybe..

src/implementations/vanilla/components/tag/_tag.scss
```diff
@@ -47,6 +47,11 @@ $tag: null !default;
...
+.ecl-tag--nowrap {
```
planctus, line 51: but..from what i see we were doing this by default. But the question is: why? :)
I mean, i do see all the potential risks involved in this while i cannot really identify use cases where this might be useful, instead.
We are lowering this risk by setting this as a variant, as opt-in thingy but i'm not sure it still make sense even this way.

planctus (commented):
I think we can improve a bit the story and the migration notes if we are about to introduce a new template here.

src/implementations/vanilla/components/tag/_tag.scss
emeryro: I don't recall exactly, but I think that we added this by default for a reason, possibly coming from EWPP.
So it's probably safer to keep it as a variant for now

planctus (approved):

PR #3130, opened 2023-12-05
feat(tabs): ECL4 tabs - FRONT-4167

emeryro:


planctus:
I wonder if we could try to tackle the accessibility issues that we still have if we re-introduce the accessibility test.
It's not straightforward, though, and it might not lead to a cleaner markup, don't know, but we could try..

planctus:
I see this is coming from a specific rule set for this component, so i guess it's by design, but do we really need this huge padding, i wonder? 
image

emeryro:
I checked the accessibility issue again, and indeed that would not be easy at all to fix. 
It would most probably implies changes of the markup, css and js, even possibly of the behaviour.
I'd rather keep that in a dedicated task, as it is not linked diretly to the design update.

And for the spacing, that's quite huge I agree. But that's what they requested...

planctus (approved):

planctus:
I've just noticed a small issue with the RTL css: 
image

PR #3137, opened 2023-12-13
feat(form): improve accessibility - FRONT-4171

emeryro:
- make helper text and invalid message readable by screen readers
- make sure that form label is also read
- add aria attribute for invalid form
- apply accessibility improvement for multi select


src/implementations/twig/components/form-group/form-group.html.twig
```diff
@@ -36,15 +40,26 @@
 {% set _input = input|default({}) %}
 {% set _width = _input.width|default('') %}
 {% set _id = _input.id|default('') %}
+{% set _group_id = _input.group_id|default('') %}
```
planctus: mmh..but why is this parameter expected in the input and not in the root here? It's hard to expect users to really use this, but it doesn't have much to do it with the rendering of a form input field, it is actually completely unused in these templates so if feels to me that this parameter should belong to the form group only and be passed when needed to the form input templates internally.

src/implementations/twig/components/form-group/form-group.html.twig
```diff
@@ -27,6 +29,8 @@
 {% set _required = required|default(false) %}
 {% set _required_text = required_text|default('') %}
 {% set _optional_text = optional_text|default('') %}
+{% set _label_aria_required = label_aria_required|default('') %}
```
planctus, line 32: I understand that we need to provide accessible markup and this is the way we can do it, but expecting that the users will always be able to set 4 internazionalided strings for setting required and optional text, plus an aria label for the two elements is really optimistic.

src/implementations/twig/components/form-group/form-group.html.twig
```diff
@@ -58,11 +73,21 @@
     disabled: true,
   }) %}
 {% endif %}
+
 {% if _invalid %}
   {% set _input = _input|merge({
     invalid: true,
   }) %}
+  {% if _id %}
```
planctus: this is not clear to me, but it might just be the consequence of defining the group_id as an inner property of the input, here they are basically the result of a duplication, indeed to me this parameter should only be id and we don't need group_id at all, in this case.

src/implementations/twig/components/form-group/form-group.html.twig
```diff
@@ -101,37 +126,66 @@
...
+          class="ecl-form-label__required"
+          {% if _label_aria_required %}
+            aria-label="{{ _label_aria_required }}"
+          {% endif %}
+        >{{ _required_text }}</span>
       {%- elseif not _required and _optional_text is not empty -%}
-        <span class="ecl-form-label__optional">{{ _optional_text }}</span>
+        <span 
+          class="ecl-form-label__optional"
+          {% if _label_aria_optional %}
```
planctus: it seems to me we should add some white space controls here, {%- if _label_aria_optional -%}  {%- if _label_aria_required -%} etc...

src/implementations/twig/components/form-group/form-group.html.twig
```diff
@@ -101,37 +126,66 @@
...
+          {% endif %}
+        >{{ _optional_text }}</span>
       {%- endif -%}
     </{{ _label_tag }}>
   {% endif %}
 
   {% if _helper_text is not empty %}
-    <div class="ecl-help-block{{ _disabled ? ' ecl-help-block--disabled' }}">
+    <div class="ecl-help-block{{ _disabled ? ' ecl-help-block--disabled' }}"
+      {% if _id %}
```
planctus, line 192: All these blocks where we alternatively use _id or group_id to set a value for the ids could be removed in my opinion by simply using _id

src/implementations/twig/components/form-group/form-group.html.twig
```diff
@@ -175,16 +236,20 @@
...
   {%- endif -%}
 
   {%- if _input.input_type == 'rating-field' -%}
     {% include '@ecl/rating-field/rating-field.html.twig' with _input only %}
   {%- endif -%}
 
   {%- if _input.input_type == 'file' -%}
-    {% include '@ecl/file-upload/file-upload.html.twig' with _input only %}
+    {% include '@ecl/file-upload/file-upload.html.twig' with _input|merge({
+      extra_attributes: _input_extra_attributes,
```
planctus: since we do this for all the templates we include, wouldn't it be easier to do it at once instead of adding them in each instance of the includes we have?

src/implementations/twig/components/select/select.html.twig
```diff
@@ -58,10 +58,10 @@
 {% set _css_group_class = 'ecl-form-group' %}
 {% set _css_class = ['ecl-select'] %}
 {% set _css_class = _css_class|join(' ') %}
-{% set _extra_attributes = '' %}
-{% set extra_attributes = [{ name: 'data-ecl-auto-init', value: 'Select' }] %}
+{% set extra_attributes = extra_attributes|default([]) %}
+{% set extra_attributes = extra_attributes|merge([{ name: 'data-ecl-auto-init', value: 'Select' }]) %}
```
planctus: We can define it only once as:
` {% set extra_attributes = extra_attributes|default([])|merge([{ name: 'data-ecl-auto-init', value: 'Select' }]) %}`

src/implementations/twig/components/datepicker/datepicker.html.twig
```diff
@@ -18,12 +18,12 @@
...
-    - "extra_attributes" (array) (default: []): format: [
-        {
-          "name" (string) (default: ''),
-          "value" (optional) (string)
-        ...
-      ],
+    - "extra_attributes" (array) (default: []) Extra attributes for the datepicker wrapper
+      - "name" (string) Attribute name, eg. 'data-test'
+      - "value" (optional) (string) Attribute value, eg: 'data-test-1'
+    - "input_extra_attributes" (array) (default: []) Extra attributes for the datepicker input
```
planctus: Do we need this additional parameter? I mean, what are the extra_attributes for otherwise? For the wrapping div i think we don't really need them and in the end we do expect these attributes to always come as input.extra_attributes, i wouldn't introduce this inconsistency, it feels really unneeded.

planctus (changes requested):
I did not get the reason why of the added parameters, group_id seems to me simply a duplicate of _id and it's not clear to me the reason why expecting this from the data passed to the input, being used in the form-group template, i would simply remove it, then. Also the input_extra_attributes in the datepicker seems to me totally unneeded, i would simply move the rendering of the extra_attributes to the input which seems consistent with all the other templates.

src/implementations/twig/components/form-group/form-group.html.twig
```diff
@@ -58,35 +65,62 @@
...
 {% if _input.input_type == 'file' %}
-  {% set extra_attributes = [
+  {% set extra_attributes = extra_attributes|default([])|merge([
     { name: 'data-ecl-file-upload-group', value: true },
     { name: 'data-ecl-autoinit', value: 'FileUpload' },
-  ] %}
+  ]) %}
+{% endif %}
+
+{% if _helper_text is not empty %}
```
planctus: i think this is not enough, we risk to render an empty aria-describedby like this, we should also check for the _aria_describedby variable for its emptyness.

src/implementations/twig/components/form-group/form-group.html.twig
```diff
@@ -58,35 +65,62 @@
...
+    {% set extra_attributes = extra_attributes|default([])|merge([
+      { name: 'aria-describedby', value: _aria_describedby },
+    ]) %}
+  {% endif %}
+  {% if _invalid %}
+    {% set extra_attributes = extra_attributes|default([])|merge([
+      { name: 'aria-invalid', value: 'true' },
+    ]) %}
+  {% endif %}
+{% elseif _input_extra_attributes is not empty %}
```
planctus: i'm terribly confused by this logic about the input.extra_attributes, this merge here only happens for input_type different than radio, checkbox rating field, but we do set some value for them all when are invalid, for instance.
Most likely this is done on purpose, and these attributes are then set in the input, probably because of some differences in the markup of these input types, if that is the case maybe we could add some comment in the template to explain what happens.

src/implementations/twig/components/form-group/form-group.story.js
```diff
@@ -24,6 +24,7 @@ const dataStandaloneCheckbox = {
   ...dataCheckbox,
   input: {
     ...dataCheckbox.input,
+    group_id: '',
```
planctus: it seems this is a leftover

src/implementations/twig/components/form-group/form-group.html.twig
emeryro: There are two things here: for some input type (checkbox, radio and rating), the extra attribute has to be added to the wrapper element, using `extra_attributes`. 
For the other cases, it is added on the input element, using `input.extra_attributes`.
I'll add some comments

planctus:
We might be improving the accessibility of form elements in the opinion of an "expert", but our accessibility tool is actually reporting a "serious" issue: 
image
In the file upload there is also another warning that might be worth taking care of.

emeryro:
I fixed the reported issues, thanks for noticing
There is still one issue left in the textarea, but I see no reason why it is reported there (text input is just fine). So I guess that it's a false positive

planctus (approved):

PR #3148, opened 2023-12-20
feat(file): add empty alt - FRONT-4186

emeryro:
Also improve file story


planctus (commented):
There is a small glitch in the default story, the label control is visible but it doesn't do anything.

planctus (approved):

PR #3149, opened 2023-12-20
feat(file): Moving aria-expanded attribute to the toggle button - FRONT-4185

planctus:


emeryro (changes requested):
The switch with javascript is working fine, but we are missing the initial aria-expanded="false", before any click
This was already the case before, but better improve it now (similar to the expandable)

emeryro (approved):

PR #3150, opened 2023-12-20
feat(file): Setting aria-labelledby programmatically on download links - FRONT-4184

planctus:


src/implementations/twig/components/file/file-translation-item.html.twig
```diff
@@ -5,6 +5,7 @@
   - "title" (string) (default: '')
   - "meta" (string) (default: '')
   - "lang" (string) (default: '')
+  - "id" (string) (default: random)
```
emeryro: instead of providing a different id for each translation item, we could rely on the id coming from the file template, and add a suffix (something like '[id]-[language-code]' or [id]-1|2|3|...)
Similar to what has been done on the menu
https://github.com/ec-europa/europa-component-library/blob/v4-dev/src/implementations/twig/components/menu/menu.html.twig#L176

src/implementations/twig/components/file/file-translation-item.html.twig
```diff
@@ -46,7 +53,9 @@
       }),
       extra_classes: 'ecl-file__translation-download',
       extra_attributes: [
+        { name: 'id', value: _link_id },
         { name: 'download' },
+        { name: 'aria-labelledby', value: '#' ~ _link_id ~ '  #' ~ _title_id ~ ' #' ~  _lang_id },
```
emeryro: aria-labelledby should use the id directly, without the "#"

also, here the link is refereeing to itself (link_id). I think that it would result in the information being said twice. It would be better to refer to the label link only instead.
If not possible (or too complex), we may try to leave the link as is, and add aria-describedby with title_id and lang_id. It should read the link first, then the additional information

src/implementations/twig/components/file/file.html.twig
```diff
@@ -128,7 +133,12 @@
       </div>
     {% endif %}
     {% if _title is not empty %}
-      <div class="ecl-file__title">
+      <div class="ecl-file__title"
```
emeryro: if I'm not wrong, the title id is used in any case for the download button. So the id should be there even without translations

src/implementations/twig/components/file/file.html.twig
```diff
@@ -166,12 +176,17 @@
         name: 'download',
         size: 'fluid'
       }),
+      extra_attributes: [
+       { name: 'id', value: _link_id },
+       { name: 'aria-labelledby', value: '#' ~ _link_id ~ ' #' ~ _title_id ~ ' #' ~ _lang_id },
```
emeryro: same remark for aria-labelledby

emeryro (changes requested):
some changes to be done on the markup, and maybe improve the logic for the id.
fyi, I downloaded a free screen reader on my side (NVDA), to be able to check how it work

src/implementations/twig/components/file/file-translation-item.html.twig
planctus: mmh..i see that in the https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby docs there is mention of the id hash, but why, i wonder..anyway i removed them.

I've created a generated id for the label of the link when it contains an icon and i'm using that in the aria-labelledby of the translation items. Good enough?

src/implementations/twig/components/file/file-translation-item.html.twig
planctus: ok, i've changed this id into {file_id}-translation-{key}

src/implementations/twig/components/file/file-translations.html.twig
```diff
@@ -3,6 +3,8 @@
 {#
   Parameters:
     - "toggle" (object) (default: {}): object of type Button
+    - "id" (string) (default: '')
+    - "title_id" (string) (default: '')
```
emeryro: as the title id is just `_id ~ '-title'`, do we need to pass it as a parameter everywhere?
It seems that it could be deduced from the global id when needed

src/implementations/twig/components/file/__snapshots__/file.test.js.snap
```diff
@@ -79,13 +79,14 @@ exports[`File Thumbnail renders correctly 1`] = `
         (16.2 MB - PDF)
       </div>
       <a
-        aria-labelledby="#ecl-file-with-thumbnail-link #ecl-file-with-thumbnail-title #ecl-file-with-thumbnail-lang"
+        aria-labelledby="ecl-file-with-thumbnail-link ecl-file-with-thumbnail-title ecl-file-with-thumbnail-lang"
```
emeryro: the first id should be `ecl-file-with-thumbnail-link-label` instead of `ecl-file-with-thumbnail-link`

emeryro (changes requested):
just a few things left
We should also add it to the documentation, as the data structure has changed a little

emeryro (approved):

PR #3152, opened 2023-12-20
feat(description list): use list structure - FRONT-4183

emeryro:


planctus (commented):
This is fine, but we do have a report from our accessibilty checks about the fact that links should not be distinguished only by using a different color (so all the standalone links when used in a mixed content, not sure what we could do but it's worth mentioning it, also a violation is triggered about the contrast-ratio. 
This is visible once clicking on the see more of the taxonomy list, when this link is present we don't get a violation but only a warning because the contrast ratio cannot be determined probably because of the hidden items.

PR #3153, opened 2023-12-21
feat(form): change feedback message position - FRONT-4189

emeryro:


planctus:
No issue with the pr, but we miss a few args in the story so we don't get the default, i think we can fix this here:

- "args.optional_text"
- "args.required_text"
- conditionally (if data.name) "args.name"

planctus (commented):
Left a comment about the missing args in the story ;)

planctus (approved):

PR #3154, opened 2023-12-21
feat(components): homogenize max width - FRONT-4190

emeryro:


src/themes/ec/maps/layout.scss
```diff
@@ -24,3 +24,4 @@ $grid-gutter-width: (
   'l': map.get($grid, 'gutter'),
   'xl': map.get($grid, 'gutter'),
 ) !default;
+$max-width: 580px !default;
```
planctus, line 27: i find this a bit confusing, we could simply set this value in the custom_properties, then overriding the custom prop would change this for all its occurencies, if you prefer it like this, anyway, no issue.

planctus (approved):
Niiice, just a minor question related to the variable definition.

PR #3161, opened 2024-01-09
feat(popover): Four directions, scroll and resize detection - FRONT-4159

planctus:


src/implementations/vanilla/components/popover/_popover.scss
```diff
@@ -52,17 +48,17 @@ $_item-width: 188px;
...
   }
 }
 
 .ecl-popover__content {
+  background-color: map.get($theme, 'color', 'white');
   font: map.get($theme, 'font', 'm');
   max-height: #{$_item-height * 4};
   min-width: $_item-width;
-  padding: map.get($theme, 'spacing', 's');
+  padding: map.get($theme, 'spacing', '3xl');
```
emeryro: we should use the css variable when possible

emeryro (changes requested):
nicely done!
I'm quite sure that there will be edge cases where the position would not be correct, but for now it seems fine
Just a small remark for the css

emeryro (approved):

PR #3162, opened 2024-01-09
feat(cleanup): remove unused colors, update website styles - FRONT-4195

emeryro:


planctus:
There is one of the examples of the background utilities in EU that appears "transparent": 
image

Furthermore the "white" background utility is actually using --ecl-color-white which seems not to be in place
Also in storybook the demo uses a ecl-u-bg-white class that is not in place.

in the border utilities the "accent-30" is actually getting the default black color: 
image

planctus (changes requested):
i've only found a couple of issues, nice clean up ;)

emeryro:
This is a nasty side effect...
Now that styles are way more different between EC and EU, I am using EC styles for both EC and EU pages on the website.
But then, if a showcase is rendered directly on the page (without iframe), it doesn't get the proper styling for EU
I have to see how this could be solved (maybe using iframes for all the showcase...)

emeryro:
In the end I implemented a compromise: the website now uses its own set of css variables, similar to the variables for ECL EC, but with a prefix.
That way all styling related only to the website does not conflict with the variables used by ECL components.
Showcase is now working fine for both EC and EU

planctus (approved):

PR #3174, opened 2024-01-12
feat(splash): One column full width - FRONT-4203

planctus:


emeryro:
The proposed improvement works fine, but I'm not really about the added value.
Language links have a fixed width currently, so even if we expand the container, it just adds empty space.

planctus:
but at least it distributes the items using the available space instead of creating an empty column on the right, no..?
image

image

In the end it should be a designer, to make this decision, let's present this in the meeting and we'll see

emeryro (approved):

PR #3175, opened 2024-01-15
feat(featured-item): add footer - FRONT-4196

emeryro:


planctus (commented):
For the gallery we've go the request to use a "footer" tag, wouldn't it better to do the same here?

planctus (approved):

PR #3178, opened 2024-01-15
feat(description list): add tags - FRONT-4198

emeryro:


planctus (commented):
It seems all good to me, but let's add this new "type" in the docs of the template and in the README, i would also mention this new type in the migration docs.

emeryro:
In fact, it was already mentioned in the twig and readme; I guess we didn't remove last time we were asked to change that...
Migration doc updated

planctus:
i'm confused, what i see in the template is:
`"type" (string): type of data (text, link, taxonomy)`

but in reality we "support" for the type param:

- taxonomy
- link
- link-inline
- tag
- (anything else) that we consider as "text"

Same thing in the README

planctus (approved):

PR #3179, opened 2024-01-15
chore(js): Adding EventManager, defining some custom events - FRONT-4205

planctus:
In this pr we introduce an helper class to register callbacks for custom events we define and a method to execute them in the context of the right methods in the different components.
A property is added in each component defining custom events to retrieve the list of available events.

To test it, use one of the stories where the event manager is being used: accordion, banner, inpage, select, menu
You can first retrieve the list of available events:

```
const select = ECL.components.get(document.getElementById('select-multiple');
select.supportedEvents;
```
and then you can define a callback for one of the events:
```
select.on('onToggle' = () => {
  console.log('The select has been toggled');
}
```


emeryro:
This works fine, thanks
I think that it may be worth spending some time here to evaluate what may be the most useful callback for our components

emeryro (approved):

PR #3182, opened 2024-01-16
feat(variables): split variables file - FRONT-4206

emeryro:


planctus:
wow, this is really suprising.. :) I've always been pretty doubtful about the idea of centralizing all the component values in a single file, but more than that to be honest what i don't really like is that is the theme that defines properties that belong to the component styles, instead. If this was acceptable before, now it becomes even more evident and i really don't understand it, also because then we need this intermediate layer to forward all these maps. In my opinion we are basically going back to the v3 approach, which had nothing wrong, but we are moving all the definitions of values for our components in the "theme", which should only be used to assign the right values to properties that belong to each component's definition.

emeryro:
To me, the main difference with the approach we had on v3 is that the components are not tied to one theme or another. As long as you provide the values for sass variables, it could be EC, EU, or any custom theme, it will just work fine.

> we are moving all the definitions of values for our components in the "theme", which should only be used to assign the right values to properties that belong to each component's definition.

Isn't it what we are doing? If I'm not wrong, we are assigning values for component's variable (and also defining all possible style values).
As the values are different between EC and EU, it made sense to have them in the place where we have this difference. 

If you have some ideas to improve this, it is always possible, but what I would like to avoid is the situation of v3, where we had files like `button-ec` and `button-eu` in the components themselves

planctus:
mmhh..ok, i agree that like this the vanilla component is theme "agnostic", but i do see this conflicting with the idea of a theme "agnostic" about the components, instead. I don't see a way to balance the two, i do not like having as many files as the components we have in the theme folder, as i didn't like having a single file to handle all the maps we needed, but i prefer the new approach comparatively. ;)

PR #3194, opened 2024-01-22
feat(tabs): fix accessibility issues - FRONT-4214

emeryro:
Fixed several issues by updating the markup and css (no change to the data structure)

There is one error remaining (for the "more" button), but it appears to be a false positive


planctus:
mmhh...to be honest the real issue we had with this component is that we were not able to successufly run the accessibility test using jest-axe and unfrtunately that is still the case with this new markup:
   Expected the HTML found at $('.ecl-tabs__list') to have no violations:

    <div class="ecl-tabs__list" role="tablist">

    Received:

    "Certain ARIA roles must contain particular children (aria-required-children)"

    Fix any of the following:
      Element has children which are not allowed: button[tabindex]
      Element uses aria-busy="true" while showing a loader

    You can find more information on this issue here: 
    https://dequeuniversity.com/rules/axe/4.7/aria-required-children?application=axeAPI

    ────────

    Expected the HTML found at $('.ecl-tabs__prev') to have no violations:

    <button class="ecl-button ecl-button--ghost ecl-tabs__prev ecl-tabs__item--hidden ecl-button--icon-only" type="button">

    Received:

    "Buttons must have discernible text (button-name)"

    Fix any of the following:
      Element does not have inner text that is visible to screen readers
      aria-label attribute does not exist or is empty
      aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
      Element has no title attribute
      Element's default semantics were not overridden with role="none" or role="presentation"

    You can find more information on this issue here: 
    https://dequeuniversity.com/rules/axe/4.7/button-name?application=axeAPI

    Expected the HTML found at $('.ecl-tabs__next') to have no violations:

    <button class="ecl-button ecl-button--ghost ecl-tabs__next ecl-tabs__item--hidden ecl-button--icon-only" type="button">

    Received:

    "Buttons must have discernible text (button-name)"

    Fix any of the following:
      Element does not have inner text that is visible to screen readers
      aria-label attribute does not exist or is empty
      aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
      Element has no title attribute

So, i'm not sure about what to do, the current markup mighr or might not improve at all this accessibility issue we have with the tabs..

planctus (approved):
Cool!!

PR #3195, opened 2024-01-22
feat(site header): move hamburger on top - FRONT-4237

emeryro:


planctus:
Wow, i wonder why doing this..Was this decision agreed with some of our accessibility experts?
I find it confusing, putting here some concerns:

- when the menu is opened the close button is "duplicated", do we need the "old" close button still?
- we expect two icons passed in a certain order to generate the right markup
- in the standardised site header currently using the keyboard you jump from the the search button to the class name for then reaching the menu toggle icon.
- I'm not sure about the EC/EU differences, the duplicated button is still in place also in EU, and it gets a display: block when the menu is open but it's not visible only due of its z-index, not sure but maybe this was a way to update the markup and the styles in EC without visually affecting EU, but it feels confusing and if we do this for EC i don't see the reason why we wouldn't do it for EU creating this asimmetry.

emeryro:
Not sure if the accessibility experts were consulted here.
As I understood, this is request coming from top management, and the hamburger menu somehow "has to" be placed here for EC header.

Now, for your concerns:
>when the menu is opened the close button is "duplicated", do we need the "old" close button still?

the mobile menu is not updated here, it is done in another PR. The new one does not has the close button (on EC)

> we expect two icons passed in a certain order to generate the right markup

that's true. I'm not really of fan of it either, but that was to use the newly introduce update to use an array of icon in the button.
If we don't want to do that, we could still have 2 different parameters (one for each icon); but that would make the data structure more complex

> in the standardised site header currently using the keyboard you jump from the the search button to the class name for then reaching the menu toggle icon.

true. The menu is positionned after in the markup so it's selected after. ~~We could try to change the markup order, and display it the way we want with css...~~ I don't see how we could change things here, as the focus order is different between mobile and desktop

> I'm not sure about the EC/EU differences

EU behaviour hasn't changed. We don't have specs for some potential new EU site header, so I assumed that it remains as is

emeryro:
Branch has been merged in https://github.com/ec-europa/europa-component-library/pull/3210

PR #3202, opened 2024-01-26
feat(icons): Updating some icons in EC set - FRONT-4152

planctus:


emeryro (commented):
I see that the icon "play-outline" has been added only for EC. Shouldn't we have it also for EU?

emeryro (approved):

PR #3206, opened 2024-01-29
feat(dependencies): review and update all component dependencies - FRONT-4226

emeryro:


src/implementations/vanilla/components/select/package.json
```diff
@@ -9,8 +9,11 @@
   "style": "select.scss",
   "sass": "select.scss",
   "dependencies": {
-    "@ecl/vanilla-layout-form": "4.0.0-alpha.0",
-    "@ecl/vanilla-layout-grid": "4.0.0-alpha.0"
+    "@ecl/builder": "4.0.0-alpha.0",
+    "@ecl/dom-utils": "4.0.0-alpha.0",
+    "@ecl/resources-ec-icons": "4.0.0-alpha.0",
```
planctus, line 14: This made think of something..it's correct to set both the icon packages as dependencies, since we import an icon in the js but here it comes the question: why do we create the svg icon in the script instead of just re-using the one from the real select..? If we do it that way then there is no need to define our icons packages as deps here.

planctus (approved):

PR #3211, opened 2024-02-01
fix(menu): Enabling external icon for first level menu items - FRONT-4238

planctus:


src/implementations/vanilla/components/menu/menu.scss
```diff
@@ -85,6 +85,12 @@ $menu: null !default;
       display: block;
     }
 
+    // For the external icon
+    .ecl-menu__item .ecl-link__label + .ecl-icon {
```
emeryro: Is it coming from the designers? The icon is quite huge here...
image

Also it may be safer to use a css selector not directly linked to the markup structure, like
```
.ecl-menu__link .ecl-icon, 
.ecl-menu__sublink .ecl-icon {
  ...
}
```

planctus: We have to override the size for the external icon when used in the first level menu items, the 2xs set by default by the link template makes it super small, the previous selector used was wrong, basically, targeting also the sublinks, i updated it now using the suggested .ecl-menu__link .ecl-icon.
I think the "m" size is fine for the large viewports, but it is an arbitrary choice, no indication by the designers.

emeryro: ok, let's wait for the designers' feedback

planctus:
ok i will push then a commit to be reverted afterwards, adding an external item in the first level of the menu.

planctus:
i've actually realized that it was all wrong, basically the problem was the flex display of the menu items, the icon was shrinking, no need to override the size which had side effects.

emeryro (approved):

PR #3226, opened 2024-02-08
feat(social-media-follow): add monochrome icons - FRONT-4250

emeryro:


planctus:
Just double checking, i guess this is intended, the pr was already approved by the designers, although this feels a bit weird, i spent time trying to understand why this icon was not getting colored, i've found the color property set on those icons only after a while.. :)
image

emeryro:
yes, this it what has been requested

PR #3230, opened 2024-02-13
feat(site-header): add basket - FRONT-4217

emeryro:


src/implementations/twig/components/site-header/README.md
```diff
@@ -48,6 +48,10 @@ npm install --save @ecl/twig-component-site-header
       - "label" (string) (default: '') Item language label, eg. 'English', 'Français', etc.
       - "path" (string) (default: '') Item language URL eg. '/example#language_en'.
       - "active" (boolean) (default: false) define if item is the active language.
+- **"basket"** (associate array) (default: {}) format:
```
planctus:
```suggestion
- **"basket"** (associative array) (default: {}) format:
```

src/implementations/twig/components/site-header/README.md
```diff
@@ -48,6 +48,10 @@ npm install --save @ecl/twig-component-site-header
...
+  - "id": (string): Id of the basket link
+  - "label": (string): Basket label
```
planctus:
```suggestion
  - "label": (string): Basket link label
```

src/implementations/twig/components/site-header/site-header.html.twig
```diff
@@ -317,6 +324,33 @@
             } only %}
           {% endif %}
 
+          {# Basket #}
+          {% if _basket is not empty %}
+            <div class="ecl-site-header__basket">
+              <a
```
planctus, line 332: i guess there is a reason for this, why not including link here instead of defining the whole markup in this template?

src/implementations/twig/components/site-header/eu-site-header.story.js
```diff
@@ -65,75 +66,139 @@ const getArgTypes = (data) => {
...
     };
   }
   if (data.site_name) {
     argTypes.show_site_name = {
       name: 'site name',
-      type: { name: 'boolean' },
+      control: { type: 'boolean' },
       description: 'Show the site name',
       table: {
+        type: { summary: 'boolean' },
```
planctus: mmh..no, i think this is confusing, we don't have a parameter show_site_name, right? this is just a toggle we use in storybook, so we shouldn't define a "table" for it

src/implementations/twig/components/site-header/ec-site-header.story.js
```diff
@@ -73,85 +74,155 @@ const getArgTypes = (data) => {
   if (data.login_box) {
     argTypes.show_login = {
       name: 'login',
-      type: { name: 'boolean' },
+      control: { type: 'boolean' },
       description: 'Show the login box',
       table: {
+        type: { summary: 'boolean' },
```
planctus: same concept, toggles that we define for storybook only should not have a table set, but in this case we could use "login box" as the label and in that case we could define the table based on the corresponding parameter, which is an empty object by default.

src/implementations/twig/components/site-header/ec-site-header.story.js
```diff
@@ -73,85 +74,155 @@ const getArgTypes = (data) => {
...
+    },
   };
   if (data.notification) {
     argTypes.show_notification = {
       name: 'notification',
-      type: { name: 'boolean' },
+      control: { type: 'boolean' },
       description: 'Show the notification box',
       table: {
+        type: { summary: 'boolean' },
```
planctus: well, it seems you did this sistematically, to me it's better to distinguish between the controls that are matching a parameter and those who are only there for displaying purposes.

planctus (commented):
A couple of "suggestions" and a general remark about the definition of controls that do not have a correspondant parameter.

src/implementations/twig/components/site-header/site-header.html.twig
emeryro: the main reason is that having the class ecl-link would alter the display (color, underline, ...), and so would require some additional overrides

PR #3232, opened 2024-02-14
feat(menu): add dynamic label - FRONT-4251

emeryro:
Update toggle label based on its status (open or closed)


src/implementations/twig/components/menu/menu.html.twig
```diff
@@ -81,6 +81,12 @@
   data-ecl-menu
   data-ecl-menu-max-lines="{{ _max_lines }}"
   data-ecl-auto-init="Menu"
+  {% if _toggle is not empty %}
+    data-ecl-menu-label-open="{{ _toggle.link.label }}"
```
planctus, line 85: this is dangerous, checking for the _toggle emptyness is not enough, _toggle.link could be undefined and it would throw an error

src/implementations/vanilla/components/menu/menu.js
```diff
@@ -893,6 +901,13 @@ export class Menu {
     this.inner.setAttribute('aria-hidden', 'false');
     this.isOpen = true;
 
+    // Update label
+    if (this.toggleLabel) {
```
planctus: I'm confused, this "label" could have already a value set, right? If we do this and these attributes are not set, or they are just empty, we would wipe this value, no? I think we should also check for the non emptyness of the attribute before using it to override this label.

planctus (changes requested):
A couple of remarks..

src/implementations/twig/components/menu/menu.html.twig
planctus: i just wanted to prevent errors, but ok, like this we are not even adding the attribute in case it's not going to have a value, which i think it's better :)

planctus (approved):

PR #3237, opened 2024-02-14
feat(news-ticker): add text for external icon - FRONT-4258

emeryro:
Update to link to allow positionning of external icon before the text (used in news ticker)


src/implementations/twig/components/link/link.html.twig
```diff
@@ -73,10 +73,10 @@
   {% set _icons = [_icon|merge({
     name: 'external',
     path: _link.icon_path,
-    size: '2xs',
+    size: icon is defined and icon.size is defined ? icon.size : '2xs',
```
planctus: do we need to open the room for the external icon size customisation..? Couldn't this happen via css if it's only for a component, instead? I mean, i faced the same situation a while ago, but it seems really irrational to pass an entire object with a size property in order to override the default, feels it's more like an "internal" solution but isn't easier to override this size in the news ticker component, so that it will just be like that, instead of relying on user data?

planctus (approved):

PR #3247, opened 2024-02-19
feat(table): add accessibility scope - FRONT-4259

emeryro:
- add  `scope` attribute to the table header


src/implementations/twig/components/table/table.html.twig
```diff
@@ -100,14 +100,17 @@
...
-              {% set _heading_attribute = _heading_attribute ~ 'rowspan="' ~ cell.rowspan|e('html_attr') ~ '"' %}
+              {% set _heading_attribute = 'rowspan="' ~ cell.rowspan|e('html_attr') ~ '"' %}
             {% endif %}
             {% if cell.colspan is not empty %}
               {% set _heading_attribute = _heading_attribute ~ ' colspan="' ~ cell.colspan|e('html_attr') ~ '"' %}
+              {% set _heading_scope = 'colgroup' %}
             {% elseif cell.label is not empty and _sortable %}
               {% set _heading_attribute = _heading_attribute ~ ' data-ecl-table-sort-toggle' %}
             {% endif %}
+            {% set _heading_attribute = _heading_attribute ~ ' scope="' ~ _heading_scope|e('html_attr') ~ '"' %}
```
planctus: we don't really need to filter the string we have just defined with the escape filter here

PR #3251, opened 2024-02-21
feat(table): add id and headers - FRONT-4260

emeryro:
- New twig parameter `id`, to set a unique id for the table (used for accessibility)
This global id is used to generate a unique id for each of the table headers
Example
```
id: 'table-id',
```
- New twig parameter `headers` for each table header and table cell. It is used to setup a relation between a cell and one or more header, especially for multi header tables.
Example
```
rows: [
    {
      data: [
        {
          label: 'Administators in Competition Law',
          headers: 'table-id-1',
          ...
        },
        {
          label: 'European Commission',
          headers: 'table-id-3 table-id-5',
          ...
        },
    }
]
```


src/implementations/twig/components/table/table.html.twig
```diff
@@ -44,6 +45,7 @@
 
 {# Internal properties #}
 
+{% set _id = id|default(random()) %}
```
planctus, line 49: Isn't it weird that we get this Id but we don't set it on the table itself..? I've created a ticket for better handling randomly generated string when we use them as ids, random() might generate invalid ids.

src/implementations/twig/components/table/table.html.twig
```diff
@@ -24,9 +26,8 @@
           "label" (string or array of string)
           "data-ecl-table-header" (string) (default: ''),
           "group" (bool) (default: false),
+          "headers" (string) (default: ''): table headers id
```
planctus: this sounds confusing, in a couple of ways:
the main object has the same key "headers" and the description is not clearly stating that this referes to existing ids to be referenced here.
Something like: "headers attribute to reference the relevant table headers ids"..?

src/implementations/twig/components/table/table.html.twig
```diff
@@ -100,11 +102,14 @@
...
     <caption class="ecl-table__caption">{{ _caption }}</caption>
     {% endif %}
+
+    {% set _index = 1 %}
     <thead class="{{ _head_css_class }}">
       {% for header in _headers %}
         <tr class="{{ _row_css_class }}">
           {% for cell in header %}
-            {% set _heading_attribute = '' %}
+            {% set _heading_attribute = 'id="' ~ _id  ~ '-' ~ _index ~ '"' %}
```
planctus: confusing, is this intentional that we first set the _heading_attribute as the id and then if rowspan is set we delete it and recreate it?

src/implementations/twig/components/table/table.html.twig
```diff
@@ -115,6 +120,9 @@
             {% elseif cell.label is not empty and _sortable %}
               {% set _heading_attribute = _heading_attribute ~ ' data-ecl-table-sort-toggle' %}
             {% endif %}
+            {% if cell.headers is not empty %}
+              {% set _heading_attribute = _heading_attribute ~ 'headers="' ~ cell.headers|e('html_attr') ~ '"' %}
```
planctus: it seems it's correctly rendered, but for "readability" of the template i would add a white space between attributes.

src/implementations/twig/components/table/table.html.twig
emeryro: this was a mistake

PR #3253, opened 2024-02-22
feat(table): add optional caption - FRONT-4274

emeryro:
- add table caption to the demo. A twig parameter `caption` is provided to add it easily
- style caption


src/implementations/vanilla/components/table/table.scss
```diff
@@ -73,6 +73,14 @@ $table: null !default;
   }
 }
 
+.ecl-table__caption {
+  caption-side: bottom;
```
planctus, line 77: considering that most likely we will receive requests to make this position parametrical, why don't we add this option already?

emeryro: The position has been decided by the designers.
So, sure, we could make it as a parameter, but this may be just an option never used. I'd rather wait to have a use case

planctus (approved):

PR #3290, opened 2024-03-18
fix(blockquote): fix blockquote in wysiwyg - FRONT-4307

emeryro:
- fix display of the blockquote in ECL4; it was wrongly displayed horizontally and the font was not correct

### DIFF
--- a/src/presets/eu/src/eu-default.scss
+++ b/src/presets/eu/src/eu-default.scss
@@ -169,6 +169,11 @@
 .ecl blockquote:not([class*='ecl-'], [class*='wt-']),
 .ecl blockquote:is([class*='ecl-u-']) {
   @extend %ecl-blockquote;
+}
+
+.ecl blockquote p:not([class*='ecl-'], [class*='wt-']),
+.ecl blockquote p:is([class*='ecl-u-']) {
+  @extend %ecl-blockquote__citation;
 
   &::before,
   &::after {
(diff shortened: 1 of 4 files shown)


planctus (approved):

src/presets/eu/src/eu-default.scss
```diff
@@ -169,6 +169,11 @@
 .ecl blockquote:not([class*='ecl-'], [class*='wt-']),
 .ecl blockquote:is([class*='ecl-u-']) {
   @extend %ecl-blockquote;
+}
+
```
planctus, line 173: Why aren't we handling the margin top here in EU as we do in EC? I think it should be the same, it would otherwise result as an inconsistent behavior in the two themes.

emeryro: true, I just skipped it

planctus (commented):
I noticed that in EU there is no handling of the margin of the paragraphs, better to apply the same behaviour in the two themes, imho..

src/presets/eu/src/eu-default.scss
```diff
@@ -175,6 +175,12 @@
 .ecl blockquote p:is([class*='ecl-u-']) {
   @extend %ecl-blockquote__citation;
 
+  margin-top: map.get(var.$blockquote, 'author', 'margin');
```
planctus: juest seeing that this is 1.5rem in EU, isn't it too much? maybe it would be better to introduce a new variable or just use something like 0.5rem in both themes, no..?

emeryro: If I'm not wrong, it is already var(--s-m), 1rem in both themes.
Sure, we could set a fixed value, but I guess that 1rem is fine here

PR #3291, opened 2024-03-18
feat(button): remove tertiary button background and update components using it - FRONT-4295

emeryro:
- remove white background on tertiary button (EC and EU)
- site header: use default tertiary button display (login, language switcher, search and menu)
- carousel: remove button previous/next (so no more tertiary button here)
- gallery: use tertiary button, but keep a white background by default. Also add an extra border on focus


planctus (approved):
All good to me, but the hamburger in EC is now barely visible..

PR #3294, opened 2024-03-19
fix(featured-item): fix picture spacing - FRONT-4310

emeryro:
- fix spacing around the picture on EU, for highlighted variant
- fix mobile spacing
- fix border radius for EU


planctus:
In mobile we have this: 
image
We used to have only a padding bottom in mobile for the highlighted, now we get also a padding left

emeryro:
In fact, the mobile spacing was wrong already, with some right/left spacing remaining.
It should be fine now

PR #3300, opened 2024-03-22
fix(carousel): fix keyboard navigation - FRONT-4316

emeryro:
- allow user to go to the slide content by pressing "tab"
- improve accessibility markup with role "tablist", "tabpanel"  and "tab"
- make focus visible on the buttons


planctus:
mmh..i find this very confusing and counter intuitive, so the user typically reaches the element using tab, it will start navigating in the "controls" of the carousel but eventually, once reaching the slides indicators it will see his focus move into the slide, tabbing again (if nothing else is focusable) he will not be able to reach the buttons but he we'll start focusing outside of the carousel.
Furthermore i don't understand why the buttons, when they get focused, they lose their styles as "current", so that only when you move the focus out of them you get the style back.

In my opinion, maybe arrow up could have been used for introducing the possibility of reaching the slide content, using tab breaks the most standard keyboard navigation use case.

emeryro:
Regine confirmed in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/ECL-891 that the behavior is as expected now

PR #3302, opened 2024-03-22
feat(button): update EU button styles - FRONT-4323

emeryro:
- icon size is now `xs` by default (same as EC), and changes to `m` when there is no label (handled via css)
- different display for active state
- background, font color and border color updated


src/themes/ec/variables/_button.scss
```diff
@@ -3,11 +3,13 @@
 
 $button: (
   radius: 2px,
+  font: var(--f-m),
   shadow: none,
   border-width: 2px,
   outline-width: 1px,
   padding-horizontal: var(--s-m),
   padding-vertical: var(--s-xs),
+  icon-only-size: map.get($icon, 'xs'),
```
planctus: i wonder why we are notsimply doing the same in EC and EU, we aligned now the default icon size but we introduced this difference in the icon only use case..

src/implementations/twig/components/button/button.story.js
```diff
@@ -146,7 +146,7 @@ const prepareData = (data, args) => {
   if (args.icon_name && args.icon_name !== 'none') {
     data.icon = {};
     data.icon.name = args.icon_name;
-    data.icon.size = system === 'eu' ? 's' : 'xs';
+    data.icon.size = 'xs';
```
planctus, line 150: Can we make the hide_label control conditional? It should appear only when an icon has been chosen

src/themes/ec/variables/_button.scss
emeryro: I'll check with the designers

src/implementations/twig/components/button/button.story.js
emeryro: Well, can we do that? :)
As far as I understood, the conditional control (with 'if') only work with boolean, I don't know if we can detect if a select list has a specific value

emeryro: I updated the story to set a default icon if the hide_label control is true.
That's not perfect, but already better

planctus: but we have already in the button story file the same condition used for the icon args:
`if: { arg: 'icon_name', neq: 'none' },`

emeryro: that's true... I'll update

planctus (approved):

PR #3304, opened 2024-03-22
feat(modal): add screen reader text to icon - FRONT-4320

emeryro:
- add a `title` to the modal icon (visible when using a variant). A twig paramter `sr_icon` is provided to add it direclty.


planctus (commented):
Can we add this also in the tests for the modal and the notification..?

emeryro:
I added the test for the modal. For the notification it was already part of the snapshots

planctus (approved):

PR #3305, opened 2024-03-25
feat(file-upload): add aria-live - FRONT-4318

emeryro:
- add `aria-live` to the file list


src/implementations/vanilla/components/file-upload/file-upload.js
```diff
@@ -128,6 +128,7 @@ export class FileUpload {
 
     // Update file list
     this.fileUploadList.innerHTML = fileList;
+    this.fileUploadList.setAttribute('aria-live', 'polite');
```
planctus: just to be sure we are really doing somethign beneficial for the accessibility, i went reading the docs and i see that: 

> The aria-live attribute is set on an empty element. When an update to the page occurs, the empty element with that aria-live attribute should be updated with a brief announcement informing the user an update has been made.

So i wonder, shouldn't this be set initially, before we populate the list, so that when the list is populated the new content will be announced..? If the request was explicitely expecting this when the content it's added, forget about this

emeryro: I don't fully know how the aria-live attribute works.. I'll ask Regine

planctus (approved):

PR #3309, opened 2024-03-27
feat(radio): improve accessibility for required radio group - FRONT-4329

emeryro:
- add role radiogroup
- add aria-required


src/implementations/twig/components/form-group/__snapshots__/form-group.test.js.snap
```diff
@@ -1859,7 +1859,6 @@ exports[`Form group  with Radio renders correctly when disabled 1`] = `
         class="ecl-radio__input"
         id="radio-default-1"
         name="radio-group-1"
-        required=""
```
planctus: Did we really receive this request..? But like this the form validation would not work, basically we are informing the screen readers which is fine but we are not really making the items required anymore, this way

emeryro: We did recieve this request... I'll double check with Regine

emeryro: "required" has been brought back

planctus (approved):

PR #3312, opened 2024-04-02
feat(banner): add xs banner - FRONT-4321

emeryro:
- add extra small banner (ratio 6/1)


planctus (commented):
While reviewing this i noticed that we were not really handling the plain-background variant regarding the dynamic height, only 4/1 aspect ratio was considered in that case. If you check in this pull request indeed it breaks, not increasing the height for the smaller variants where the content is taller than the resulting height coming from the aspect ratio.
I created this pr, you can evaluate it so that you can also integrate in this: https://github.com/ec-europa/europa-component-library/pull/3313 but in any case the demo is not so "clear", we increased the content size for testing the dynamic height, but like this setting the size has not a direct, recognizable effect.

PR #3322, opened 2024-04-03
feat(multi-select): fix arrow down navigation - FRONT-4335

emeryro:
- prevent the dropdown from closing when pressing arrow down on the last item

### DIFF
--- a/src/implementations/vanilla/components/select/select.js
+++ b/src/implementations/vanilla/components/select/select.js
@@ -856,9 +856,10 @@ export class Select {
    * Private method to handle the focus switch.
    *
    * @param {upOrDown}
+   * @param {loop}
    * @private
    */
-  #moveFocus(upOrDown) {
+  #moveFocus(upOrDown, loop = true) {
     const activeEl = document.activeElement;
     const hasGroups = activeEl.parentElement.parentElement.classList.contains(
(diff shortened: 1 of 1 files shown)


planctus:
This is something i don't understand, i think that the issue is that using arrow down stops working once your focus is moved back to the input, not that it behaves like other keyboard keys when you reach the last option.
Now we are differentiating the user experience, instead, and it's not clear the reason why, it would seem natural to just be able to keep using arrow down also to move from the input to the dropdown. instead..

emeryro:
This PR is only about the flow when reaching the end of the list.
According to Regine, it is best to not loop when using the arrows:
> Once the user reached the last item of the submenu, it should go back to the first item or (even better) do nothing (like the single select), but not close the submenu.

It is not unusual to have a different flow between arrows and tab.

Opening the input by using arrow down is already done in [this PR](https://github.com/ec-europa/europa-component-library/pull/3320) (this was just not updated here)

planctus:
To me it's just an overhead, i don't understand how something that goes against the logic, breaks consistency, forces the user to change the key used to navigate with the keyboard can be beneficial for the accessibility.
Also the code, to support this, shows this inconsistency, but Regine said so, ok...

PR #3324, opened 2024-04-03
feat(media-container): update caption style - FRONT-4324

emeryro:


src/themes/ec/variables/_media-container.scss
```diff
@@ -5,8 +5,11 @@ $media-container: (
   border-radius: 0,
   shadow: none,
   caption-background-color: transparent,
-  caption-color: var(--c-d),
-  caption-font: var(--f-s),
+  caption-color: var(--c-d-80),
+  caption-font: normal normal map.get($font-weight, 'regular') #{map.get(
```
planctus: wouldn't it be easier to use var(--f-s) and create another prop to set the right line height..?

PR #3333, opened 2024-04-08
feat(gallery): fix aria-label and alt text for video - FRONT-4364

emeryro:
- remove aria-label on the gallery items link (it was preventing the other information to be read)
- add textual information when an item is a video (added in the end of the image alt). Twig parameter `sr_video_label` added to handle this


src/implementations/twig/components/gallery/gallery-item.html.twig
```diff
@@ -10,6 +10,7 @@
     - "embedded_video" (optional) (object) (default: {})
     - "share_path" (optional) (string) (default: '')
   - "icon_path": (string) (default '')
+  - "sr_video_label" (string) (default: ''): additional label for the video items; for screen readers
```
planctus, line 13: String game..
so now we need two string to be provided separately and then we go combining them, it's already very optimistic to think that those strings will be provided, but the behavior of appending one of them to the other might result simply in the user being confused by what is happening.
The amount of doubts i have regarding these series of "accessibility improvements" is growing day by day as the list of related task created about them.

emeryro: I know... And I'm quite sure that most of these strings would not be filled correctly (if at all)
But I don't really see other ways that adding new twig paramters every time and hoping for the best

emeryro: The hope I see here is that this new "Video" label is not something meant to be filled by an editor. It is most likely something that would be added automatically by the implementation.
So worse case, if the user does not fill the alt it will still read "Video" (in the correct language)

PR #3337, opened 2024-04-10
feat(colors): update EU colors - FRONT-4326

emeryro:
- add new tint for semantic colors
- deprecate a few colors:
```
--ecl-color-primary-130
--ecl-color-secondary-10
--ecl-color-dark-140
--ecl-color-dark-120
--ecl-color-error-80
--ecl-color-branding
--ecl-color-visited
--ecl-color-visited-100
--ecl-color-visited-40
```


src/themes/eu/_custom-properties.scss
```diff
@@ -178,4 +174,24 @@
...
+  --ecl-color-secondary-10: #{map.get($color, 'secondary-10')};
+  --c-s-10: var(--ecl-color-secondary-10);
+  --ecl-color-dark-140: #{map.get($color, 'dark-140')};
+  --c-d-140: var(--ecl-color-dark-140);
+  --ecl-color-dark-120: #{map.get($color, 'dark-120')};
+  --c-d-120: var(--ecl-color-dark-120);
+  --ecl-color-error-80: #{map.get($color, 'error-80')};
+  --c-er-80: var(--ecl-color-error-80);
+  --ecl-color-branding: #{map.get($color, 'branding')};
+  --c-br: var(--ecl-color-branding);
```
planctus, line 190: I understand keeping the custom properties and deprecate them, but do we have to do it also for aliases? There is one usage in the banner for the --c-br one, but it's "hardcoded! in the common scss, not sure whehter what is being deprecated in EU will also be deprecated in EC..

emeryro: The EC color are not meant to be modified, as far as I know. 
I didn't check where the deprecated EU colors may be used, but as I understood, design is doing some updates on all the EU components (at least for fonts and spacing). So I assume that the deprecated colors would be not be used after that.

planctus (approved):

PR #3351, opened 2024-04-17
feat(layout): update max width - FRONT-4381

emeryro:
- update global max width to 80ch


planctus (commented):
Question: we noticed that in some cases we are setting the max-width on a container with multiple elements, like a title, a description, as a "group" but this has consequences, the result is not the same as it would if the max-width would be set on each element individually. Shoouldn't we take care of this in the context of this pr..?

emeryro:
This is a quick first step here. DG COMM is investigating for each component what should use the max width or not.
This will be done in a future task

PR #3352, opened 2024-04-18
fix(menu): fix resize on mobile - FRONT-4382

emeryro:
- fix the issue on mobile where the scroll jumped back to top when scrolling (due to the resize event)

### DIFF
--- a/src/implementations/vanilla/components/menu/menu.js
+++ b/src/implementations/vanilla/components/menu/menu.js
@@ -144,6 +144,7 @@ export class Menu {
     this.currentItem = null;
     this.totalItemsWidth = 0;
     this.breakpointL = 996;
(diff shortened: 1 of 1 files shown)


emeryro:
@planctus I don't know exactly why we are scrolling back to top when doing the menu resize, but if that's still needed, we should probably apply the kind of hack of the PR

planctus:
mmh..forcing the page scroll to top when resizing is needed to ensure that the re-positioning of the menu will work, not that i recall exactly the use case, but the calculation would most likely fail if the header is not visible, for instance.
This fix seems harmless regarding this, preventing only the unexpected trigger of the resize event when just scrolling, so i would assume it is ok

planctus (approved):

PR #3353, opened 2024-04-18
feat(media container): disable Youtube shortcuts - FRONT-4363

emeryro:
- disable keyboard shortcuts for Youtube player (it is interfering with the screen reader shortcuts)
note: this is only for Youtube videos currently, as each player may have its own way of handling shortcuts, and it may or may not be an issue


planctus (commented):
Mmhh..but, this works with youtube videos only, i don't really think that adding this in the demo would be understood in its relevance concerning the accessibility.
It feels that we should do this via the js more than expecting users to append this string to the url and if that exists, it would also be nice to be able to pass the right "option" to the right player.
At the same time the first link that i followed, leading to the a discussion in the drupal community, is this: https://www.drupal.org/project/video_embed_field/issues/3228051 the decision was not to append the option to disable the keyboard shortcuts..

emeryro:
I'll double check this

emeryro:
As it seems, this is still require to disable these shortcuts (that's a shame...)

It has been done for Youtube, as it was the initial request. If needed for other players, that would require more investigation

PR #3354, opened 2024-04-18
feat(gallery): add item title for share link - FRONT-4369

emeryro:
- add unique id for the gallery. This is done via a new twig parameter `id`. If empty, it is randomly generated.
- add item title for each gallery item. This is done via a new twig parameter `title`
- add `aria-describedby` to the share and download links, using the item title


src/implementations/twig/components/gallery/gallery-item.html.twig
```diff
@@ -19,6 +22,7 @@
 
 {# variables #}
 
+{% set _id = id|default('') %}
```
planctus, line 25: but we are not generating any id as we say in the doc comment. I would use random() also here, then we will change them all at once.

src/implementations/twig/components/gallery/gallery-item.html.twig
```diff
@@ -112,6 +117,11 @@
         class="ecl-gallery__description"
         data-ecl-gallery-description
       >
+      {%- if _item.title is defined and _item.title is not empty -%}
```
planctus, line 120: once we will ensure we have an id this will be ok.

src/implementations/vanilla/components/gallery/gallery.js
```diff
@@ -553,6 +556,7 @@ export class Gallery {
     );
     if (shareHref != null) {
       this.overlayShare.href = shareHref;
+      this.overlayShare.setAttribute('aria-describedby', `${id}-title`);
```
planctus: we can assume that an id will be always set when using the template we provide, but before using it i would still check if its not false

planctus (commented):
Just a couple of checks to be added in the js..

PR #3358, opened 2024-04-23
fix(inpage): Renaming selector for active element - FRONT-4386

planctus:
This turn out to be a kind of conflict between the selector added to the active "heading element" and the default css which excludes h2s with a selector starting with ecl. So the fix is only renaming this selector to prevent this exclusion.


emeryro (commented):
I'm a little worried by this change. If users are relying on the css class for something on their side, it would be a (limited) breaking change. Also I see that the spyClass is still using 'ecl-inpage-navigation__item--active'. Is it on purpose?

The only other solution I see would be to add an exception in our css. That's quite dirty also, though...
image

planctus:
I don't know, it doesn't really feel like anything critical, we do not style that selector at all, we can mark the pr with "markup changes" in case, if we really worry about any usage of this selector, it seems cleaner doing this than introducing an exception for this selector in the default css.
For the other selector, that is not problematic and we have some styles associated to.

emeryro (approved):

PR #3369, opened 2024-04-30
feat(icon): add screen reader label for external icons - FRONT-4373

emeryro:
- new twig parameter `sr_external` in the the following components, to add a label on the external icon: link, menu, mega menu
- update demo data of several components to use this parameter


src/implementations/twig/components/link/link.html.twig
```diff
@@ -76,7 +78,15 @@
     size: '2xs',
   })] %}
 
-  {% set _link = _link|merge({ icon_position: _link.icon_position|default('after') }) %}
+  {% set extra_accessibility = {
+    title: _link.sr_external,
```
planctus: mmh..we could be rendering an empty "title" attribute this way, shouldn't we also check for the non emptyness of this parameter before setting the extra_accessibility oblject?
But, furthermore, this way we would be introducing a breaking change, if in an existing implementation someone was already passing this object it would be overridden by this.
And this makes me think..do we need this change at all..? It was already possible to pass the needed data to the link component, indeed, but ok, this surely simplifies that, but we would at least need to make this backward compatible..

src/implementations/twig/components/link/link.html.twig
```diff
@@ -76,7 +78,17 @@
     size: '2xs',
   })] %}
 
-  {% set _link = _link|merge({ icon_position: _link.icon_position|default('after') }) %}
+  {% if _link.sr_external is not empty or extra_accessibility.title is not empty %}
```
planctus: mmh..it seems more logical to me to do this:
```
  {% if _extra_accessibility.title is empty and _link.sr_external is not empty %}
    {% set _extra_accessibility = _extra_accessibility|merge({ title: _link.sr_external }) %}
  {% endif %}
```
Basically we would use the new parameter only in case this is not passed the "old way". Currently with the code  we have we could still generate an empty title tag in the icon, with this code we wouldn't.
Furthermore we are merging the object so whatever is passed in it other than the title would be preserved.
If you follow this suggestion _extra_accessibility needs to be defined as usual with a merge using the default parameter and then there is no need to set again the default when passing it in the include.

src/implementations/twig/components/link/link.html.twig
```diff
@@ -171,7 +182,7 @@
         {%- include '@ecl/icon/icon.html.twig' with _icon|merge({
           icon: icon|default({}),
           extra_classes: _icon_extra_classes,
-          extra_accessibility: extra_accessibility|default({}),
+          extra_accessibility: _extra_accessibility|default({}),
```
planctus: we don't need anymore the default here, since we already do that at the top of the template

planctus (commented):
Just a last tweak for the template, no need to set a default for the extra_accessibility in the include anymore

planctus (approved):

PR #3372, opened 2024-05-02
feat(menu): add aria-current - FRONT-4405

emeryro:
- add aria-current=page to identify the current page in the menu
- add aria-current=true to identify selected menu item


planctus (changes requested):
Mmhh..if we are about to do so, we need to demo only one current item, otherwise we would be setting the current page multiple times on different links which would not be ok, but we should also update the docs related to this parameter, in the mega-menu i used this: - "is_current" (boolean) Whether the item links to the current page

emeryro:
The markup has been updated to use `aria-current="true"` for first level items. That would be read as "this is the selected item in the list". So we would only have `aria-current="page"` on the final level link.
The same logic has been applied to the mega menu

planctus (approved):

PR #3378, opened 2024-05-06
feat(typography): update EC and EU text styles - FRONT-4392

emeryro:
- font size and line height updated for EC and EU
- new `ui` font added (s and m)
- font `prolonged` are now deprecated


planctus (commented):
Shouldn't we update the website pages about typography as well..?

PR #3380, opened 2024-05-06
feat(featured-item): add simple variant - FRONT-4413

emeryro:
- add new variant `simple` for featured item

### DIFF
--- a/src/specs/components/featured-item/demo/data--simple.js
+++ b/src/specs/components/featured-item/demo/data--simple.js
@@ -6,7 +6,7 @@ module.exports = {
   title:
     'Title: Ut enim ad minim veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo',
   description:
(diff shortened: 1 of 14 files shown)


planctus (changes requested):
It seems broken without the image:
image

src/specs/components/featured-item/demo/data--simple.js
```diff
@@ -0,0 +1,54 @@
+const publicUrl = process.env.PUBLIC_URL || '';
+const exampleLink = `${publicUrl}/example`;
+
+module.exports = {
+  type: 'simple',
+  title:
+    'Title: Ut enim ad minim veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo',
+  description:
+    '<p class="ecl-u-type-paragraph-m">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.</p>',
```
planctus: Why do we demo this markup..? Is this a leftover maybe from the past? Or maybe this is to demonstrate that the <p> is not adding any margin when used here..? Like this a user would think that this utility is part of the markup of the component, i would remove it als from the other "variants".

planctus (approved):

PR #3386, opened 2024-05-08
fix(mega menu): fix accessibility issues - FRONT-4424

emeryro:
Fix most important accessibility issues, before component update
- remove `aria-hidden`, preventing screen reader to access the menu
- add `aria-expanded` on the link elements
- remove unneeded `aria-label`; Twig parameter `trigger_aria_label` is no longer used


src/implementations/vanilla/components/mega-menu/mega-menu.js
```diff
@@ -982,6 +987,8 @@ export class MegaMenu {
         'ecl-mega-menu__subitem--expanded',
         'ecl-mega-menu__subitem--current',
       );
+      const itemLink = queryOne(this.subLinkSelector, level2);
```
planctus, line 990: confusing, we were and we are still setting the aria-expanded on the list items as well, if we set it on the inner link and if that makes sense then i guess we should remove it from the list item

emeryro: that's true, but lots of code is depending on the aria-expanded at list item level (js and css).
I didn't want to do major changes now as the mega menu will have to be updated soon anyways

src/specs/components/mega-menu/demo/data.js
```diff
@@ -32,15 +32,13 @@ module.exports = {
     {
       label: 'News and media',
       path: exampleLink,
-      trigger_aria_label: 'Access news and media children',
```
planctus, line 35: this parameter was added in order to be able to provide the screen reader users the information about the fact that the "trigger" is only a button in the UI, clicking on it is not moving the user anywhere while the first item accessed after the "click" is again a link with the same text and this time this is really a link.
Might not have been the best way to make this distinction clear, but it seems to me that removing it will create a pretty confusing experience for the user. (this thing will change, by the way, not sure it would be relevant also in the updated version of the menu)

planctus (commented):
Left a couple of comments, see if they are relevant or not, there are a few conflicts to be solved ;)

src/specs/components/mega-menu/demo/data.js
emeryro: ideally the first level items should be buttons instead of links, as they only open the dropdown. This would remove the confusion.
anyways, let's see that in the "updated" version

PR #3387, opened 2024-05-14
feat(footer): update eu footer links - FRONT-4394

emeryro:
- update links to have the full label clickable
- update font size for footer titles


planctus (commented):
> update links to have the full line clickable

Just a question, was this foreseen and then the request has changed? The links are inline so only their label is clickable..

emeryro:
this was wrongly written. It is not the full line clickable, but the full label (there were some standard text before)

planctus (approved):

PR #3389, opened 2024-05-14
feat(lists): Styles - FRONT-4417

planctus:


src/implementations/vanilla/components/ordered-list/ordered-list.scss
```diff
@@ -40,6 +40,10 @@ $list: null !default;
   &:last-of-type {
     margin-bottom: var(--s-xs);
   }
+
+  .ecl-link {
+    text-decoration: none;
```
emeryro, line 45: shouldn't we use standalone links here instead of forcing the display? There may be links inside paragraphs that would still need to be underlined

emeryro (commented):
Text is very close to the "bullet" here"
image

Also spacing is quite different between ordered and unordered list, not sure if this was intentional
image

src/implementations/vanilla/components/ordered-list/ordered-list.scss
planctus: ehm..we do not include the link component here, we expect markup to be passed to the template, instead, and this would mean that the users would have to add the ecl-link--standalone every time they want to have a link.
I do not understand the reason for this change, but i'd say that if the expected look and feel with links is without the underline better doing this than introducing a breaking change.

src/themes/eu/variables/_description-list.scss
```diff
@@ -3,9 +3,9 @@
 
 $description-list: (
   term-color: var(--c-d-140),
-  term-font: var(--f-l),
-  term-margin-bottom: var(--s-xs),
-  term-margin-top: var(--s-m),
+  term-font: var(--f-m),
+  term-margin-bottom: var(--s-2xs),
+  term-margin-top: var(--s-l),
   definition-font: var(--f-p-m),
```
emeryro:
```suggestion
  definition-font: var(--f-m),
```
font prolonged is deprecated on EU

emeryro:
On EU, the bullets are going a little outside the container
image

This is also the case on EC when there are more than 9 items
image

I don't think that this is real issue, but we may receive remarks about that

planctus:
true, although i haven't investigated the reason why this difference in EC and EU, is it just the marker used, i wonder..? Anyway, it seemed relevant to align the "initial" content of the lists with the content in the grid, so i added a minimal padding left in EU.

emeryro (approved):

PR #3390, opened 2024-05-15
feat(gallery): design update - FRONT-4387

emeryro:
- display title instead of description in the teaser
- reorder elements in the overlay (previous/next buttons)
- display title in the overlay
- remove truncation of the description, but display it full width


planctus:
something broke, the same happened in the slider, guess it's about buttons: 
image

emeryro:
@planctus indeed that seem to be an issue introduced when we worked on the single icons set.
I set the display to inline-flex, it should be fine now.

planctus (approved):

PR #3393, opened 2024-05-16
feat(list with illustration): add center for vertical list (icons) and improve print - FRONT-4428

emeryro:
- add center display for vertical list with icons
- apply latest changes to print display


src/implementations/twig/components/list-illustration/list-illustration.story.js
```diff
@@ -66,8 +66,11 @@ const getArgs = (data, variant) => {
 
   if (variant.includes('horizontal')) {
     args.column = 2;
+  }
+  if (variant !== 'vertical-image') {
```
planctus, line 66: couldn't we use conditional controls instead of wrapping them in a condition? I see that you used "includes" in some of these condition and this is not something we can do with the conditional controls directly, but probably with multiple conditions we can. Just for consistency since we are using this approach from the moment we discovered the conditional controls..

emeryro: I did a test, and this does not seem to work like that.
image
According to the documentation, the `if` can be done only on other control args, and here we are testing another variable
image

planctus: out of curiosity i have tried to do this and we can add a control without showing it, so variant could become a control and we would be able to use its value as the condition to show the control but multiple conditions do not seem to work, anyhow, so in this case no use for conditional controls.

PR #3402, opened 2024-05-27
feat(media-container, gallery): embedded video audio description information - FRONT-4362

emeryro:
- add possibility to add information related to the embedded video audio description. New twig parameter `sr_video_audio` for this in the gallery items and media container
- example of values for this field provided in the demo, this could be used as standardized value
	- "In the video below, there is no audiodescription available yet"
	- "In the video below, there is no audiodescription available because all the content is in the captions and default audio track"
- add title to html video, using existing parameters


src/implementations/twig/components/gallery/gallery-item.html.twig
```diff
@@ -77,6 +78,9 @@
   {%- if _media_href is not empty and _item.embedded_video is not empty -%}
     data-ecl-gallery-item-embed-src="{{ _media_href }}"
   {%- endif -%}
+  {%- if _item.sr_video_audio is not empty -%}
+    data-ecl-gallery-item-embed-audio="{{ _item.sr_video_audio|e }}"
```
planctus: mmh..this seems a bit critical, not sure what is the intention here, it seems the right thing to do to escape a value considering that we are in the context of an html attribute but:

- in our demo this doesn't change anything, which is weird
- we cannot exclude double escaping, basically, in our env the autoescape is disabled but in the ewpp env is enabled instead

Then the js retrieves this value and uses that setting it as the content of an element. This value should theoretically contain html entities, at this point, if the string contains a link.
In order to convert back this value to the original html it seems that we would first need to set the value as the innerHTML of an element and then retrieve it again, for finally setting it in the final element, but this is untested on my side.

So, not sure, really, i think it's probably better to let the application define the string as safe, as we normally do basically, otherwise if we escape it ourselves then we should probably render it with |raw to prevent double escaping and the js should convert this value back to html.

Otherwise we could build the link internally, but this would need probably more data, like an Url and separately a string..

emeryro: good news here, according to Regine it is no longer needed to put a link to the audio description file, as it would be integrated in the video itself.
So I removed the escaping, it should be fine now.

PR #3407, opened 2024-05-28
feat(table): table without header - FRONT-4437

emeryro:
- add a border top to the table when there is no header
- add a control in Storybook to preview it


src/implementations/twig/components/table/table.story.js
```diff
@@ -28,7 +40,15 @@ const getArgTypes = () => ({
   },
 });
 
-const prepareData = (data, args) => Object.assign(data, args);
+const prepareData = (data, args) => {
```
planctus, line 44: this way we break the simple control, we should still return an Object.assign to get it working again.

planctus (changes requested):
Small issue in the story introduced by this pr

planctus:
right, but isn't it better actually to completely hide the "simple" control after setting in it to true when the user chooses to hide the headers ? What happens currently is that you choose to remove the headers but then you still see them in mobile because of the attribute set, which appears as "magical".. :)

planctus (approved):

PR #3420, opened 2024-06-04
feat(range): add bubble for value - FRONT-4436

emeryro:
- display current value in a bubble

Note: mouse hover is handled with javascript, to avoid an issue where the bubble is also displayed when hovering the form label
image


planctus:
But why do also show the old counter..? Questionable per se the need for this new display, guess that we are copy pasting design ideas from material, basically, but like this it really feels is redundant, no..?

emeryro:
That's not really redondant: the bubble is only displayed when the range is hover or focus, while the "static" value display is always there

planctus:
if you say so..that is the definition of a duplication, i'd say, the info is already present and then we show it a second time, Anyway, fine, surely not the biggest concern i have regarding all that we are doing.

planctus (approved):

PR #3421, opened 2024-06-05
feat(form): update style of all form elements - FRONT-4432

emeryro:
- update font and spacing for several form elements
- reduce differences between EC and EU


src/implementations/twig/components/form-group/form-group.story.js
```diff
@@ -19,22 +18,20 @@ import dataFileUploadMultiple from '@ecl/specs-component-file-upload/demo/data--
...
 };
-dataText.invalid_icon.size = system === 'ec' ? 's' : 'm';
-dataSingle.input.icon_size = system === 'ec' ? 'xs' : 's';
 
 const getArgs = (data) => {
   const args = {
     show_label: true,
     show_helper: true,
     show_error: true,
+    width: data.input.width || 'm',
```
planctus: with this we get the width control everywhere, even where it is not relevant, like in the checkbox

planctus (approved):

PR #3430, opened 2024-06-12
feat(accordion): update styles - FRONT-4455

emeryro:
- update font and spacing of accordion


planctus:
Just a doubt, it has always been like this, so maybe there is no reason why questioning this now but it seems we would benefit from an alignment to the center of the "toggle" icon, what do you think?
image

emeryro:
In other cases where there is text on multiple line and icons, the designers prefers to have the icon stay on top. So I guess that this is the same here.

PR #3434, opened 2024-06-13
feat(gallery): add grid display - FRONT-4459

emeryro:
- alternative display using css grid. Twig parameter: `grid`
- when in grid display, added possibility to change number of column and aspect ratio. Twig parameters `column` and `ratio`


planctus:
The columns control doesn't seem to do anything, i see the reason why, the css is missing an & for the columns selectors
```
@include breakpoints.up('m') {
  .ecl-gallery--grid {
    .ecl-gallery__list {
      grid-template-columns: repeat(3, 1fr);
    }

    .ecl-gallery--col-2 {
      .ecl-gallery__list {
        grid-template-columns: repeat(2, 1fr);
      }
    }

    .ecl-gallery--col-4 {
      .ecl-gallery__list {
        grid-template-columns: repeat(4, 1fr);
      }
    }
  }
}
```

src/implementations/twig/components/gallery/gallery.html.twig
```diff
@@ -66,6 +72,18 @@
   {% set _css_class = _css_class ~ ' ' ~ extra_classes %}
 {% endif %}
 
+{% if _grid %}
+  {% set _css_class = _css_class ~ ' ecl-gallery--grid' %}
+
+  {% if _column is not empty %}
```
planctus, line 78: we set the default to 3 for the column..

src/implementations/twig/components/gallery/gallery.html.twig
```diff
@@ -66,6 +72,18 @@
...
+    {% set _css_class = _css_class ~ ' ecl-gallery--col-' ~ _column %}
+  {% endif %}
+
+  {% if _ratio is not empty %}
```
planctus, line 82: same here for the ratio, it cannot be empty, basically..

planctus (approved):

PR #3436, opened 2024-06-17
feat(select): Making the select all and search field optional in the multiselect - FRONT-4462

planctus:
In this pr three tickets are tackled:

- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-4472
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-4462
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-4470

For what concerns the usage of arrow down, the attempt here is to mitigate the effects of this pr: https://github.com/ec-europa/europa-component-library/pull/3322 where the navigation with arrowDown was stopped on the last checkbox to prevent the closure of the dropdown.

Now, there is no need to prevent the user from accessing the toolbar using arrowDown, the closure of the dropdown was due to a wrong operation performed by us, so now what happens is that if the toolbar is in place arrowDown will continue to work after the last checkbox, if it's not shown, instead, to take into account the Regine's recommendation, arrowDown stops working on the last option not restarting the loop, although also this looks unreasonable to me, i'm not surprised we received this "request" considering the concerns i expressed in the aforementioned pr.


emeryro (approved):

emeryro (changes requested):
On my side I am still unable to access the buttons with arrow down; it stops at the last checkbox (on Chrome and Firefox)
Going from the button to the checkbox works fine

planctus:
It seems i forgot to push the last commit, did it now, most likely it's where i applied the changes to the arrow down navigation..

emeryro (approved):

PR #3445, opened 2024-06-20
feat(menu): improve focus loop - FRONT-4402

emeryro:
- Fix focus trap in the default menu
- Set focus on first item when opening the menu and mega menu


src/implementations/vanilla/components/menu/menu.js
```diff
@@ -1168,6 +1163,16 @@ export class Menu {
       }
     });
     this.checkMegaMenu(menuItem);
+
+    // Focus first item
+    console.log(menuItem);
```
planctus: forgotten debug here

planctus (commented):
Seems all good to me, just the console.log to be removed

PR #3447, opened 2024-06-21
feat(news-ticker): Icon or image for each item, iteration on styles for EC and EU - FRONT-4463

planctus:


src/implementations/twig/components/news-ticker/news-ticker.html.twig
```diff
@@ -71,21 +72,28 @@
...
     <div class="ecl-news-ticker__content">
       <ul class="ecl-news-ticker__slides" id="{{ _aria_controls }}">
       {% for _item in _items %}
+        {% set _icon = '' %}
+        {% if _item.icon is not empty and _item.icon.name is not empty %}
+          {% set _icon %}
+            {% include '@ecl/icon/icon.html.twig' with {
+              icon: _item.icon|merge({
+                path: _icon_path,
+                name: _item.icon.name,
```
emeryro, line 89: can't we default the icon name to 'information' (at least until next major version)?
That way there won't be breaking changes here

src/implementations/vanilla/components/news-ticker/news-ticker.scss
```diff
@@ -27,20 +28,25 @@ $news-ticker: null !default;
 .ecl-news-ticker__icon {
   fill: var(--c-p);
   flex-shrink: 0;
-  margin-inline-start: map.get($news-ticker, 'padding-horizontal');
+  height: 1.5rem;
```
emeryro: we should use the sass maps for the icon size

src/specs/components/news-ticker/demo/data.js
```diff
@@ -12,34 +12,53 @@ module.exports = {
   sr_pause: 'Pause news ticker',
   items: [
     {
-      content: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit',
+      content:
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean et finibus mi. Sed hendrerit velit id ex ornare interdum. Duis suscipit tortor vitae mi feugiat suscipit. Morbi consequat tortor magna, ac malesuada nunc placerat non. Vivamus pulvinar augue tristique lectus lobortis iaculis.',
       link: exampleLink,
+      icon: {
+        name: 'euro',
```
emeryro, line 19: maybe we should have different icons on some slides, to showcase this possibility

emeryro (changes requested):
The focus on the link seems to be broken (only the left border is displayed)
image

src/implementations/twig/components/news-ticker/news-ticker.html.twig
planctus: mmh..i'd say no, because we would alter the behavior of the component, currently you can set or not an icon for each item, if we default it to "something" it would always be there instead, and it would be totally arbitrary to show the info icon at this point. It would not work as backward compatibility, for that to happen we should check if in any of the item there is an icon and in case there is none we could "assume" that it's "old data" and apply this fallback but it seems an overhead for the template to do so.

src/specs/components/news-ticker/demo/data.js
planctus: Done it, although for some reasons i was explicitely requested to use the euro icon..

emeryro (approved):

emeryro (approved):

PR #3449, opened 2024-06-24
feat(tabs): EU styles - FRONT-4451

planctus:


src/implementations/vanilla/components/tabs/tabs.scss
```diff
@@ -36,11 +36,16 @@ $button-width: 44px;
   margin: 0;
   padding-inline-start: 0;
   transition-duration: 0.4s;
+
+  .ecl-tabs[data-ecl-auto-initialized] & {
```
emeryro, line 40: I don't think that this rule will be applied, as there is no .ecl-tabs children of .ecl-tabs
Also targeting auto-initialized data would exclude user doing the initialization on their side. If this is to delay the css after javascript init, better add a css class from the js

planctus: i'm confused, this selector is nested in the .ecl-tabs-list selector, so this results in:
.ecl-tabs[data-ecl-auto-initialized] .ecl-tabs-list and it is correctly applied.

the data-ecl-auto-initialized attribute might sound confusing, but we do this in all of our classes in the init() method, so it's not only limited to implmentations using the autoInit.
this.element.setAttribute('data-ecl-auto-initialized', 'true');

emeryro (approved):

PR #3450, opened 2024-06-25
feat(mega-menu): New info in mobile, mobile scroll, limiting element's height - FRONT-4447

planctus:


emeryro (changes requested):
The changes are fine for the mobile display, but I noted a regression when the menu is not in the site header. Dropdown is going on top of the menu bar
image
This is not a real use case, but as it is visible, we would receive remarks

emeryro (approved):

emeryro (approved):

PR #3451, opened 2024-06-25
feat(menu): update focus on mobile - FRONT-4403

emeryro:
- when accessing a sub level, put the focus on the first element (the "back" link), to follow visual order

### DIFF
--- a/src/implementations/vanilla/components/mega-menu/mega-menu.js
+++ b/src/implementations/vanilla/components/mega-menu/mega-menu.js
@@ -1364,6 +1364,11 @@ export class MegaMenu {
           }
         }
       }
+
+      // Focus on back item
+      if (this.back) {
+        this.back.focus();
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


planctus:
At least fow what it concerns the mega menu, in this pr: https://github.com/ec-europa/europa-component-library/pull/3450 this is handled differently:
when selecting an item the focus goes "inside" of it, reaching the "discover more" if present or the first subitem.
In the second panel, the focus goes on the selected item, so that it can be "clicked" to follow the link.
I personally feel that this is preferable.

Focusing on the back button seems to force the user to always perform an additional step, it doesn't seem what it would want to get the focus on, if he is traversing the menu using the keyboard.
Not sure for the menu, same logic could be applied also there, but we are handling this in two different prs so better removing this from this pr.

emeryro:
Confirmed by Regine that what we have currently is also valid (focus on the first sub item).
Closing the PR

PR #3456, opened 2024-06-27
feat(twig): cleanup and update parameters - FRONT-4438

emeryro:
- update and cleanup some documentation (twig comment and readme)
- use randomly generated as default for some id


src/implementations/twig/components/notification/notification.html.twig
```diff
@@ -18,7 +18,7 @@
 
 {# Internal properties #}
 
-{% set _variant = variant|default('') %}
+{% set _variant = variant|default('info') %}
```
planctus, line 21: this is good, but at the sametime it might impact some implementations, who knows if someone used this component without setting any variant, indeed...
Anyhow i think that it's the right thing to do.

src/implementations/twig/components/rating-field/rating-field.html.twig
```diff
@@ -19,9 +19,9 @@
 
 {# Internal properties #}
 
-{% set _id = id|default('rating') %}
+{% set _id = id|default('') %}
```
planctus: any reason why here we do not generate a random id..?

src/specs/components/mega-menu/demo/data.js
```diff
@@ -128,7 +124,6 @@ module.exports = {
     {
       label: 'Key priorities',
       path: exampleLink,
-      link_aria_label: 'Link to Key priorities page',
```
planctus, line 130: if we don't need anymore this, and it seems indeed that the data-ecl-parent-aria-label is not used anymore in the js, then we can also remove the logic that would render it: 
```
  {% if _item.link_aria_label is not empty %}
    {% set _menu_list_item_attributes = _menu_list_item_attributes ~ ' data-ecl-parent-aria-label="' ~ _item.link_aria_label ~ '"' %}
  {% endif %}
```
mega-menu-item.html.twig line 85 as well as:

```
          {% if child.link_aria_label %}
            {% set _subitem_attrs = _subitem_attrs ~ ' data-ecl-parent-aria-label="' ~ child.link_aria_label ~ '"' %}
          {% endif %}

line 217 and we don't need to pass it to the template:
`link_aria_label: child.link_aria_label|default(''),`
line 272
```
Also the docs are still mentioning the parameter

src/specs/components/timeline/demo/data.js
```diff
@@ -8,100 +8,73 @@ module.exports = {
     from: 5,
     to: -2,
   },
-  button: {
```
planctus, line 11: I see that in the timeline template we define a _toggler variable, we set that initially to false and when we have to render the toggle we switch it to true. The question is, why..? :) It seems this is totally useless

src/specs/components/timeline/demo/data.js
```diff
@@ -8,100 +8,73 @@ module.exports = {
...
-      size: 'xs',
-      transform: 'rotate-180',
-    },
-    variant: 'secondary',
-    type: 'button',
-  },
   icon_path: '/icon.svg',
   items: [
     {
-      id: '0',
```
planctus, line 25: a bit confused... i guess that these ids where requested, indeed we still support the id for the single items, but shouldn't it behave lke the others id we set? I mean that in this pr we are always setting a default for them, here we are not and we are not demoing or testing the id anymore, i think we should set a default as in the other cases, so it will always be there and the user can choose his own ids by passing it in the data.

planctus (commented):
Very nice, a couple of different things:

- a cleanup of the mega menu item template 
- automatic generation for rating and timeline items ids

PR #3458, opened 2024-07-01
feat(select): Removing unneeded aria-label - FRONT-4477

planctus:


src/implementations/twig/components/select/select.test.js
```diff
@@ -8,8 +8,14 @@ import { axe, toHaveNoViolations } from 'jest-axe';
 import specSingle from '@ecl/specs-component-select/demo/data-single';
 import specMultiple from '@ecl/specs-component-select/demo/data-multiple';
 
-const dataSingle = specSingle.input;
-const dataMultiple = specMultiple.input;
+const dataSingle = {
+  ...specSingle.input,
+  extra_attributes: [{ name: 'aria-label', value: 'country' }],
```
emeryro, line 13: why are we putting this aria-label again here?
Also it is does not seem to be rendered

planctus: maybe because otherwise the accessibility test fails when tested in isolation, because there is no label.
This is the select test file, the aria-label is rendered and the test is passing indeed, you were probably looking at the form-group test where this  aria-label is not needed.

emeryro (approved):

PR #3460, opened 2024-07-02
feat(custom-properties): Adding custom props for font family, using them in the components - FRONT-4490

planctus:
Didn't make sense to apply this only to the link component, so all the usage of font-family in our sass are now using the custom-properties for the font-family added in this pr.


emeryro (approved):
That's fine for me. We would just have to make it clear to the requester that we are using different css properties here (and not the full font properties), so they have to override them

PR #3465, opened 2024-07-04
feat(banner): new options and video - FRONT-4488

emeryro:
- new options for the banners: 
  - horizontal and vertical alignment (`horizontal` and `vertical`)
  - 2 font sizes set (`font_size`)
  - 2 font colors (`font_color`)
  - 3 background options (`box_background`)
- title can now be a link
- banner can now display a video (`video`, `sr_pause`, `sr_play`)


src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -23,15 +49,29 @@
...
 {% set _description = description|default('') %}
 {% set _image = image|default('') %}
 {% set _credit = credit|default('') %}
 {% set _link = link|default({}) %}
+{% set _sr_play = sr_play|default('') %}
+{% set _sr_pause = sr_pause|default('') %}
+{% set _icon_path = icon_path|default('') %}
+
+{# DEPRECATED #}
+{% set _variant = variant|default('') %}
```
planctus, line 70: - confusing, why is the variant "deprecated"? We still set the corresponding selector, we still have styles associated to it, we still kind of show a story per variant..

- We say in the docs (readme e comment block) that plain-background is the default for this parameter but instead there is no default.

- In the demo it seems really odd to have "image" and "text-box" looking almost exactly the same, "image" seems a bugged implementation of "text-box" with a missing value for the variant, like this...

I don't know, i find the stories really confusing, in the end it seems we have:

- plain background
- text-box (with image or video)
- text-overlay (with image or video)

Why not showing these three stories and having a control to switch between image and video, then..?

src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -43,8 +83,24 @@
   {% set _css_class = _css_class ~ ' ecl-banner--' ~ _size %}
 {% endif %}
 
-{% if _centered %}
-  {% set _css_class = _css_class ~ ' ecl-banner--centered' %}
+{% if _font_size != 'm' %}
+  {% set _css_class = _css_class ~ ' ecl-banner--font-' ~ _font_size %}
```
planctus, line 90: but..if there are only two options available and we only set this class in case it's not 'm'; isn't it better to harcdode the "l", to avoid potential randomness..?

src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -43,8 +83,24 @@
...
+{% endif %}
+
+{% if _font_color != 'dark' %}
```
planctus: A bit the same as before, we do not check for the value of the parameter to be in the "list" of allowed values, so if we only have two options and we "exclude" one, the other is given, no need to use the string as it comes from the user.

src/implementations/twig/components/banner/banner.html.twig
planctus, line 73: We still have in the docs plain-background as the default for this, both in the template and in the readme

src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -43,8 +83,24 @@
...
+  {% set _css_class = _css_class ~ ' ecl-banner--color-' ~ _font_color %}
+{% endif %}
+
+{% if _background != 'light' %}
```
planctus: Shouldn't be this two tied together, meaning that if the background is dark then the font is light and viceversa? We request two parameters to handle two options, basically:

-  dark background with light font
- light background with dark font

The choice of a light or dark font should be limited to the transparent background (cool that we added this option, so now people will start complaining because of the text not being readable in random viewports depending on the image or video)

Currently in the stories you can set one of the two resulting in a unreadable text, but i'm saying that this should happen at the template level, imho, not just in the storybook.
Also, the name of the parameter can be confusing, generically backgrond makes you think that you will change the background color of the whole element

src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -68,19 +124,72 @@
...
+        extra_image_classes: 'ecl-banner__image',
+        extra_attributes: [{ name: 'data-ecl-banner-image' }],
+      } only %}
+    {% if _credit is not empty %}
+      <footer class="ecl-banner__credit"><small>{{ _credit }}</small></footer>
+    {% endif %}
+    </figure>
+  {% elseif _video is not empty %}
+    {# Video #}
+    {% set _video_sources = _video.sources|default({}) %}
```
planctus: being these two iterables they should be an array by default, no..?

src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -68,19 +124,72 @@
...
+    {% set _video_tracks = _video.tracks|default({}) %}
```
planctus: i think it would make sense to do the same we do for the picture also for the video (meaning having a dedicated template) now that we are spreading it in different components. Maybe not in this pr, but that way we could also add support for custom classes and attributes.

src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -68,19 +124,72 @@
...
+    <div class="ecl-banner__video-container">
+      <video class="ecl-banner__video" poster="{{ _video.poster }}" autoplay loop muted data-ecl-banner-video>
+        {% if _video_sources is not empty and _video_sources is iterable %}
+          {% for video in _video_sources %}
+            <source src="{{ video.src }}" type="{{ video.type }}">
+          {% endfor %}
+        {% endif %}
+        {% if _video_tracks is not empty and _video_tracks is iterable %}
+          {% for track in _video_tracks %}
+            <track src="{{ track.src }}" kind="{{ track.kind }}" srclang="{{ track.src_lang }}" label="{{ track.label }}">
```
planctus, line 168: we never had issues with it in the media container but this lacks a number of checks, all these attributes shoult not be set if there is no value for them. If we move this into a dedicated template we might take care of this there.

src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -90,7 +199,15 @@
       <div class="ecl-banner__content">
         {% if _title is not empty %}
           <div class="ecl-banner__title">
-            <span class="ecl-banner__title-text">{{ _title }}</span>
+            <span class="ecl-banner__title-text">
+            {%- if _title.link is defined -%}
+              {% include '@ecl/link/link.html.twig' with _title|merge({
+                extra_classes: 'ecl-banner__title-link'
```
planctus: this is a link object, right, so it might already contain extra_classes, we should merge them with the one we add here.

src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -23,15 +49,29 @@
 {% set _css_class = 'ecl-banner' %}
 {% set _extra_attributes = 'data-ecl-banner data-ecl-auto-init="Banner"' %}
 {% set _picture = picture|default({}) %}
-{% set _variant = variant|default('plain-background') %}
+{% set _video = video|default({}) %}
 {% set _size = size|default('m') %}
-{% set _centered = centered ?? false %}
+{% set _font_size = font_size|default('m') %}
```
planctus, line 57: wow this is what they want when a user chooses the "m" font..? 
image
ugly..

planctus (commented):
Total madness, now the editors they have to study to be able to use this banner, so many options that it's even difficult to get what they are about.
It seems mostly all good to me, only left a few comments that might lead or not to some improvement, just the docs maybe need to be updated for sure about the variant param. Thaanks!

src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -43,8 +86,35 @@
...
+{% if _font_size is not empty %}
+  {% set _css_class = _css_class ~ ' ecl-banner--font-' ~ _font_size %}
+{% endif %}
+
+{% set _color = '' %}
+{% if _font_color is not empty %}
+  {% set _color = _font_color %}
+{% endif %}
+
+{% if _box_background is not empty %}
```
planctus, line 98: mmh..how would this be empty..? For a transparent box we expect a string "none" and since we set the default to light it cannot be empty.

planctus (approved):

PR #3473, opened 2024-07-11
feat(inpage-nav): improve mobile keyboard navigation - FRONT-4501

emeryro:
- use arrows to open/close the inpage navigation
- focus first item on open

### DIFF
--- a/src/implementations/vanilla/components/inpage-navigation/inpage-navigation.js
+++ b/src/implementations/vanilla/components/inpage-navigation/inpage-navigation.js
@@ -93,6 +93,8 @@ export class InpageNavigation {
     this.observer = null;
     this.stickyObserver = null;
     this.isExpanded = false;
+    this.toggleElement = null;
+    this.navLinks = null;
 
     // Bind `this` for use in callbacks
     this.handleClickOnToggler = this.handleClickOnToggler.bind(this);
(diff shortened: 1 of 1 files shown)


src/implementations/vanilla/components/inpage-navigation/inpage-navigation.js
```diff
@@ -332,15 +334,15 @@ export class InpageNavigation {
       onDeactivate: () => this.closeList(),
     });
 
-    if (this.attachClickListener && toggleElement) {
-      toggleElement.addEventListener('click', this.handleClickOnToggler);
+    if (this.attachClickListener && this.toggleElement) {
+      this.toggleElement.addEventListener('click', this.handleClickOnToggler);
```
planctus, line 338: did not look at the whole file but it seems that we are adding event listeners and not removing them in destroy() here

planctus:
surely not because of this pr, but we have a flaw in the focus styles for the element, the label moves horizontally, maybe we can fix this here. Also the position of the icon seems not to be centered anymore ( in EC)

planctus (commented):
i've just noticed a couple of issues related to the css, we can fix those here or just create a separate task for them, the scope of the pull request is already fullfilled

emeryro:
let's handle this in another PR. This may be related to the scale/zoom setting, but on my side only the icon is "moving" with the focus, so to be investigated

planctus:
Oh true, i forgot i had the scaling still active in my desktop ;)

PR #3474, opened 2024-07-11
feat(button): setting height, using same values for padding in default and hover - FRONT-4499

planctus:
Pretty straightforward, we do not have situations where the button really needs a different border in default/hover, so the size of it is set by variant and stay the same in the two states and so the padding.
This might be enough, but still we would get sub-pixeling in scaled desktops so a min-height has been set.


emeryro (changes requested):
That's nice for the default and hover, but there seem to be some issue for the focus. I didn't check all the possible options, but for instance some outline are smaller now (1px instead of 2), and there is no focus anymore on the ghost inverted
image
image

planctus:
Right..i haven't even looked at the focus styles assuming that at least those would consistently use the outline, but it was not the case, we had of mix of border and outlines, now all the focus styles are coming from an outline so not bothering us about dimensions at all.

emeryro (changes requested):
Almost good, there is just a display issue for the EU search form, when in the site header
image

Also I guess that we should do the same thing for the "link buttons" (could be in another task)

planctus:
Indeed, not sure it was the right thing to "extend" the scope of this pr out of the button itself, but probably better not breaking things around for then having to fix them.
So i worked on the search form as well, removing tons of styles related to padding customisations in hover and focus, i've also cleaned up the site header css, ec and eu removing most of the customisations as well, this also involved minor changes to menu and mega menu styles (for the button).

emeryro (commented):
That's a nice cleanup!
We are indeed going outside of the scope of the PR...
I noticed another side effect of the button update: in the menu the carret is no longer centered in the button.
I let you see if you would rather check it here or in a task related to the menu
image
image

emeryro (approved):

PR #3477, opened 2024-07-12
feat(input): Using outline for focus style - FRONT-4504

planctus:


emeryro (approved):
Looks fine.
If I recall, we used the border instead of outline because outline did not follow border radius; that's not the cas anymore as it seem, so all good.

Shouldn't we also apply this to the text area?

planctus:
yep, i'm going component by component, starting from the ones that are currently "breaking", like the ones in the site header, still have to review all the components but textarea is surely on the list ;)

PR #3486, opened 2024-07-15
feat(mega-menu): Adding aria-describedby in info link - FRONT-4520

planctus:
- add aria-describedby in the info item title
- add support of "space" to open menu items


src/specs/components/mega-menu/demo/data.js
```diff
@@ -36,6 +36,7 @@ module.exports = {
       path: exampleLink,
       info: {
         title: 'About the News and Media',
+        title_id: 'about-the-news-id',
```
emeryro: do we really need to set a new parameter for this? Couldn't we just generate a unique id based on the global mega menu id?

planctus: it's the usual situation where we need to generate ids but we cannot use those for the tests, if you prefer i can remove these from the specs and only add them in the tests, in the template if this is not provided an automatic id is generated

emeryro:
Well, my point was: instead of using random ids, can't we generate it based on the id provided to the menu item?
Something similar to what is done a few lines before
image
If I'm not wrong, there is just one featured title, so we could have `_id ~ "-featured-title"` for instance

emeryro:
I may be mixing up things a little, as the id has to be set on the menu item title, but you get the idea

planctus:
I've added support for using Space on the menu items in this pull request.

emeryro:
This id are fine 👍, but now there is a small issue with the space behavior.
You can open the menu item, but not close it with space

planctus:
To be honest there was no intention to have a further action pressing again on the space bar. When you click Enter, for instance, a following press on it will not close the dropdown, it will instead behave as the Tab. Anyway, now we perform a click on the element when using the space bar, so it behaves as if it was clicked.

emeryro (approved):

PR #3488, opened 2024-07-15
feat(mega-menu): Not moving the focus when opening the menu - FRONT-4518

planctus:


emeryro:
It is fine for desktop, but on mobile the focus is lost, as the first level item is not displayed. 
I'd say that we should put the focus on the "back" when it is there (this is the first visual item), and leave it on the menu item if not

planctus:
it's not lost, is there where Regine asked to keep it, it's just not visible. Anyway, done it the way you wanted it.

emeryro (approved):

PR #3489, opened 2024-07-15
feat(social-media-follow): add position right - FRONT-4505

emeryro:
- new option to display the Social Media Follow on the right. Twig parameter: `position`

### DIFF
--- a/src/implementations/twig/components/social-media-follow/README.md
+++ b/src/implementations/twig/components/social-media-follow/README.md
@@ -10,7 +10,7 @@ npm install --save @ecl/twig-component-social-media-follow
 
 - **"variant"** (string) (default: '') Can be 'vertical'
 - **"description"** (string) (default: '')
-- **"position"** (string) (default: 'left') pPosition (can be 'left', 'right')
+- **"position"** (string) (default: 'left') Position (can be 'left', 'right')
 - **"links"** (array) (default: []) Array of links for social media, following ECL Link structure
 - **"extra_classes"** (optional) (string) (default: '') Extra classes (space separated)
 - **"extra_attributes"** (optional) (array) (default: []) Extra attributes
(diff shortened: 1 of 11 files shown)


src/implementations/twig/components/social-media-follow/README.md
```diff
@@ -10,6 +10,7 @@ npm install --save @ecl/twig-component-social-media-follow
 
 - **"variant"** (string) (default: '') Can be 'vertical'
 - **"description"** (string) (default: '')
+- **"position"** (string) (default: 'left') pPosition (can be 'left', 'right')
```
planctus:
```suggestion
- **"position"** (string) (default: 'left') Position (can be 'left', 'right')
```

src/implementations/vanilla/components/social-media-follow/social-media-follow-print.scss
```diff
@@ -26,6 +28,8 @@ $social-media-follow-print: null !default;
 }
 
 .ecl-social-media-follow__list {
+  display: flex;
+  flex-direction: column;
```
planctus, line 32: Why do we go vertical in print..? Was this requested, it takes much more space and being used at the bottom of the pages most of the times it doesn't seem a good idea to make it a vertical list, also because the alignment is pretty odd when we do it.

emeryro: I was wondering the same, and the best explanation I got was that we don't really have a choice, as the extra links would just take too much space horizontally (they can be quite long url)

planctus: we could just use flex-wrap in print instead of going vertical, if that is the issue, but what can i say, let's keep doing things irrationally

PR #3490, opened 2024-07-16
feat(mega-menu): Making the hamburger a button with javascript - FRONT-4519

planctus:
…


emeryro (changes requested):
That's one way to handle these recurring request, indeed...
Not really of fan of these DOM modification, but if it could close this topic for once, fine for me.

One thing missing though: we no longer have the button/link label changing based on the context (previously it switched between "Menu" and "Close")

planctus:
I would hardly define myself as a fan of this, this was a way to fulfill the request while still maintaining the no-js fallback approach. I fixed the label switch

emeryro (approved):

PR #3494, opened 2024-07-17
feat(mega-menu): Using buttons for parent elements - FRONT-4514

planctus:


emeryro (commented):
Not sure if this is linked to the changes here, but on my Chrome (only), there is a small position difference for the button label when it is selected. Cannot reproduce on Firefox
image
image

planctus:
yep, i also see that in chrome, it is related to the "custom" line height used in the mega-menu. Now, i cannot entirely predict what is going to happen putting everything together, we have this change in the markup, a pull request where we "standardize" this custom line-height and another one adding a border to the items (second level).
We identified the issue, but let's see what happens once we have everything in place, it will be easier to tacke it if it's still in place.

emeryro (approved):

PR #3499, opened 2024-07-18
feat(form): update required fields - FRONT-4495

emeryro:


planctus:
oh, wow..i struggle to understand the sense of this, basically we reduced the visibility of the "default" marker a lot, also making it "blue" close to a dark text is visually far from being appealing, i do not really get the reason why of this style change. It's also a bit confusing the fact that the "required" marker is red when invalid while the optional stays "blue".
Anyway, if this is what they want..

planctus (approved):

PR #3501, opened 2024-07-18
feat(inpage): Consistent border and padding in different states - FRONT-4508

planctus:
…


emeryro (changes requested):
On EU we lost this outline on focus (mobile)
image

emeryro (approved):

PR #3504, opened 2024-07-18
feat(gallery, media container): add title to videos - FRONT-4526

emeryro:
- update markup to have the title for HTML videos
- provide title to iframe video
- for media container, parameter `sr_video_label` has been renamed `sr_video_player`, to be consistent with gallery. Backward compatibility is in place.


planctus (commented):
Oh wow, what a mess now..
It's unbelievable the growing complexity of all these operation to enhance accessibility beyond reasonability, imo. We were trying to standardize the video, now we will have to evaluate this, merge it and then re-work again the video template, lovely..

src/implementations/twig/components/media-container/media-container.html.twig
```diff
@@ -106,7 +116,14 @@
 
 {# Print the result #}
 
-<div class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+<div class="{{ _css_class }}"{{ _extra_attributes|raw }}
```
planctus, line 127: Maybe i just fail to understand this, but these two separate parameters are actually going to be concatenated in a pretty elaborated and risky way by the js. So why do we need two parameters? As far as i can see they are simply never used individually, we build in a complicated way an aria-label combining the two, and we provide two separate html attributes for the js to combine their value and set a title in an iframe.
I do not understand:

- why do we have two parameters
- why, if we combine these two string in the template to generate an aria-label then we expect the js to retrieve the two values independently and re-combine the same string (hopefully, cause those strings might get escaped).
- so, to me we could/should use one parameter only and use that in the html attr, with all the risks involved by setting a string as the value of an html attribute, ther would not be any need to combine those strings and surely not to replicate the same procedure in js.

Forgive me if i missed something..

src/specs/components/media-container/demo/data--embed-video.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
+  sr_video_player: 'Video player',
+  title: 'New digital strategy',
   description:
     'The European Commission has put forward ambitious yet realistic proposals for a modern EU budget. It is time for an EU budget that reflects rapid developments in innovation, the economy, the environment and geopolitics, amongst others.',
   embedded_media:
-    '<iframe title="Video player" width="350" height="197" src="https://www.youtube.com/embed/fgi-GSCB6ho" frameBorder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowFullScreen></iframe>',
+    '<iframe title="New digital strategy - Video player" width="350" height="197" src="https://www.youtube.com/embed/fgi-GSCB6ho" frameBorder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowFullScreen></iframe>',
```
planctus: why do we hardcode the title here? This way we are not really going to see this attribute added by the js..

src/implementations/vanilla/components/gallery/gallery.js
```diff
@@ -480,8 +483,19 @@ export class Gallery {
...
 
+      // Update iframe title
+      let videoTitle = '';
+      const title = queryOne(this.titleSelector, selectedItem);
+      if (title) {
+        videoTitle = title.innerHTML;
+      }
       if (this.videoPlayerLabel) {
-        mediaIframe.setAttribute('title', this.videoPlayerLabel);
+        videoTitle = videoTitle
```
planctus: but also here...i mean, but why would we have to do something like this..? Retrieving two strings and combining them? It is also something that happens without the user being aware of that, why don't we simply expect the final label we construct from ONE single parameter, instead? It's a crazy overhead for no reason, in my opinion, the user will still have to pass two params and not knowing that they will be magically combined together, isn't it just better to expect from this videoPlayerLabel whatever the string to be added as the title of the iframe?

planctus (commented):
It looks to me we are way over a limit of reasonability, all of this to get an aria-label or a title attribute as the result of so much info and so much logic that might easily generate buggy strings due to the escaping.
In the commnents i left i've tried to simplify the approach, maybe i'm wrong and it cannot be simplified, but it really feels like shooting with a cannon for catching a butterfly..

src/implementations/twig/components/media-container/media-container.html.twig
emeryro: That's interesting. What you say is true indeed for the media container: we have 2 fields and only use them in the iframe title.
But there are a few other elements to be considered:
- currently it is the same thing on the gallery (2 different fields), the main difference being that the title is also displayed on the screen there
- the title in media container may or may not be used later for something else (this is not limited to videos)

So I would say that it's up to us to decide: keep the same structure for gallery and media container (even if that's a bit useless for now on the media container), or use a single field only in the media container

src/implementations/vanilla/components/gallery/gallery.js
emeryro: I'm not sure that the editor would understand that they are supposed to use a label like "[video title] - Video player" (that's what is requested at accessibility level...)
What we could do is concatenate the string before sending it to the js, so we only check for one data-* thingy

planctus: yes, that's already better, but my point of view is that the user will not understand anyhow what to do with these things, the more they have the less they will understand..

src/implementations/twig/components/media-container/media-container.html.twig
planctus: And let's keep the two parameters, but at least we combine this string in twig only and we don't duplicate this in the js

PR #3506, opened 2024-07-22
feat(video): Adding template for video, using it in different components - FRONT-4497

planctus:


src/implementations/twig/components/gallery/__snapshots__/gallery.test.js.snap
```diff
@@ -3138,7 +3138,7 @@ exports[`Gallery Default renders correctly with old data 1`] = `
             </div>
             <video
               aria-label="EU law - Video"
-              class="ecl-gallery__image"
+              class="ecl-video ecl-gallery__image"
               poster="https://inno-ecl.s3.amazonaws.com/media/examples/example-image5.jpg"
```
emeryro: For some reason we lost the change in the gallery data, using another html video.
It could be good to bring it back while we are changing the gallery and also use it in the video component data)
https://github.com/ec-europa/europa-component-library/pull/3554/files#diff-330b93090781d35d9db384dca62ddc8ab9726a69caccb37617070e162ff45beaR92

src/implementations/twig/components/video/video.html.twig
```diff
@@ -0,0 +1,110 @@
+{% apply spaceless %}
+
+{#
+  Parameters:
+    - "image" (string) (default: '')
```
emeryro: should be `poster` here I guess

src/implementations/twig/components/media-container/media-container.html.twig
```diff
@@ -3,22 +3,7 @@
...
-      {
-        "src" (string) (default: ''),
-        "kind" (string) (default: ''),
-        "src_lang" (string) (default: ''),
-        "label" (string) (default: ''),
-      },
-      ...
-    ],
+  - "video": (associative array) (default: {}) Object of type ECL video
   - "image" (string) (default: ''): Image to be used as the video placeholder
```
emeryro: the `image` parameter should be deprecated

emeryro (changes requested):
Just some small data updates

emeryro (approved):

PR #3507, opened 2024-07-23
feat(logo): update logo markup - FRONT-4480

emeryro:
in splash page, site header and site footer:
- remove unneeded aria-label
- set title as optional


planctus:
But, why in the splash page the aria-label for the logo link is not needed and we removed it while we are keeping it in the site-header as well as in the site-footer? Same for the picture "title" attribute

emeryro:
This task was for the splash page, but we can extend it to the site header and footer indeed.
For the title, I guess that it is not strictly needed to have it if we have the alt, but as it adds something also for sighted user I kept the option open (just not filled by default in our example)

PR #3510, opened 2024-07-23
feat(footer): 22px line height (from the standalone link) for all the links, EC and EU - FRONT-4511

planctus:


src/themes/ec/_custom-properties.scss
```diff
@@ -66,6 +66,7 @@
   --ecl-font-4xl: #{map.get($font, '4xl')};
   --ecl-font-ui-s: #{map.get($font-ui, 's')};
   --ecl-font-ui-m: #{map.get($font-ui, 'm')};
+  --ecl-font-ui-link: #{map.get($font-ui, 'link')};
```
emeryro: I think that this custom property is not used, nor defined, as there is no corresponding sass map

emeryro (changes requested):

src/themes/ec/_custom-properties.scss
planctus: ah, forgot to remove that in the last commit ;)

emeryro (approved):

PR #3515, opened 2024-07-24
feat(featured-item): change order - FRONT-4476

emeryro:


planctus:
This rule is now creating an additional spacing in the item containing the media: 
image

planctus:
is this accessibility request implying also a total change in the mobile design..? 
image

we used to have the image on top and now it's the opposite

planctus:
We have this in mobile for the highlighted: 
image
no "spacing"after the link, but maybe this comes simply out of the inversion of the items order

planctus (approved):

PR #3519, opened 2024-07-25
feat(splash-page): use headings - FRONT-4481

emeryro:
- use headings instead of div for titles


planctus (commented):
Also in this case, we know that this markup is just the same also in the language switcher, why changing it only here in the splash..? I mean, what needs to be imprvoed is the code of the "language list", not anything specifical to the splash page.

emeryro:
We can't really put headings in the language switcher. It is placed before the page header, containing the H1.
In the splash page there is no other content, so the main title is clearly defined

PR #3523, opened 2024-07-26
chore(deps): Upgrading iframe-resizer, removing usage of svg4everybody - FRONT-4539

planctus:


emeryro (changes requested):
I didn't check all the components, but there seem to be some issue with some iframe height on the website.
For instance tabs and modal
image
image

planctus:
Yep, they removed maxHeight, maxWidth from the options, we have now to set those as inline styles and as the script is going to set its own height we set instead min-height so that it will not be overridden by a lower computed value.

emeryro (approved):

PR #3527, opened 2024-07-29
chore(sass): Declarations before nested rules - FRONT-4540

planctus:
https://sass-lang.com/documentation/breaking-changes/mixed-decls/


emeryro (changes requested):
it seems that the css reordering had some effect on the pagination (current item):
image
image

planctus:
mmh...these styles in the pagination are a bit confusing.. :)

emeryro (approved):

PR #3533, opened 2024-07-30
fix(site-header): Preventing rendering of an empty banner - FRONT-4550

planctus:


emeryro (changes requested):
I think that the banner is needed, even if empty.
Without it the menu is not displayed correctly on mobile
image
Indeed on core site the banner is empty, but it usually contains the site name

planctus:
The issue is that we were assuming the banner was in place in the mega-menu js, so it was not getting the dynamic positioning anymore, fixed.

PR #3534, opened 2024-07-30
feat(mega-menu): aria-labelledby for the featured links - FRONT-4543

planctus:


emeryro:
That should be what was was asked, but it seems to be wrong: the link label longer read by the screen reader...
I'll send the preview to Regine

emeryro (approved):

PR #3538, opened 2024-07-31
feat(gallery): grid templates - FRONT-4622

emeryro:


src/specs/components/gallery/demo/data.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
       // Image with thumbnail
       thumbnail: {
         img: {
-          src: 'https://inno-ecl.s3.amazonaws.com/media/examples/example-image.jpg',
+          src: 'https://commission.europa.eu/sites/default/files/styles/oe_theme_medium_no_crop/avportal/P-035170/00-08.jpg?',
```
planctus: are we going to use a file from the EC production website..?

src/implementations/twig/components/gallery/gallery.html.twig
```diff
@@ -80,6 +82,10 @@
 {% if _grid %}
   {% set _css_class = _css_class ~ ' ecl-gallery--grid' %}
 
+  {% if _grid_template is not empty %}
+    {% set _css_class = _css_class ~ ' ecl-gallery--template-' ~ _grid_template %}
+  {% endif %}
+
   {% if _column is not empty %}
```
planctus: this selector is irrelevant when using a "grid template" right? If i'm correct better not setting in those cases and also in the story, if possible, would be less confusing if we would hide the control for the number of columns

planctus (commented):
I guess that the feeling is more or less the same about this, this overengineering of the gallery component doesn't seem to come to an end. Anyway, the different layouts work well, i just left a couple of comments. Thanks

src/implementations/twig/components/gallery/gallery.html.twig
emeryro: In fact it is used only when using the "default" grid template (a plain grid without any modification). I'll add a test for that in the twig template, and check the story

PR #3543, opened 2024-08-01
feat(form): update for optional elements - FRONT-4319

emeryro:
- allows custom optional text for single checkbox, for consistency
- fix single checkbox demo controls
- remove unneeded aria-label in the form demo (no modification of the components themselves)


planctus (changes requested):
I guess this is a side effect of this pr: 
image
but it's not entirely clear to me what we are doing here, because also this doesn't seem right: 
image
but this would mean informing the form element about the presence of a label from the group template...

emeryro:
that was partially done, but with some edge case multiple "optional" were displayed. This is fixed.
This idea is to have the default form group status, as in any other form element, and have the one displayed after the checkbox only if it is a single checkbox (it was already the case for the "required" label)

planctus:
but i don't understand this: 
image
guess that when the form group renders a label the marker should only be shown there, even in a single checkbox, no?

emeryro:
How did you get that display? The last fix I pushed should prevent this...
Indeed, if the checkboxes are in a form group, there should be no "optional" on single checkbox

planctus:
in the preview :)
Just show the label in the single checkbox and you will get this, both for required and optional ;)

emeryro:
I should be missing something here... Indeed for the single checkbox we can have the optional or required (that was the case before for the required, I just added the optional), but when it is in the form group, single checkboxes no longer have this extra label
image
image

planctus:
but also a single checkbox can have a form group label no..? might not be mandatory but it is possible, at least, that is the problematic use case when we have the duplicated label for the required/optional, in production it looks like this, since we were not applying the label to the form element: 
image

emeryro:
ok, finally got it, should be fine now

planctus (approved):

PR #3546, opened 2024-08-05
fix(breadcrumb): fix wrap when expanding - FRONT-4558

emeryro:
- fix an issue where the breadcrumb was not displayed entirely when expanded


planctus (approved):

src/implementations/vanilla/components/breadcrumb/breadcrumb.js
```diff
@@ -208,6 +208,7 @@ export class Breadcrumb {
    */
   handleFullExpand() {
     this.element.classList.remove('ecl-breadcrumb--collapsed');
+    this.element.classList.add('ecl-breadcrumb--wrap');
```
planctus, line 212: We should remove this class we add on destroy() if it's present, it's not really going to happen but imagine the case where the instance is destroyed and re-initialized, it would not prevent wrapping.

PR #3551, opened 2024-08-07
feat(menu/mega-menu): remove aria-current=true - FRONT-4545

emeryro:
- information about opened item is provided by the aria-expanded


planctus (commented):
Really the justification for this is that we use an aria-attribute to mark the item as expanded and then aria-current would be redundant?
It is pretty much clear that the two aria attribute are meant for different things, to me here an argument is being used only to impose a questionable decision.
I like this in code, i mean that it simplifies things, but really it feels like we are confusing apples for pears here.

PR #3552, opened 2024-08-07
feat(headings): set spacing - FRONT-4566

emeryro:
- set same spacing for all headings; it applies to utilities (`.ecl-u-type-heading-x`) and heading used in wysiwyg, using `.ecl hx`


planctus:
mmh..does it make sense to add this to utilities that are only meant to deal with the font styles? I mean, if a utility class is used then it's not an issue to add another one, to set a margin being able to decide this margin every time.
I thought this would have only been affecting the ecl-default, basically, to make it possible to quickly create a layout with heading and text without the overhead of creating a spacing between them.

emeryro:
I would say yes and no... We can indeed guess that if a user adds a utility somewhere, he is able to add 2 more. But then he would have to know what spacing should be used, to keep consistency.
Adding it directly in the heading utilities is more straightforward, even if that goes a little outside of the "typography" scope...

planctus:
but, imagine all the current usage of these utilities, now they will get a margin where they hadn't before, i really don't see why doing this, it's not just a matter of consistency, it's really that doing this will surely have an impact and i don't see the need for this, we can offer a way to handle spacing through headings in a specific context like what we do in the default css, no need to extend this to the utilities, imo.

emeryro:
Just to be clear: there is already a margin on all headings; it is just provided by the browser. This update would set it to a fixed value.
But let's check that with COMM, to be aligned

PR #3553, opened 2024-08-08
feat(gallery): add thumbnail - FRONT-4565

emeryro:
- add possibility to have a thumbnail, different from the full picture
- this thumbnail is optional; if not provided, the main image will be used as it was the case before


planctus:
mmh..i understand that this new feature might actually come from the request to fix a "wrong" behavior in the overlay, but this way we are not really doing much to improve the situation we had regarding the amount and the heaviness of the images loaded, actually we are doing worst, since now we will have two picture objects per items, both images we will be downloaded by the browser.
To me we should do more or less what we have been doing in BCL, where we set the "src" attribute via js when the overlay opens, so that it will only be downloaded when it is requested in the overlay.

planctus (commented):
- We cannot use the same picture element for the thumbnail and the overlay, if sources are provided nothing is ensuring that the "thumbnail" image will be shown in the gallery list.

- The previous implementation seemed ok, we probably don't need to have a picture element for the thumbnail itself, a simple img could be fine for that, but in any case the problem was that the second picture element had a src attribute defined, and with that the image would have been downloaded at page load.

So the idea was just to set an attribute in the second (hidden) picture element instead of the src, for then replacing this via js when needed.

- there is a problematic css at line 96, with multiple sources defined each src would take some space and the rendered image would not really "cover" the available space, this currently happens both in the list and in the overlay
image

So, my proposal is:

- make the thumbnail a simple img or keeping it as a picture, but separated from the overlay one

- having a picture object for the overlay image hidden at page load

- rendering the img in the overlay picture element without the src attribute, like with a data-src attribute

- replacing the data-src attribute with src when the item is clicked

- introducing some examples with multiple sources, we have none at the moment and this is probably the reason why certain issues cannot be identified in the current demo

src/implementations/twig/components/gallery/gallery-item.html.twig
```diff
@@ -98,9 +104,11 @@
   {% endif %}
   >
   {% if _item is not empty %}
-    {% set _image_alt = _item.picture.img.alt|default('') %}
+    {% set _image_alt = _thumbnail.img.alt|default('') %}
```
planctus, line 107: i think this was the case before, _thumbnail is a string now

src/implementations/twig/components/gallery/gallery-item.html.twig
```diff
@@ -119,15 +127,19 @@
         </div>
       {% endif %}
     {% endif %}
-    {% if _item.video is defined and _item.video is not empty %}
+
+    {% if _item.embedded_video is not empty %}
+      {# Embedded video #}
+      <img src="{{ _thumbnail }}" alt="{{ _picture.img.alt|default('') }}" class="ecl-picture ecl-gallery__thumbnail ecl-gallery__image" />
```
planctus: mmh...i'm thinking that this might break something on the application side, it is not possible to pass custom attributes to this img tag and i think it's required on ewpp side, probably for analytics, so we could go back to the object for the thumbnail, or we could add another param where to pass those attrs and then process them in the template. Forget about this in case even before we did not support custom attributes for the picture, in that case no need to add it now.

src/implementations/vanilla/components/gallery/gallery.scss
```diff
@@ -99,9 +100,18 @@ $_description-height-desktop: 108px;
...
+.ecl-gallery__picture {
+  display: flex;
+  height: 100%;
+
+  &:not(.ecl-gallery__slider-image) {
+    display: none;
+    flex-grow: 1;
+  }
+
+  * {
```
planctus: why do we need this..? From what i see these two properties are overridden but this will still apply some styles to the "sources" although it doesn't seem to break anything, but still it seems this could be removed.

src/specs/components/gallery/demo/data.js
```diff
@@ -167,10 +170,10 @@ module.exports = {
       title: 'About the EU',
       description: 'About the EU',
       meta: 'Copyright, Author, Licence for image 9',
-      share_path: '/share#example-imag9e.jpg',
+      share_path: '/share#example-image9.jpg',
     },
     {
-      // Image
+      // Image without thumbnail
```
planctus, line 199: maybe it would be easier for the users if we apply the thumbnail to the "majority" of the items we demo, otherwise they will have to look for the one that has a different image defined as the thumbnail and the one in the picture tag. I mean, i would keep like one without, just to test the backward compatibility of what we are doing.

planctus (commented):
Minor things, i think we are ok with this, we should only ensure that any "attribute" they could pass before is still going to work, if none was available before, then we are safe..

src/specs/components/gallery/demo/data.js
emeryro: I thought about it, but we don't have any more images available, and reusing them could be confusing.
I will put the thumbnail on the first image, so that's more visible

src/implementations/twig/components/gallery/README.md
```diff
@@ -63,6 +64,12 @@ npm install --save @ecl/twig-component-gallery
   sr_video_player: 'Video player',
   items: [ 
     { 
+      picture: {
```
planctus: it's almost irrelevant since this will be overridden in storybook with the real data, but it should be "thumbnail", right, not "picture"

emeryro: updated

src/implementations/twig/components/gallery/gallery-item.html.twig
```diff
@@ -119,15 +127,34 @@
...
-      {% include '@ecl/gallery/gallery-video.html.twig' %}
-    {% else %}
+
+    {% if _item.embedded_video is not empty %}
+      {# Embedded video #}
       {% include '@ecl/picture/picture.html.twig' with {
-        picture: _item.picture|merge({
-          img: _item.picture.img|merge({
+        picture: _thumbnail|merge({
+          sources: [],
```
planctus: mmh..why do we do this..? it's not surely that relevant to define many images for a thumbnail and we don't expect that, but why resetting the sources, in case someone spent the time to define different images for their thumbnail? If this is really data passed as thumbnail we would delete part of it, if it's not and we are "cloning" the main picture, we are basically in the situation we had before, no reasons why resetting the sources, then.

src/implementations/twig/components/gallery/gallery-item.html.twig
```diff
@@ -119,15 +127,34 @@
...
         }),
+        extra_classes: "ecl-gallery__thumbnail",
+        extra_image_classes: "ecl-gallery__image",
+      } only %}
+    {% elseif _item.video is not empty %}
+      {# HTML video #}
+      {% include '@ecl/gallery/gallery-video.html.twig' %}
+    {% else %}
+      {# Image #}
+      {% include '@ecl/picture/picture.html.twig' with {
```
planctus: this is just for avoiding some redundancy, couldn't we "set" this include once and then we just render it twice instead of redefining the whole object?

planctus (approved):

PR #3556, opened 2024-08-08
feat(inpage): Making the list scrollable, moving scrollbar to the left in EC - FRONT-4563

planctus:
The functionality is in place, but in EC it currently looks pretty bad, i had to move the scrollbar on the left because of the design but maybe in this case designers could choose a different color for the scrollbar in EC since it's the same as the border of the items.


emeryro (commented):
I'm not really convinced by the sidebar on left (I didn't realize that we are using custom sidebars on EC)... The position switch complexify the code
It may be better to check with the designers what we should do here, as we may have the same issue later in other components

Also, on EU we are using a color that does not exists
image

planctus:
I agree, but the same already happened for the mega menu, the style the designers create is most of the times not taking into account the scrollbar, therefore when we introduce them there is something wrong about the look and feel.
Handling this is not too complex, but per se is questionable the fact that the design of the component is "mimicking" a sidebar, so regardless of where we place it, it looks a bit odd, it seems there are two scrollbars, left and right, or right right.
Anyway, this is only problematic in desktop where we don't really expect to see the scrollbar in the inapge navigation often, or maybe just never, we support it but we don't really expect it to needed most of the times.

PR #3559, opened 2024-08-12
feat(menus): Removing role application from the inner section of the menus - FRONT-4568

planctus:


emeryro:
So, it seems to work for the mega menu (with some limitation considered as normal by Regine), but not for the single menu: items can no longer be opened when using a screen reader.
I'll try to see why

emeryro:
It should be fine now.
@planctus I let you have a look with the latest change, to check for regression

PR #3561, opened 2024-08-13
fix(site-header): fixing language switcher in rtl - FRONT-4569

planctus:


emeryro:
This looks fine, but we are in a confusing situation I think: we have the extra css for rtl, and still we are adding things related to rtl in the main css.
We should try to clean up a little that topic (probably in a major version)

planctus:
Yep, true, but i would really find it hard to handle situations like this in two separated css, it would be really unconvenient and the advantage in terms of heaviness of the "main" css is really minor, moving those rules outside of it.
I think i have never looked at the rtl css, i would expect it to be something more "generic" to make adaptations when needed, but handling the single components styles in it seems unconvenient, in the end it's a matter of the amount of changes needed for the rtl to work properly, in this case it's not really much and hopefully the same for many other components, otherwise we would probably decide to generate two scss files (three including the print) per component.

emeryro (approved):

PR #3562, opened 2024-08-14
feat(utilities): Adding margin-top and margin-bottom auto in the spacing utilities - FRONT-4573

planctus:


emeryro (changes requested):
To be consistent we should also add `.ecl-u-mv#{$infix}-auto` utilities

emeryro (approved):

PR #3569, opened 2024-08-16
feat(popover): Adding support for buttons as triggers - FRONT-4350

planctus:
I did this, but honestly if feels like this component should only deal with its content, the trigger should be whatever the user want to.
If you agree, we could do now something in between, keeping the support for a trigger defined by the template but extending the functionality to any element with a certain markup, like a data attribute.
Then in v5 we could definitively remove the logic and the markup related to the trigger from the template.


emeryro (commented):
Just one small thing: for the last example, it may be better to keep the popover as a link (or simply to remove that example...)
image

emeryro (approved):

PR #3571, opened 2024-08-19
feat(banner): copyright position, background and image height - FRONT-4570

emeryro:
- allows images to take 100% height, so we don't have white space around it when the content is too large
- remove background on the banner image (and gallery image), to allow semi transparent images
- change position and display of the image copyright


planctus:
not really sure it is an issue, maybe it's just as it should be, but since in the previous version we did not have this "overflow" i don't know: 
image
this happens when you align the content on the right and bottom.
Currently it looks like this: 
image

planctus (approved):

planctus:
I see that in print we don't display the image, shouldn't the same happen for a video?

planctus:
This is not coming from this pr, but it seems to me we have a problem in EU:
image
The controls container looks white in figma, and used to be white in the past, now it has this greysh background just a bit different from the credit's background

planctus:
also this doesn't seem to come from this pr, but i cannot really understand it: 
image
why do we set this additional margin in the carousel? it doesn't seem to be defined in the specs, there it seems everything is aligned at the container's edge.

emeryro:
I fixed the print display
For the carousel, there is another task to update the styles, so let's handle what you found there (that would be reported by the designers anyways...)

PR #3583, opened 2024-08-23
feat(carousel): Styles update - FRONT-4571

planctus:


emeryro (approved):
Great! I didn't know that `scroll-snap-align` property...

PR #3587, opened 2024-08-27
feat(table): improve border display - FRONT-4584

emeryro:
This is not a perfect solution, as there may be a darker border is the multi header element is at the end, but that's already better than the current situation
image
image


planctus:
mmhh..i don't know, on linux with firefox i do not have the "missing" border, it's really tricky with all those overrides, but i don't seem to be able to find a simpler approach, just that it's pretty easy to see the limits of this approach:
image
and although the border color would not be correct with this markup in the current situation it would still be better than what we would get now.

emeryro:
I pushed an update. On my side it seems to fix the issue on Chrome and Firefox

planctus (approved):

PR #3593, opened 2024-08-29
docs(showcase): provide example of full pages - FRONT-4383

emeryro:
Provide a few page example using ECL components

Note: pages are visible only in storybook, not sure that having them in a small iframe on the website would make sense

Available pages

/story/page-examples-home--default
/story/page-examples-inner--default
/story/page-examples-search--default


src/implementations/twig/pages/page-home/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@ecl/twig-page-home",
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "4.6.3",
+  "description": "ECL Example page - Home",
+  "publishConfig": {
+    "access": "public"
```
planctus: mmmh..these packages should be private, i think, someone might think that we are really providing a template for the home page and inner pages.

src/implementations/twig/pages/page-home/page-home.html.twig
```diff
@@ -0,0 +1,75 @@
+<!-- Site header -->
+{%- include '@ecl/site-header/site-header.html.twig' with site_header -%}
+
+<main>
+  <div class="ecl-container">
+    <!-- Banner -->
+    {%- include '@ecl/banner/banner.html.twig' with banner -%}
+
+    <h2 class="ecl-u-type-heading-2">Maecenas sagittis est</h2>
```
planctus, line 9: i noticed that we are not doing what i thought with the margins we added in the default css.
We set the margin for any h2 that contains an ecl utility class: .ecl h2:is([class*=ecl-u-] and i don't understand why.
It shouldn't be the case, this way we are basically adding this margin to any utility class used on an h2 inside an .ecl container and instead we should only add it in case no utility is set, just a generic h2, otherwise the reasoning is that if a utility is used, this utility might be already setting a margin or if it's not, another utility class can be added to set a margin on it.

So, i would fix this and in the examples i wouldn't set any utility class for the h2s since we have them in ecl container. Otherwise we can add both the utilities for font style and margin

src/implementations/twig/pages/page-inner/page-inner.test.js
```diff
@@ -0,0 +1,22 @@
+import { renderTwigFileAsNode, renderTwigFileAsHtml } from '@ecl/test-utils';
```
planctus: it takes fifteen second to run this test in my env, i wonder a bit what we are really testing here, we test components, i do not see the value in doing the same for a "page example".

planctus (commented):
To me this will open the door again for requesting an example for each content type or anything they might like ECL to do instead of them, any of these example will be hard to be maintained on our side, i personally do not see the real advantage of doing this but i see the potential, likely to happen, drawbacks.

src/implementations/twig/pages/page-home/page-home.html.twig
emeryro: So basically we should have something like
```
.ecl h2:not([class*='wt-']) {
  @extend %ecl-heading-2;
}
```
I also noticed that the default style for the separator is not correct either; it would stop to work as soon as we add any utility to it.
I'll create a separated ticket to handle all of that

planctus: mmh..it might be that we have some misunderstanding about how these selectors should be defined based on what they are supposed to do: my approach is that if a user is using an ecl-utility that is not "generic" html, if there is an ECL selector we might assume that the styles are already handled by those, also because we cannot know which utility is being used and having two diferent styles set using !important would really create an unpredictable situation.
Maybe i haven't shared these thoughts about this css in a clear way before, but that is the idea, generic means something that doesn't use ECL already, imo.
Open to different considerations, also because we are distributing this css since v4 so at this point better not altering the existing behaviors, but for the future maybe this can be taken into account.

src/implementations/twig/pages/page-home/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@ecl/twig-page-home",
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "4.6.3",
+  "description": "ECL Example page - Home",
+  "publishConfig": {
```
planctus, line 7: shouldn't this simply be a private package?

src/implementations/twig/pages/page-home/page-home.story.js
```diff
@@ -0,0 +1,95 @@
...
+        );
+        head.appendChild(createLink('./styles/ecl-eu.css', 'screen'));
+        head.appendChild(
+          createLink('./styles/optional/ecl-eu-utilities.css', 'screen'),
+        );
+        head.appendChild(createLink('./styles/ecl-eu-print.css', 'print'));
+        head.appendChild(
+          createLink('./styles/optional/ecl-eu-default-print.css', 'print'),
+        );
+      } else {
```
planctus: it's ok eh, but being this a duplicated section with the only parameter being "system" we could avoid this duplication and the maintenance needed

src/specs/pages/page-home/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@ecl/specs-page-home",
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "4.6.3",
+  "description": "ECL Example page - Home Specs",
+  "publishConfig": {
```
planctus, line 7: also here, i mean, more or less the result is the same, the package will not be installable by non members of the INNO organisation on npm, but for consistency i would use "private" explicitely as we did for all the other packages that we don't want to make public.

src/implementations/twig/pages/page-home/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@ecl/twig-page-home",
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "4.6.3",
```
planctus: let's keep this aligned with the current version

src/implementations/vanilla/components/file/file.scss
```diff
@@ -259,10 +259,6 @@ $file: null !default;
     width: auto;
   }
 
-  .ecl-file__info {
```
planctus, line 260: is this a needed fix? In that case i would also apply this as a regular fix on v4-dev directly without depending on this pr

planctus (commented):
Fine by me, just we need to update the package(S) version and make them "private", although they are already, in a way.

src/implementations/twig/pages/page-home/package.json
emeryro: As I understood, there is no "private" option for npm packages, just "public" or "restricted"
https://docs.npmjs.com/cli/v7/commands/npm-access

PR #3594, opened 2024-08-30
fix(video): Initializing the _extra_attributes variable - FRONT-4590

planctus:


src/implementations/twig/components/video/video.html.twig
```diff
@@ -38,6 +38,7 @@
 {% set _title = title|default('') %}
 {% set _sr_video_player = sr_video_player|default('') %}
 {% set _sr_video_label = sr_video_label|default('') %}
+{% set _extra_attributes = '' %}
```
emeryro, line 41: shouldn't it be an empty array instead of a string?

planctus: no, this is the local variable, it's a string

PR #3597, opened 2024-08-30
feat(banner): Resetting padding for the content in no-box variant - FRONT-4592

planctus:


emeryro:
In Figma we still have the padding.
Has it been agreed by the designers?
image

planctus:
it was requested and approved by Vlad, i guess he agreed with the designers

PR #3601, opened 2024-09-04
fix(site-header): fix icon alignment - FRONT-4596

emeryro:
Fix a small misalignment of icons on mobile

### DIFF
--- a/src/implementations/vanilla/components/site-header/site-header-eu.scss
+++ b/src/implementations/vanilla/components/site-header/site-header-eu.scss
@@ -329,14 +329,18 @@ $language-list: null !default;
   .ecl-site-header__login-toggle,
   .ecl-site-header__search-toggle,
   .ecl-site-header__language-selector {
+    align-items: center;
+    display: flex;
     font-size: 0;
+    justify-content: center;
     line-height: 0;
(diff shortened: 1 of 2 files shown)


src/implementations/vanilla/components/site-header/site-header-eu.scss
```diff
@@ -329,14 +329,18 @@ $language-list: null !default;
...
     font-size: 0;
+    justify-content: center;
     line-height: 0;
     padding: calc(var(--s-xs) - 1px) 0;
     text-align: center;
 
     .ecl-icon {
-      width: 1.5rem;
       height: 1.5rem;
+      margin-bottom: 0;
```
planctus: i see that we have a margin-bottom set for the selector: .ecl-site-header__icon and we reset it here, but the question is: when do we need that margin..? It seem we could just remove that instead of adding another override, which i think it's not even the only one since we have something like: 
.ecl-site-header__language-icon, .ecl-site-header__login-toggle {
    margin-bottom: 0;
} in the l viewport, maybe we can clean up a bit these styles if i haven't missed something

planctus (approved):

PR #3603, opened 2024-09-04
feat(video): Adding support for muted, autoplay and loop - Media container adding autoplay video option - FRONT-4600

planctus:
- add parameters to the video component, to toggle different states (muted, loop, autoplay, controls)
- add single toggle in media container to switch between default video, and autoplay one (also having loop, muted and hidden controls)


src/implementations/twig/components/banner/banner.html.twig
```diff
@@ -158,16 +158,13 @@
...
-          { name: 'data-ecl-banner-video' },
-        ]|merge(_video.extra_attributes|default([]))  %}
       {% include '@ecl/video/video.html.twig' with _video|merge({
         controls: false,
+        autoplay: true,
+        muted: true,
+        loop: true,
         extra_classes: video_classes,
-        extra_attributes: video_attributes,
+        extra_attributes: video.attributes|default([])|merge([{ name: 'data-ecl-banner-video' }]),
```
emeryro: shouldn't it be `_video.extra_attributes`?

src/implementations/twig/components/media-container/media-container.story.js
```diff
@@ -10,11 +10,19 @@ import mediaContainer from './media-container.html.twig';
...
     show_description: true,
     description: data.description,
     full_width: false,
   };
+  if (data.video) {
+    args = {
+      ...args,
+      autoplay: false,
+      muted: false,
+      loop: false,
```
emeryro: The idea of COMM here is to have a single control that would toggle the different video options: autoplay, muted, loop and controls.
We should keep the parameter in the video, as you have them, but just expose one control in the media container (either we have a normal video player, or we have an autoplay video like on the banner)

emeryro (changes requested):

emeryro (approved):

PR #3604, opened 2024-09-05
feat(picture): zoom effect - FRONT-4599

emeryro:
Add zoom effect on some interactive pictures:
- card
- content item
- gallery


planctus:
I find this a bit grotesque in the context of the library of components of the European Commission/Union.
Beside this general feeling, what i really struggle with is to understand the reason why of these requirements:

- gallery only if there is an overlay
- card and content item only if the image is linking somewhere

There is a logic, but is that logic anything understandable by the users? I foresee that we will receive bug reports for those cases where the image will not zoom, together with other bug reports about the fact that the image IS zooming, iinstead.
I wonder what is the real need for this animation, maybe they requested it for ONE single website or homepage, now this will change the behavior of the card and the gallery in ALL the websites using ECL.

I think a way to moderate this would be to set this always to false by default, so as an opt-in feature, if they choose to use it they will not complain afterwards, but if they are going to get this mandatorily from the template, well, i would perfectly understand if they do.

emeryro:
I tend to agree with you...
The reason why it is only activated in some situation is because it is only for interactive elements (when you can really click on the image). That's how it was requested.
But indeed having it disabled by default may be safer. Let's check with COMM

src/implementations/twig/components/gallery/gallery.html.twig
```diff
@@ -49,6 +50,7 @@
 {% set _grid = grid|default(false) %}
 {% set _column = column|default(3) %}
 {% set _ratio = ratio|default('3-2') %}
+{% set _picture_zoom = picture_zoom|default(false) %}
```
planctus, line 55: but why do we add a parameter in all these components? Shouldn't it be just an additional key in the picture component, and then you just pass it along its object?

emeryro: I would say to make EWPP's life easier. That's kind of the same logic as for the autoplay video: it could have been just playing with the video parameters, but we still bundled it at the component level.
Also for the gallery the zoom has to be added to pictures and video, and for some reasons the data structure of both is a little different

planctus: well, ok, just a question: like this in the gallery you will have either all the items animated or none of them, is this the expected behavior? For how i see this we are adding a property to video and picture, it should simply be possible to activate or deactivate this for the single media in the data provided, there is no need to involve the parent templates, but fine.

emeryro: I think that it is expected to act on all images at once. We will confirm in today's meeting

src/implementations/vanilla/components/picture/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@ecl/vanilla-component-picture",
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "4.6.3",
```
planctus: let's update the versions of these packages before mergin this

PR #3610, opened 2024-09-06
fix(mega-menu): Showing the panels only when ready - FRONT-4604

planctus:


emeryro:
Not fully sure about this... The result is indeed cleaner, but now we are introducing a small delay every time we open a mega menu item. 
On the other hand that would be tricky to do it differently, as we need to have the component "displayed" to do the calculus...
I did some test by setting height:100% on `.ecl-mega-menu__mega` (instead of calculating it with js. It seems to improve things a little (without your js update), but there are some other side effects when the mega item is smaller by default.

planctus:
But to me it's just a matter of "getting used" to this new "smoothness", we still had to "wait" a bit to get everything in the right place, so we just avoiding showing the items while still adjusting them. We are not used to do this, but recently we applied the same approach to the carousel and also there it feels like an improvement, no need to render immediately something that then might have things re-adjusted in it.

PR #3611, opened 2024-09-06
fix(form): Fixing styles for the label - FRONT-4605

planctus:


emeryro (approved):
The only drawback I see is that the star (or required) may end up alone on a new line. 
But that's already an edge case
image

PR #3617, opened 2024-09-10
docs(menu): remove single menu showcase - FRONT-4494

emeryro:
Use site header to showcase the menu


planctus (commented):
This is only changing the website demos using the whole site header, but why do we keep the standalone demos in storybook, instead?
To me that is the most critical thing, developers are confused since they refer to them while they should refer to the site header demos.

emeryro:
I planned to remove the stories too, seems that I forgot :)

planctus (approved):

PR #3624, opened 2024-09-11
fix(inpage): Adding scroll listener and recalculate height when scrolling - FRONT-4615

planctus:


emeryro:
The fix is working fine, but depending on where the inpage nav is situated at first, we may still have a "broken" display until the user scrolls or resize.
image

Maybe we could add a min height, to mitigate the issue (it will still resize correctly when there is more space)

planctus:
mmh.. introducing a min-height might just break the functionality making items unreachable in those situations. Is that we were not really using all the available space, otherwise it is all good, items can normally be "cut" if there is no room for them. I've added a small delay and checked again the calculation done, which was not reliable without this delay.
Now it should just do what it is supposed to.

emeryro (approved):

PR #3629, opened 2024-09-16
feat(inpage): Using the right element to determine the top position in mobile - FRONT-4620

planctus:


emeryro (commented):
I'm a little confused here. By checking the code, it seems that this `topPosition` is in fact used to calculate the max height (for the scrolling).
So it subtracted 16px on mobile, which correspond to the space that we have at the bottom
image
With the proposed fix, this space is reduced to 12px on EC, and no space at all on EU (there is no padding on the wrapper)
image
I don't recall if this space was requested by the designers

planctus:
i don't recall designers having ever considered the presence of scrollbars. :) The difference here was due to the negative margin set on the wrapper in EC, now we take also that into account.

To explain a bit, before we were using the button to get this "top position", and then we were manually "compensating" this with an harcoded value, now we get the wrapper position and it can be used as such, just that there is this little overhead of the negative margin for EC which introduced this misalignment. Currently there will be no spacing under the dropdown in EC and EU when the scrollbar will appear.

emeryro (approved):

PR #3633, opened 2024-09-17
fix(unordered-list): Setting list style position to outside also in print - FRONT-4602

planctus:
…


emeryro (approved):
This may be seen in another PR, but we have a very different display for list if the "default" css is not there. I guess that we are missing things in the print css itself
image

planctus:
but i'm not really sure about what to do here, the fact that there is a difference with or without the default css i'd say it's normal, the default css duplicates all the rules in the standard css for the ecl selectors, so it's like saying that there is a difference between the main style and the print style.
I have adapted some obvious things, like the markers and checked more or less the consistency of all the variants in unordered and ordered lists. There is less padding in print, a smaller line-height, but that seems to be coherent with the print styles.

emeryro (approved):

PR #3637, opened 2024-09-23
fix(table): fix rtl display on mobile - FRONT-4624

emeryro:
Fix margin on mobile rtl


planctus (commented):
image
It seems that we are "losing" (it's already the case in production) alignment between the headings and the cells in EU, maybe we can take care of this here.

emeryro:
The spacing was wrong on EU anyways (compared to figma).
It's fixed

PR #3642, opened 2024-09-24
feat(page-header): transparent background on EC - FRONT-4361

emeryro:
remove white background on EC page header


yarn.lock
```diff
@@ -2652,85 +2652,85 @@
     estree-walker "^2.0.2"
     picomatch "^2.3.1"
 
-"@rollup/rollup-android-arm-eabi@4.21.3":
```
planctus, line 2655: are you using node 20..? I wonder the reason why of all these changes in yarn.lock

planctus:
i guess we are doing this not because anyone requested it but because the designers wanted to add another potentially breaking change. Am i correct? :)

emeryro:
not really a breaking change this time, they expect to have a slightly different color for page background, so it became visible that some component did not handle that correctly

planctus:
i said potentially breaking because we are making transparent something that was not, who knows if someone has some background there that might not cope well with the text color of the page header, hopefully no one did, but the risk is there.

planctus (approved):

PR #3644, opened 2024-09-24
fix(file-download): fix transparent background - FRONT-4628

emeryro:
fix display issue when the file download is on a colored background


planctus (approved):
Funnily enough, we are fixing here something that we might be breaking on the page-header :)

PR #3647, opened 2024-09-25
fix(menu): fix touch devices - FRONT-4632

emeryro:
fix touch event not working correctly for simple menu


emeryro:
This was tricky...
The issue is that the mouse hover event is still triggered with touch devices. So in some cases we had the hover and click event triggered together.
I added a bool to make sure that when using a touch device, the hover event is not triggered

Note: if you try switching between touch devices and mouse back and forth, there may be some cases where the hover event is still disabled until you click on the caret.
I choose to ignore this scenario as it does not seem to be a real use case

planctus (approved):

PR #3652, opened 2024-09-27
chore(test): Migrating from jest to vitest

planctus:


planctus:
The result of this test is that simply replacing jest with vitest with the current tests we perform is not convenient, tests are running faster on just than vitest, indeed. It might still be the case in the future to reonsider vitest with a wider scope, considering the integration with storybook, but at the moment we have no reasons to migrate.

PR #3653, opened 2024-09-30
fix(tabs): fix icon position on mobile - FRONT-4321

emeryro:
Fix alignment of the icon for previous/next on mobile


planctus:
This is very tricky to test, the demo of the tabs seems to have similar issues to the ones we had with the breadcrumb or other js driven components for which we had to delay a bit the processing to get the right behavior.

planctus (approved):

PR #3660, opened 2024-10-02
fix(mega-menu): Setting opacity: 1 on the container when opened - FRONT-4641

planctus:


emeryro:
Fix works fine.
I guess that we didn't see that issue because the container is empty. Maybe we should add some dummy content in it...

emeryro (approved):

emeryro:
Thanks
I just noticed that there is a "flickering" happening when we open the container item
2024-10-09-menu

emeryro (approved):

PR #3671, opened 2024-10-10
feat(print): ECL4 improve print for fact pages - FRONT-4648

emeryro:


emeryro:
Changes have been introduced in ECL 5

PR #3672, opened 2024-10-14
feat(featured-item): Fixing RTL display, fixing alignment - FRONT-4652

planctus:
Didn't really expect to have to kind of refactor the component, but in order to fix the RTL display many of our shorthands properties had to be converted to inline and block margins and paddings.
Furthermore the specs are not so precise, the alignment has been double and triple checked, it seems as close as possible to what we have in figma in the different variants (not all the different layouts are really specified).


emeryro (changes requested):
It seems to be fine on almost all situation, except for EU simple right aligned; we are missing some spacing here
image

emeryro (approved):

PR #3697, opened 2024-10-28
feat(featured-item): Styles adapted to support no title use case - FRONT-4669

planctus:
…ONT-4669


emeryro:
If for some weird reasons we have no title and no description, there is the same issue with the link
image
Won't it be better to say that the first element in `ecl-featured-item__item` doesn't get a margin top?

planctus:
yes, it would be probably smarter to deal with the first item but i didn't want to go into a refactoring of it, considering that this "issue" is in place since when we introduced this component, i did not even know whether we had to support this use case, actually.
Now we are also supporting the case of a lonely link but i haven't changed the approach, so the relative styles are targeting that selector specifically in relation to the other selectors in the component.

emeryro (approved):

PR #3699, opened 2024-10-28
feat(forms): add option to hide label - FRONT-4668

emeryro:
Add an option to keep the label in the code (for accessibility reasons), but hide it visually.
Label should always be provided


src/implementations/twig/components/form-group/form-group.html.twig
```diff
@@ -178,6 +180,7 @@
       id="{{ _id }}-label"
     {% endif %}
       class="{{- _label_class -}}
+      {{- _hide_label ? ' ecl-u-sr-only' -}}
```
planctus: mmh..fine eh, but i would stick to the idea of not using directly utilities inside the components markup, otherwise in a way we are also "accepting" this practice by our users.
We can add a selector with a semantic name and extend the ecl-u-sr-only in the component's sass file.

emeryro: indeed. That's how we do it in the buttons already, but I guess that it could be updated there too in the future

emeryro: (not the button, but the rating field)

src/implementations/twig/components/form-group/README.md
```diff
@@ -8,12 +8,12 @@ npm install --save @ecl/twig-component-form-group
 
 ### Parameters:
 
-- **"label"** (string) (default: '') - Label for the form group
```
planctus, line 11: was this removal intentional?

planctus:
As an accessibility improvement to me it's fine, hard to say that it will be used properly though, i feel that if they will not want to see a label they will just not set it, they should make it mandatory for that to happen.
Anyway, the thing is that without a label, we do not show in any way whether a field is required or not, it has always been like this, so maybe we don't have to tackle this now, but it still feels wrong.

src/implementations/twig/components/form-group/README.md
emeryro: yes, the label is listed in the blocks; that was the case already in the twig file, so I just applied it in the readme

PR #3711, opened 2024-11-11
fix(default-css): Fixing rules for links inside lists, fixing description list - FRONT-4676

planctus:


emeryro:
Just to double check here: the description list has an underlined link. This is is the only case in our examples where the link is underlined in a list. Is it on purpose?
image

src/implementations/twig/utilities/html-tag/index.story.js
```diff
@@ -54,7 +54,7 @@ export const Default = () => `
         <li>Ordered list</li>
         <li>Ordered list
           <ol>
-            <li>Nested ordered list</li>
+            <li><a class="ecl-link">Nested ordered list</a></li>
```
emeryro:
```suggestion
            <li><a class="ecl-link" href="#">Nested ordered list</a></li>
```

planctus:
mmh..better asking the designers, this was orginated by a request related to unordered and ordered lists, the description list was not touched, i only fixed some styles in the default css

PR #3712, opened 2024-11-11
feat(navigation-list): Adding image-as-illustration variant - FRONT-4678

planctus:


emeryro (approved):
Fine for me. We could just add an example on the website

PR #3713, opened 2024-11-12
chore(v5): POC of a single package per component (accordion)

planctus:


src/components/accordion/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@ecl/accordion",
```
emeryro, line 2: for the package name (removing "component"), is it to avoid confusion with existing packages?
My concern is that we may have some conflict later between component and other ECL packages. For instance, if we have to add a component named "builder" or "event-manager" (not sure why, but who knows...).

emeryro (commented):
I like this new structure! 
We could refine it a little to make it even clearer for users. For instance, story and test files are for us only, so we could differentiate what really concerns ECL users or not.

src/components/accordion/package.json
planctus: the story and test files are ignored by npm, so they are not in the distributed package, as it is already in v3, i wouldn't create "folders" to nest those files in...

About the naming, it seems straighftorward now that we would unify everything in a single package to just use the component's name, we can add any prefix we want, like "component-accordion" but what is that for..?

Currently, for instance, the builder package is already simply @ecl/builder, i hardly see how in the future this might conflict with other packages that we distribute as ECL.

PR #3736, opened 2024-11-26
feat(icon): Webtools icons - FRONT-4688

emeryro:
Add possibility to use Webtools icons instead of the ones provided directly by ECL

Main ideas:
- icon template has a new parameter `wt_markup`. Setting it to true would render the webtools markup, later replaced by the icon via Webtools javascript
- components using icons have a new parameter in twig template: `icon_wt_markup`. This had to be done at component level, as some icon are handled directly here, and do not expose the full icon structure


src/implementations/twig/components/accordion/README.md
```diff
@@ -15,6 +15,7 @@ npm install --save @ecl/twig-component-accordion
     - "label" (string) (default: '')
   - "content" (string) (default: '')
 - **"icon"** (array) OR (object) (default: []) Two icons in an array that will be toggled
+- **"icon_wt_markup"** (boolean) (default: false): should the icon use the Webtools markup?
```
planctus, line 18: it's a boolean but it mentions markup in its name, feels confusing, anyway do we need it at all? I understand that in certain components we might be defining the icons ourselves, and there we need somehow to add a param to handle this, but here and in many other components we could simply use the param as part of the icon object passed. Also, it might not really be a real use case, but if i want to use one icon only from webtools in components that have multiple it would not be possible this way.

src/implementations/twig/components/icon/icon.html.twig
```diff
@@ -65,7 +72,7 @@
 {% endif %}
 
 {% for property in ['size', 'transform', 'color'] %}
-  {% if _icon[property] is defined and _icon[property] is not empty %}
+  {% if _icon[property] is defined and _icon[property] is not empty and _icon[property] != 'none' %}
```
planctus, line 75: is this additional check needed because of the storybook demo..?

src/implementations/twig/components/icon/icon.html.twig
```diff
@@ -96,30 +112,50 @@
...
-  {{ _extra_attributes|raw }}>
-{% if _extra_accessibility.title is not empty %}
-  <title
-  {% if _extra_accessibility.title_id is not empty %}
-    id="{{ _extra_accessibility.title_id }}"
+{% if _icon.wt_markup %}
+  {% set _wt_classes = '' %}
+  {% set _full_name = '' %}
+  {% if _icon.category is not empty %}
+    {% set _full_name = _full_name ~ '-' ~ _icon.category %}
```
planctus, line 119: this is weird, _full_name is an empty string, guess that it can simply get _icon_category as its value when its present.

src/implementations/twig/components/icon/icon.story.js
```diff
@@ -19,25 +19,36 @@ const getArgs = (data) => ({
   color: 'default',
   transform: 'none',
   title: '',
+  description: '',
```
planctus, line 22: not sure about the whole plan regarding this, but shouldn't we expose a control to demo this?

src/implementations/twig/components/link/link.html.twig
```diff
@@ -77,10 +79,11 @@
   {% set _link = _link|merge(link) %}
 {% endif %}
 
-{% if _link.external and _link.icon_path is not empty %}
+{% if _link.external and (_link.icon_path is not empty or _link.icon_wt_markup) %}
```
planctus, line 82: this is one of those cases where i don't understand why adding the parameter at the component level, we merge the data provided by the user, why do we need to add this complexity and additional processing with also the consequence of forcing all the icons to follow the same approach?

planctus (commented):
I think we don't need to expose a parameter in each component using icons, it's a lot of code and logic to handle something that should belong to the icon, not to the component implementing it. Also, if the intention is to always use the other icons, then it would be just a matter of toggling the value of the parameter in the icon template and all these additional parameters would become obsolete and useless.

emeryro:
I'll check the different points, but to address your main concern: 
Initially I tried to put the parameter only in the icon, and in a few components where it was needed, but even on my side it became hard to know where to put the boolean in the data. So I decided to expose it at component level every time.
I agree that it may be a little too much, but at least this is consistent (and would probably save some back and forth with EWPP, or the need for a dedicated documentation)

PR #3738, opened 2024-11-28
chore(pnpm): Replacing yarn with pnpm, completed migration - FRONT-4693

planctus:
This entirely replaces yarn with pnpm,
Install pnpm globally in your machine and execute pnpm install instead of yarn, the rest should work transparently as it used to do in v4, just use pnpm instead of yarn anytime you want to execute some specific command or script.


emeryro:
I tried to do a clean build, removing all node_modules and yarn.lock, but it ended with an error. Not sure if I miss something on my local. (on v5-dev it works fine)
image

planctus:
i think it's better having two different clones of the repo to work independently on v4 and v5, otherwise everytime switching between different branches would need some manual operation like removing the node modules folders.
Anyway, did you install pnpm on your machine? I forgot to add the command to be executed now for the installation which is pnpm install, patch-package should then find the libraries to be patched where it expects them to be.

emeryro:
I have a different clone for ecl 5, but I wanted to test with a fresh install.
Anyways, it works fine now for the website, but when trying to run only storybook I have an error and the local does not load
image

emeryro:
solved by cleaning up: removed all node_modules and `pnpm store prune`

emeryro (approved):

PR #3740, opened 2024-12-03
feat(typography): font tokens and utilities - FRONT-4696

emeryro:
Update all typography tokens and default font family.
See migration guide and documentation for more information


planctus:
Took a look at the result in each component, nothing completely broken, but the date block looks like this: 
image
being this the only really problematic thing of this global change maybe we can tackle it directly here in this pr

src/utilities/typography/typography.scss
```diff
@@ -12,6 +12,17 @@ $typography: null !default;
 /*
 * Headings
 */
+.ecl-u-type-display {
```
planctus: mmh..not sure how to handle this situation, things that are only defined in one system and not in the other, there are a few use cases now about the typography, for instance we have all the "compact" and "extended" utilities in EU although there is no "display" whatsoever in EU, but still it sets a font-size and ensures that if a utility is used both in EC and EU website it will be defined in both, at the same time it's not really doing what the user might expect, so it's kind of questionable.

Here is a bit worse, we get things like: font: !important; harmless maybe but not nice, still it would be also questionable to remove them from the EU css.

My personal preference would be to only define those utilities that are really applying styles based on their definition, if those styles are not defined in one of the two systems no utility would be generated.

planctus (commented):
Left a couple of comments, overall the change is obviously very noticeable and as questionable as all the other things happening in this library but nothing really breaks badly.

planctus (approved):

PR #3753, opened 2024-12-19
feat(website): Showing custom properties hovering on color name in website showcase - FRONT-4703

planctus:


emeryro (commented):
it may be easier to use if the css property was displayed when hovering the full color element (li), and not only the title

emeryro (approved):

PR #3762, opened 2025-01-07
chore(docs): Replacing yarn with pnpm in the docs - FRONT-4708

planctus:


tibi2303 (approved):
Run every command and it's working properly.

PR #3766, opened 2025-01-10
feat(v5): Merging components packages - A to B - FRONT-4713 

planctus:


tibi2303 (approved):
I will approve, still we need to keep in mind, after release to change also in here the paths **src/specs/pages/**

PR #3767, opened 2025-01-13
feat(color): add color modes - FRONT-4710

emeryro:
First implementation of the color modes structure:
- color palettes
- color modes css generation
- example with tag component


src/implementations/twig/components/tag/tag-set.html.twig
```diff
@@ -44,6 +46,7 @@
       <li class="ecl-tag-set__item">
         {% include '@ecl/tag/tag.html.twig' with _item|merge({
           icon_path: _item.icon_path|default(_icon_path),
+          color_mode: _color_mode,
```
planctus: mmh...don't you foresee the fantasy of an editor introducing a tag list with each tag using a different color mode..? :)

src/implementations/twig/components/tag/tag-set.story.js
```diff
@@ -50,6 +51,8 @@ const prepareData = (data, args) => {
     item.tag.external = args.external;
   });
 
+  data.color_mode = args.color_mode;
```
planctus: we do an Object.assign why do we need this..?

src/playground/eu/.storybook/preview.js
```diff
@@ -45,6 +45,12 @@ export const parameters = {
       picked: true,
       hideCode: true,
     },
+    {
```
planctus: why do we add this to EU..? Isn't it an EC thingy only..?

src/playground/ec/.storybook/preview.js
```diff
@@ -45,6 +45,12 @@ export const parameters = {
       picked: true,
       hideCode: true,
     },
+    {
```
planctus, line 48: i don't know, in the end this is going to be a pretty small file, not sure why keeping it separated, maybe because there might be sites not using it at all..? I mean, do we have any advantage by doing this? What i initially thought was that we could have produced one file per color mode, so that the single website would have been only loading the one corresponding to its chosen "priority", but if we are about to ship them all together then i don't necessarily see why keeping it separated, it will be combined by the application in any case..

planctus (commented):
Pretty straightforward, i just don't see why defining everything in EU, didn't we set a default for the values in case no color mode is defined? Then why do we need to even touch EU..? Also about defining a css on its own for the color modes seems unneeded to me, it's a small file and i don't necessarily see when that would not be needed.

src/playground/ec/.storybook/preview.js
emeryro: The main reason why I propose to split it is because of possible request to add/update color mode quickly, if there is some business need.
Having it in its own little css would allow to update the colors first, and take the time to really catch up with the ECL release. I agree that it is an edge case (and somehow hacky), so maybe we should see first if that may be a real case or not

PR #3768, opened 2025-01-13
feat(v5): Merge components packages from D to E - FRONT-4715

planctus:


src/components/expandable/package.json
```diff
@@ -1,19 +1,26 @@
...
   "version": "4.9.0",
   "description": "ECL Expandable",
+  "publishConfig": {
+    "access": "public"
+  },
   "main": "expandable.js",
   "module": "expandable.js",
   "style": "expandable.scss",
-  "sass": "expandable.scss",
+  "scss": "expandable.scss",
```
tibi2303: as we decided, let's remove all of these

PR #3769, opened 2025-01-14
feat(v5): Merge components packages starting with F - FRONT-4716

planctus:


tibi2303 (commented):
image
image


There are still some paths that needs updating.

tibi2303 (approved):

PR #3788, opened 2025-01-20
chore(v5): Merge packages T-V. - FRONT-4722

tibi2303:


src/components/table/package.json
```diff
@@ -1,21 +1,19 @@
...
+    "access": "public"
+  },
   "main": "table.js",
   "module": "table.js",
+  "style": "table.scss",
   "dependencies": {
-    "@ecl/builder": "4.9.0",
     "@ecl/dom-utils": "4.9.0",
-    "@ecl/resources-icons": "4.9.0",
-    "@ecl/vanilla-layout-grid": "4.9.0"
```
planctus, line 15: it seems we removed this dep, it should still be there but now the package is @ecl/grid

src/components/tag/package.json
```diff
@@ -1,19 +1,18 @@
...
   "author": "European Commission",
   "license": "EUPL-1.2",
   "version": "4.9.0",
   "description": "ECL Tag",
   "publishConfig": {
     "access": "public"
   },
+  "main": "tag-print.js",
+  "module": "tag-print.js",
+  "style": "tag-print.scss",
```
planctus: mmh..is not really used anywhere i think, but the "main" css is just tag.scss

src/components/form-group/package.json
```diff
@@ -17,12 +17,12 @@
...
     "@ecl/select": "4.9.0",
-    "@ecl/twig-component-text-area": "4.9.0",
-    "@ecl/twig-component-text-input": "4.9.0"
+    "@ecl/text-area": "4.9.0",
+    "@ecl/text-input": "4.9.0"
   },
   "devDependencies": {
-    "@ecl/specs-component-text-area": "4.9.0",
-    "@ecl/specs-component-text-input": "4.9.0"
+    "@ecl/text-area": "4.9.0",
```
planctus: it seems these are already dependencies, so we don't need to duplicate them as devDependencies

src/components/timeline/package.json
```diff
@@ -1,20 +1,19 @@
...
-  },
   "publishConfig": {
     "access": "public"
   },
+  "main": "timeline.js",
+  "module": "timeline.js",
+  "style": "timeline.scss",
+  "dependencies": {
+    "@ecl/dom-utils": "4.9.0",
+    "@ecl/button": "4.9.0"
```
planctus, line 15: if the vanilla component had @ecl/vanilla-layout-grid as a dependency now we should still have it but as @ecl/grid

src/playground/ec/.storybook/environment.js
```diff
@@ -1,10 +1,7 @@
 const path = require('path');
 const { TwingEnvironment, TwingLoaderFilesystem } = require('twing');
 
-const componentAbsPath = path.resolve(
-  __dirname,
-  '../../../implementations/twig/components',
-);
+const componentAbsPath = path.resolve(__dirname, '../../../components');
```
planctus, line 4: We are duplicating this, it was already there, but named componentNewAbsPath, so please remove the latter also in loader.addPath()

src/playground/eu/.storybook/environment.js
```diff
@@ -1,10 +1,7 @@
 const path = require('path');
 const { TwingEnvironment, TwingLoaderFilesystem } = require('twing');
 
-const componentAbsPath = path.resolve(
-  __dirname,
-  '../../../implementations/twig/components',
-);
+const componentAbsPath = path.resolve(__dirname, '../../../components');
```
planctus, line 4: same as for EC

src/playground/ec/.storybook/environment.js
```diff
@@ -1,7 +1,6 @@
 const path = require('path');
 const { TwingEnvironment, TwingLoaderFilesystem } = require('twing');
 
-const componentAbsPath = path.resolve(__dirname, '../../../components');
```
planctus: mmh..no, doesn't make much sense to have it named componentNewAbsPath, remove that one, instead of componentAbsPath :)

PR #3795, opened 2025-01-22
feat(media container): color modes - FRONT-4738

planctus:


emeryro (changes requested):
As this component is affected by the color modes (even if this is just the border), we may add the storybook control for it

planctus:
Removed the images styles, the component is not changing anymore with the color mode being used, not even indirectly, so no reason why adding a control in storybook

emeryro (approved):

PR #3799, opened 2025-01-23
feat(site-footer): new EC core footer - FRONT-4754

emeryro:


src/components/site-footer/site-footer-ec.scss
```diff
@@ -146,17 +183,38 @@ $site-footer: null !default;
...
+          align-self: flex-end;
+          flex-grow: 1;
+        }
+      }
+    }
+
+    .ecl-site-footer__row:last-child {
+      display: block;
+
+      // stylelint-disable-next-line no-descending-specificity
```
planctus, line 215: not an issue, but this is not needed, right..? I don't see any different "display" property set for the columns, once we set block for the parent..

src/components/site-footer/site-footer.html.twig
```diff
@@ -141,6 +142,9 @@
           {% if _sections.links_inline %}
             {% set _list_class_name = _list_class_name ~ ' ecl-site-footer__list--inline' %}
           {% endif %}
+          {% if _sections.links_separator %}
```
planctus, line 145: brilliant, that's all we needed..? :)

planctus (approved):

src/components/site-footer/site-footer.html.twig
emeryro: for the EC core footer yes. We'll see later for the other ones...
And in any case I still think that it is worth refactoring this component

src/components/site-footer/site-footer-ec.scss
emeryro: this is broken on the EC core footer without this line. That's what make the last section take the full width
image

PR #3803, opened 2025-01-24
feat(navigation-list): Color mode. - FRONT-4731

tibi2303:


emeryro (changes requested):
Just a small thing: the background of the image should change based on the color mode (color surface-lowest), similar to what has been done on https://github.com/ec-europa/europa-component-library/pull/3795/files
This may be visible if the image is partially transparent (or broken)

emeryro:
After clarification, no need to put background color

emeryro (approved):

PR #3805, opened 2025-01-24
feat(content-item): Color mode. - FRONT-4729

tibi2303:


src/themes/ec/variables/_content-item.scss
```diff
@@ -2,8 +2,9 @@
 @use '../index' as *;
 
 $content-item: (
-  divider-color: var(--c-n),
+  divider-color: var(--cm-border-neutral-lowest),
   padding: var(--s-xl),
-  image-border-color: var(--c-n),
+  image-border-color: var(--cm-border-neutral-lowest),
```
planctus: mmh..based on the reply by the designers this border color should follow the color mode, border-neutral doesn't, guess that it should be --cm-border-low

planctus (changes requested):
The border color of the image needs to follow the color mode being used but we also need to add support for the color modes in the template and a control in storybook.

src/themes/ec/variables/_content-item.scss
tibi2303: thanks, also removed the background for the image.

planctus (approved):

PR #3809, opened 2025-01-27
feat(site header, menu): color mode - FRONT-4755

emeryro:


src/components/site-header/site-header-print.scss
```diff
@@ -29,11 +30,11 @@ $theme: null !default;
 }
 
 .ecl-site-header__banner-top {
-  color: var(--c-d);
+  color: var(--cm-on-surface-dark);
```
planctus: mmh..but this file is common between EC and EU, so in EU this color will be undefined...now, it's the class name and i don't even know if we should keep "supporting" it since we never show it on our side, but anyway..
While doing this test i also noticed that with the logo centered, when there is a notification in place in the site header, it looks really bad because the notification is limited in its width: 
image

planctus (approved):

PR #3812, opened 2025-01-27
feat(tabs): Tabs extra classes on items. - FRONT-4758

tibi2303:


src/components/tabs/tabs.html.twig
```diff
@@ -83,6 +84,9 @@
           {% else %}
             {% set _item_extra_attributes = _item_extra_attributes|merge([{ name: 'aria-selected', value: 'false' }]) %}
           {% endif %}
+          {% if _item.extra_classes is defined %}
```
planctus: i think that "is not empty" might be enough, in this template, one after the other we use three different approaches:

- {% if _item.is_current is defined and _item.is_current is not empty %}
- {% if _item.extra_classes is defined %}
- {% if _menu is not empty %}

maybe we should make a decision.. :)
As said i don't think we would have issue using "is not empty" only, but the correct way is probably still to check for both the existence and the emptiness of a property in the user data, can you apply this consistently in this template? We might then want to make all the components consistently using this approach.

planctus (commented):
It's good, let's just make the code a bit more consistent.

src/components/tabs/tabs.html.twig
tibi2303: i improved the conditions to have both is defined and is not empty.. for the 3: menu, extra_classes and is current (this last already had it)

PR #3828, opened 2025-02-03
feat(range): Fixed duplicated aria-describedby. - FRONT-4725

tibi2303:


emeryro (changes requested):
The modification applied here would have side effects: the aria-describedby which is removed is supposed to convey information about the helper and invalid text. This is passed as extra_attributes.
This is the case here for instance, in the text input
image

The range component has an extra aria-describedby for the value. The idea would be to merge these two aria-describedby into one (it can get multiple ids)

tibi2303:
@emeryro thanks romain, i pushed now a commit.
I know that the snaps for form-group are changed, but in my opinion that should be the expected behaviour..

So the form-group is sending that attribute describedby and it's being added, no longer hardcoded in the template. Now if this attribute comes through extra_attributes, it's being added, if not, it's keeping the hardcoded one.

emeryro:
I think this is going in the right direction, but the aria-describedby coming from the range component itself should be added to the ones coming from the form-group (and not just be a fallback)
So, taking the example, we are getting potentially `example-range-id-helper` and/or `example-range-id-invalid` from form-group, and we add `example-range-id-value` to it, to have something like `aria-describedby="example-range-id-helper example-range-id-invalid example-range-id-value"`

tibi2303:
@emeryro thanks, romain. Indeed, now i pushed the fix. It adds to the value, not replacing

emeryro (approved):

PR #3829, opened 2025-02-03
feat(icons): Changed vimeo icon. - FRONT-4726

tibi2303:


emeryro:
There has been a misunderstanding: the icon to be replaced is vimeo-color. The goal is to get rid of the black "v" in it.
The default vimeo icon should still be dark
image

emeryro (approved):

PR #3837, opened 2025-02-05
feat(popover): ECL Popover v3 - FRONT-4769

planctus:
What's in this pr..?

In the current implementation the popover is an element that is positioned choosing the biggest available space in the viewport and can extend to the edges of the viewport itself, if it goes beyond the boundaries, though, is pushed to the edge in order to be fully visible.

But: the problem is that if there is a scrollable parent, meaning an element handling the overflow, the popover can get clipped with a portion of it no being visible if it goes beyond the boundaries of the scrollable element.
This is what happens in the case that originated this work, where the popovers are placed into a scrollable element holding a table.

The solution is to check for a scrollable element and if it's found using it as the limiting area for the popover display.
To demonstrate the new behavior a story has been added using the markup from the website where the problem was identified, it should be possible to test all the four different positioning and the relative "push" when the popover goes beyond the limits. 
This story will be removed, mind the fact that in storybook the body element is not taking the full available height in the viewport, so it is possible that the lowest placed toggles might "trigger" a push-bottom without a real need for that.

A couple of basic events have been added, onOpen and onClose


src/components/popover/popover.js
```diff
@@ -167,30 +224,121 @@ export class Popover {
...
+      }
+
+      parent = parent.parentElement;
+    }
+
+    return document.body;
+  }
+
+  /**
+   * Find the closest scrollable parent.
```
emeryro:
```suggestion
   * Calculate available space for the popover
```

emeryro (approved):
Just a small copy/paste, but the improvement is fine for me

PR #3841, opened 2025-02-06
fix(modal): Center alignment in the footer buttons - FRONT-4772

planctus:


emeryro:
When there is an icon, or when one of the label is longer, the buttons no longer have the same size.
Maybe we could add a flex-basis: 50% too, while we are at it

planctus:
Yep, i also added a center aligment for the elements inside the buttons so that if it the text goes on two lines for one button only the other has the text centered and not aligned to the top

emeryro (approved):

PR #3847, opened 2025-02-11
feat(featured-item): Removed margin top for featured item description. - FRONT-4790

tibi2303:


src/components/featured-item/featured-item.scss
```diff
@@ -139,7 +139,7 @@ $featured-item: null !default;
     }
 
     .ecl-featured-item__description {
-      margin: map.get($featured-item, 'description-spacing-right');
+      margin-left: 0;
```
planctus: We support RTL display for the components, so we don't use left right top and bottom for spacings like paddings and margins, so this should be margin-inline-start.
To test it, you have a checkbox in the Css resources panel that enables the RTL css

planctus (changes requested):

src/components/featured-item/featured-item.scss
tibi2303: thanks, the margins were wrong also on rtl a lot... i compared with what we have on v4 prod and it's the same now

planctus:
I see this in the "default" featured item in EC with the print css: 
image

I was also thinking that they will expect now the color modes to affect the print as well, i don't see any background currently in the "highlighted" for print, which is a regression, but we should also add support for the color modes in print.

Having said this we are supposed to replace this implementation of the featured item next week, so not sure whether this makes really sense, anyway let's fix it for the current version.

tibi2303:
@planctus thanks. I did not think about highlight and simple but i resolved both of them. I am not ok for using stylelint disable for 2 rules on 2 lines but i could not make the order properly. I checked all of the cases - eu/ec simple default and highlight - normal and print / rtl ( i hope i didnt missed any of them) and should be okay, in most places also i changed using margin-block approach for rtl. Also for print i changed the background to use color modes vars in highlight.

src/components/featured-item/featured-item.scss
```diff
@@ -492,3 +493,17 @@ $featured-item: null !default;
     max-width: 210px;
   }
 }
+
+/* stylelint-disable-next-line selector-not-notation */
```
planctus, line 498: this is the exclusion you referred to in the comment..? stylelint is complaining because of the two :not, merge them and you will not need to exclude any rule. So, like: .ecl-featured-item:not(.ecl-featured-item--highlight, .ecl-featured-item--simple)

src/components/featured-item/featured-item-print.scss
```diff
@@ -31,13 +31,22 @@ $featured-item-print: null !default;
 }
 
 .ecl-featured-item__title {
-  margin: map.get($featured-item-print, 'title-spacing-desktop');
+  margin-bottom: map.get($featured-item-print, 'title-spacing-desktop-bottom');
```
planctus, line 34: it feels like this went a bit beyond the initial expectations, i know that this css was the results of dozens of iterations which messed it up, but now we get lot of changes in this pr which was supposed to be a one line fix. What i would do, at this point, is to use margin-block-start and margin-block-end instead of the old margin top and bottom, it has not a direct impact like the inline margins but it's how it should be.
To save some writing, though, when you define both the inline or the block spacings, you can use a single line shorthand, like here it would be margin-block: map.get($featured-item-print, 'title-spacing-desktop-top') map.get($featured-item-print, 'title-spacing-desktop-bottom');

planctus (changes requested):
As said, we will have a new featured item and probably it will be better to rewrite the styles from scratch, so this effort might be soon become just an exercise, since you aligned the component using margin-inline let's do the same for margin-block, then.

PR #3848, opened 2025-02-13
feat(site-header): Changes in the logo for the site header in EC - FRONT-4791

planctus:


src/components/site-header/demo/data--eu.js
```diff
@@ -4,6 +4,7 @@ const exampleLink = `${publicUrl}/example`;
 module.exports = {
   logo: {
     alt: 'European Union',
+    breakpoint: 'l',
```
emeryro, line 7: we could check with the designers if the value has to stay different in EU

emeryro (approved):

PR #3850, opened 2025-02-13
feat(utilities): add color mode utilities - FRONT-4783

emeryro:
Add color mode utilities for background, border and typography

notes: 
- utilities for old colors will be removed for EC, reducing the size of the css file
- we are providing border utilities for each breakpoints. I kept it as is, but I'm not sure that it is really needed


planctus (approved):
I agree that responsive border utilities are a bit too much, we can reduce a bit the size of this css removing them.

PR #3851, opened 2025-02-14
feat(mega-menu): Link with images in featured panel - FRONT-4793

planctus:


src/components/mega-menu/mega-menu.html.twig
```diff
@@ -27,7 +27,7 @@
       "featured" (oject) (optional) {
         - "title": (string) Title of the featured panel
         - "content": (string) Top content of the featured panel
-        - "items": (associative array) (optional) Object of type ECL link
+        - "items": (associative array) (optional) Array of items with image and link
```
emeryro, line 30: this should be updated in the readme file too

emeryro (commented):
On "tablet" display, the menu is no longer taking 100% width.
I don't really see that as an issue, but that may be reported by the designers
image

planctus:
It has always been the case, this screenshot is taken from v4
image

emeryro (approved):

emeryro:
Could you add a few lines in the migration doc to highlight the changes here (mostly the data)?
Then we can merge it

PR #3854, opened 2025-02-17
chore(grid): Adding XXL breakpoint at 1368px - FRONT-4794

planctus:


emeryro (approved):
This is fine, but we should review the utilities using breakpoints, as the size is starting to explode

emeryro:
If this new breakpoint is validated, we should also update this page: https://v5-dev--europa-component-library.netlify.app/ec/utilities/layout/grid/usage/

PR #3855, opened 2025-02-17
feat(typography): Update components A to B - FRONT-4797

emeryro:
Updates:
- banner
- blockquote


src/components/banner/__snapshots__/banner.test.js.snap
```diff
@@ -51,7 +51,7 @@ exports[`Banner Image renders correctly 1`] = `
                 class="ecl-link ecl-banner__title-link"
                 href="/example"
               >
-                Lorem ipsum dolor sit amet consectetuer adipiscin
+                Lorem ipsum
```
planctus, line 54: i did not underdtand at the beginning why changing the data, then i realized the reason why, it's all pretty silly, we will surely have to iterate on this multiple times in the near future..

emeryro, line 54: yes, this is really huge...

planctus (approved):

PR #3861, opened 2025-02-20
feat(banner): Supporting display without media - FRONT-4808

planctus:


emeryro (commented):
We have a parameter "font_color", which may be quite useful in this situation (constrast is quite low for some color currently). But the control is not available with plain background.
I see that text color is forced to white here. Was it a design request?

planctus:
in figma the token "text-inverted" is used, i thought this maps to the cm-on-surface-inverted on our side, is that wrong?
I wouldn't leave the freedom to the user to make mistakes, we know that it would happen and that we are the ones who would be blamed, so if there are some of the color modes that should instead have a dark text i would expect the on-surface-inverted to be dark in those cases, still not leaving to the user the "responsibility" to choose the color.

emeryro (commented):
This is fine, but should we keep the other parameter marked as deprecated ("centered")?

emeryro (approved):

PR #3868, opened 2025-02-24
chore(fonts): Adding inter font in the ec preset package - FRONT-4814

planctus:
What's in this pr?

-  the file chosen are the variable ones, since with only two files we get everything that we were getting with the google import, weights from 100 to 900, all the subsets, one file is the regular font, the other the italic one.
- These two files are only available in the github release, the fountsource npm package doesn't have them and there is no npm package distributing the files in the same way they are released on github. 
-  a bash script to download the latest release of the inter font from github and to extract the two needed files
-  a bash script to check whether a new version of inter is available
- the two files are added in a fonts folder to the ec preset package in order to distribute them
- in the ec.css the font is loaded as ../fonts/Inter..., this means that the package is already providing everything in the right place, the styles folder and the fonts folder can be simply copied and moved where needed
- for our convenience a preloading link has been added in storybook EC to reduce as much as possible FOAT, this solution can be implemented also by ewpp


emeryro (approved):
great work!

PR #3870, opened 2025-02-25
feat(spacing): Changed demo pages spacing. - FRONT-4809

tibi2303:


planctus (commented):
I don't know, the request was related to paragraphs and sections, we decided to apply this to the headings, but i'm not really convinced, it's true that we were already setting vertical margins for headings and regardless of the heading type, here we are just increasing them, but it feels wrong to me and this would have a relevant impact on sites using the utilities and the default css.
Let's see together with Romain as well, in any case why is the PR titled "horizontal spacing", we are acting on the vertical spacing instead.

tibi2303:
@planctus i only changed now the demo pages spacing, as we talked in the meeting.

planctus (approved):

PR #3872, opened 2025-02-25
feat(list): Use inline links in ordered and unordered lists. - FRONT-4787

tibi2303:


planctus (changes requested):
We have the same also in the unordered list styles, @tibi2303  ;) Please fix also the pr title, these titles will be in the changelog and here what we are doing is to "use inline links in ordered and unordered lists"

tibi2303:
@planctus thanks planctus, indeed, i cleaned up for unordered too and changed the title.

planctus (approved):

PR #3877, opened 2025-02-26
feat(typography): Update components T to V. - FRONT-4806

tibi2303:


src/themes/ec/variables/_tag.scss
```diff
@@ -2,6 +2,7 @@
 @use '../index' as *;
 
 $tag: (
+  font: var(--f-ui-s, var(--f-s)),
```
planctus: mmh..this doesn't make much sense, this was defined in the component scss so that if --f-ui-s was defined it would have been used otherwise the fallback" would have been used instead.
It was a not super nice way to set two different value in the two systems, because EC doesn't have the ui font, so it would have used the other.
If you move this in the system variables, then there is no need to use a fallback, in this case in EC is --f-s and in EU is --f-ui-s

src/themes/ec/variables/_text-area.scss
```diff
@@ -2,6 +2,7 @@
 @use '../index' as *;
 
 $textarea: (
+  font: var(--f-ui-m, var(--f-m)),
```
planctus: same as for the tag

src/themes/ec/variables/_text-input.scss
```diff
@@ -2,6 +2,7 @@
 @use '../index' as *;
 
 $text-input: (
+  font: var(--f-ui-m, var(--f-m)),
```
planctus: You got it.. :)

planctus (changes requested):
There was a misunderstanding about the usage of the var fallback in the components scss. I'm not against moving this in the system's variable, but then no need to use a fallback anymore

src/components/unordered-list/unordered-list.scss
```diff
@@ -12,7 +12,7 @@ $list: null !default;
 .ecl-unordered-list,
 %ecl-unordered-list {
   color: map.get($list, 'text-color');
-  font: var(--f-m);
+  font: map.get($list, 'unordered', 'font');
```
planctus, line 15: this was not really needed, when we don't have differences in the two systems there is no need to parametrize values. Fine, anyway

src/themes/ec/variables/_tag.scss
```diff
@@ -2,6 +2,7 @@
 @use '../index' as *;
 
 $tag: (
+  font: var(--f-ui-s),
```
planctus: Did you check the result of this..? Because there is no font ui in EC. This is the value for EU, instead.

src/themes/ec/variables/_text-area.scss
```diff
@@ -2,6 +2,7 @@
 @use '../index' as *;
 
 $textarea: (
+  font: var(--f-ui-m),
```
planctus: i see, you did it everywhere..so --f-ui variables are only  existing in EU

src/themes/ec/variables/_text-input.scss
```diff
@@ -2,6 +2,7 @@
 @use '../index' as *;
 
 $text-input: (
+  font: var(--f-ui-m),
```
planctus: same

src/themes/eu/variables/_text-area.scss
```diff
@@ -2,6 +2,7 @@
 @use '../index' as *;
 
 $textarea: (
+  font: var(--f-ui-m),
```
planctus, line 5: i thought you had inverted the values between EC and EU, instead you set --f-ui in both and with the same value, all of this was not needed at all, but since you did it, please ensure that you are not altering anything that is not supposed to be updated.

planctus (changes requested):
..

src/themes/ec/variables/_table.scss
```diff
@@ -2,32 +2,49 @@
...
+    background: var(--c-n),
+    border: 1px solid var(--c-n-60),
+  ),
+  header: (
+    border: none,
+    separator-mobile: 2px solid var(--c-n-160),
+    separator-desktop: 1px solid var(--c-n-160),
+    group-separator: 1px solid var(--c-n-160),
+    padding: var(--s-m),
+    font-weight: map.get($font-weight, 'bold'),
```
planctus: We have 600 in figma which is semi-bold

src/themes/ec/variables/_table.scss
```diff
@@ -2,32 +2,49 @@
...
+  ),
+  sort: (
+    icon-color: var(--c-d),
+    icon-color-active: var(--c-d-60),
+    icon-spacing: -8px,
+    arrow-top: calc(var(--s-m) + 3px),
+    arrow-left: var(--s-m),
+  ),
+  caption: (
+    font: var(--f-s),
```
planctus: i see xs in figma...

PR #3889, opened 2025-03-05
feat(documentation): Fixed documentation parameters. - FRONT-4826

tibi2303:


src/components/blockquote/README.md
```diff
@@ -14,7 +14,6 @@ npm install --save @ecl/blockquote
 - **"lang"** (string) (default: 'en')
 - **"extra_classes"** (optional) (string) (default: '') Extra classes (space separated)
 - **"picture"** (associative array) (default: {}): Image for the blockquote, following ECL Picture structure
-- **"extra_classes"** (optional) (string) (default: '') Extra classes (space separated)
```
planctus, line 17: isn't it better to remove the other..? :) That way we would have extra_classes and extra_attributes one after the other as usual

src/components/featured-item/featured-item.html.twig
```diff
@@ -8,9 +8,6 @@
   - "link" (object) (default: {}): object of type Link
   - "media_container" (object) (default: {}): object of type Media container
   - "position" (object) (default: 'left'): Media position
-  - "footer_description" (string) (default: ''): short description in the footer
```
planctus, line 11: but those are still there, you only removed this docs from the template but the parameters are still in place and used, then i guess that in another pr all of this will be cleaned up, this is a bit inconsistent here, here, imagine if we release this and not the pull request where the parameters will be effectively removed.

src/components/file/README.md
```diff
@@ -10,7 +10,7 @@ npm install --save @ecl/file
 
 - **"id"** (string) (default: random string)
 - **"icon"** (object) (default: {}): object of type Icon; file type
-- **"variant"** (string) (default: '') Thumbnail
+- **"variant"** (string) (default: '') Name of the variant [default, thumbnail]
```
planctus: why mentioning "default" as a variant..? default is no variant, basically

src/components/media-container/README.md
```diff
@@ -41,7 +41,7 @@ npm install --save @ecl/media-container
   - "src_lang" (string) (default: ''),
   - "label" (string) (default: ''),
   - "description" (string) (default: ''),
-- **"image"** (string) (default: ''): Image to be used as the video placeholder
+- **"image"** (string) (default: ''): Image to be used as the video placeholder. It serves as a fallback if "picture.img" is empty.
```
planctus: it's not really that we use it as a fallback for the image we expect, is that this was the way to support legacy data, for this to be backward compatible, we can still mention how it is used as a fallback but that is the "image to be used as the video placeholder".  anyway, if it feels too confusing, don't be bothered by this, skip it

src/components/search-form/search-form.html.twig
```diff
@@ -2,8 +2,9 @@
 
 {#
   Parameters:
-    - "text_input" (associative array) default: A predefined structure for EC Text Input
-    - "button" (associative array) default: A predefined structure for EC Button
+    - "label" (string) (default: '') Label of the search form
+    - "text_input" (associative array) (default: A predefined structure for EC Text Input)
```
planctus: ECL maybe, why EC..? i know it was like this, but we have the chance to fix this.

planctus (commented):
Thank you, really useful, also to see how much randomness we could introduce in these templates during time. A few comments, but overall it seems a consistent work on all the templates.

src/components/featured-item/featured-item.html.twig
tibi2303: those are already in deprecated part below, that's why i though they are reduntant

src/components/file/README.md
tibi2303: actually, in the template the docs are like this:   - "variant" (string) (default: 'default') Name of the variant [default, thumbnail] which is fair, that's happening in template. I updated it here

src/components/search-form/search-form.html.twig
tibi2303: yes, i updated it everywhere i could see this.

src/components/media-container/README.md
tibi2303: i removed it

src/components/blockquote/README.md
tibi2303: indeed.

PR #3891, opened 2025-03-06
feat(timeline): Fixed spacing top on timeline items. - FRONT-4833

tibi2303:


emeryro:
the proposed fix would work partially, but the paragraph also add a margin below, which should be removed when possible.
Also, if there are multiple elements in the content area, we should keep the margin between paragraphs, this is why I suggested to target the first and last children
image
And last thing: we should keep the habit to use marin-block-start and margin-block-end instead of margin-top and margin-bottom

emeryro (approved):

PR #3893, opened 2025-03-06
feat(typography): Changed letter spacing and font-weight of headings.- FRONT-4832

tibi2303:


src/themes/eu/variables/_typography.scss
```diff
@@ -8,32 +8,38 @@ $typography: (
   heading1: (
     'mobile': var(--f-4xl),
     'desktop': var(--f-5xl),
-    'font-weight': map.get($font-weight, 'regular'),
+    'font-weight': map.get($font-weight, 'light'),
```
emeryro: there is no "light" font weight in EU, so it should stay regular

emeryro (changes requested):
Just a small thing for EU

Also I noticed that the markup is not correct for the headings previews: the headings are not supposed to be inside the "display" (it should be one tag per class)
https://ecl-preview-3893--europa-component-library.netlify.app/playground/ec/?path=/story/utilities-typography--heading&globals=viewport:responsive
image
We can fix here

src/themes/eu/variables/_typography.scss
tibi2303: thanks, it's rollback and also fixed the tags for headings in story.

emeryro (approved):

PR #3894, opened 2025-03-10
chore(website): Migrate to vike (vite SSR)

planctus:
What's in this PR..?

- webpack replaced by vite
- build made using vike (SSR)
- integration with the CI

Main changes:

- not rendering twig anymore in the website, there is no twing usable in a vite environment, this would involve migrating to twig js
- as a consequence the way to retrieve the html source code of all the stories in the showcase has changed, now it's really coming from the iframe we load, we get the html from there.
- the website is completely ESM, it's not possible to use any of the spec packages, currently

Pros:
- speed, bulding the website takes something like 20 seconds now
- modern,  in the context of this migration react was updated, in the website we were forced to keep it in an old version
- clean, to a certain extent, no patches involved, but read also the CONS
- easier, the configuration of vite compared to the two separated conf we had for webpack is way much easier
- lighter, with webpack we were needing lots of dependencies, we do still need a few plugins but the ecosystems of dependencies related to the website is now much smaller

Cons:
- at the moment building with vike in SSR is removing some important part of the code defined in the website, basically all the code meant to be excuted client side, so to make everything work a couple of scripts have been added to inject scripts in the html pages after the build. The html source in the showcase for instance works this way.


emeryro (commented):
Very nice!

A few things that I noticed so far:
- the font seems to be back to arial (compared to v5-dev)
- when clicking on a link (a component for instance), the left navigation goes back to the top. This may be a little annoying while browsing the website (currently it stays where it is). Not sure if this something that we can fix easily...

planctus:
Thanks @emeryro for taking a look, the reality is that this is working in this pr because we recreate all the needed client side behaviors, typically everything done via js in dedicated scripts that we inject.
Here indeed the hydration doesn't work, or better doesn't simply happen, and therefore many things would not work, like the iframe resizing, or the sidebar toggle, the syntax highlighting, etc.. the behavior you noticed with the sidebar is also due to this.
So, i have just created another pr https://github.com/ec-europa/europa-component-library/pull/3906 where instead the hydration works, so we don't need anymore to hack the generated html files to reproduce all the behaviors mentioned before.
But still there are issues, this hydration should happen automatically with scripts injected by vike, which doesn't happen so again we have to build a bridge to make it work, but it's a step forward, closer to something we could really migrate to.
To be continued then in the other pr, this one will be closed most likely, i created a separated one to have the chance to evaluate both in comparison.

PR #3895, opened 2025-03-10
feat(typography): use font cdn - FRONT-4829

emeryro:


planctus (commented):
We have also this in src/playground/*/.storybook/preview-head.html:
```
<link
  rel="preload"
  href="./fonts/InterVariable.woff2"
  as="font"
  type="font/woff2"
  crossorigin="anonymous"
/>
```

emeryro:
As far as I can see, we don't need that declaration in the preview-head, as the font is already declared in the css, so I removed it

planctus:
fine, it was a "preloading" link, anyway, a way to reduce the risk of FOUT

PR #3904, opened 2025-03-14
feat(site-header): Custom action. - FRONT-4819

tibi2303:


src/components/site-header/_site-header-language-switcher.scss
```diff
@@ -269,47 +300,63 @@ $language-list: null !default;
   // Horizontal sections
   .ecl-site-header__language-content:not(
       .ecl-site-header__language-content--stack
+    ),
+  .ecl-site-header__custom-action-content:not(
```
emeryro: there are things not needed for the custom action in this css file; the language switcher has some specific display in column, but that's not the case here

src/components/site-header/site-header-custom-action.html.twig
```diff
@@ -0,0 +1,101 @@
+{% apply spaceless %}
```
emeryro: do we really need to have a separated twig file for the custom action?
The language switcher comes with a custom structure, so that may be justified, but here I think that it may be in the main template

src/components/site-header/demo/data--ec.js
```diff
@@ -22,6 +22,25 @@ module.exports = {
     label: 'Log out',
     href: exampleLink,
   },
+  custom_action: {
+    label: 'Notifications',
```
emeryro: even if there is an overlay we should have an href defined, otherwise the link is not valid

src/components/site-header/eu-site-header.story.js
```diff
@@ -33,6 +33,7 @@ dataHarmonised.has_menu = true;
 const getArgs = (data) => {
   const defaultArgs = {
     show_language_selector: true,
+    show_custom_action: true,
```
emeryro: custom action should be hidden by default

src/components/site-header/site-header.html.twig
```diff
@@ -49,6 +49,20 @@
               ],
             },
         },
+    - "custom_action" (associative array) (default: {}): optional custom link with optional popover, e.g.:
```
emeryro, line 52: can't we use the default ECL Link structure, instead of defining each fields?
I know that this is the case for the language switcher, but maybe we will be able to also update it in the future

src/components/site-header/site-header.js
```diff
@@ -681,6 +785,117 @@ export class SiteHeader {
...
+    if (isHidden) {
+      this.openCustomActionOverlay();
+      this.focusTrapCustomAction.activate();
+      this.updateCustomActionOverlay();
+    } else {
+      this.focusTrapCustomAction.deactivate();
+    }
+  }
+
+  updateCustomActionOverlay() {
```
emeryro: would it be possible to have something more generic here? I assume that this code is more or less duplicate of the language switcher, so maybe we could have a single function for both
Note: this may be the case for some other function specific to the custom action here

emeryro (changes requested):
a few remarks:
- the custom action should be placed first, before the login
- let's use a generic label and icon (for instance "custom") instead of "notifications"
- the scss sub file where you put the code is called "_site-header-language-switcher". So we should either rename it or split it

src/components/site-header/_site-header-language-switcher.scss
tibi2303: yes, i cleaned up

src/components/site-header/site-header-custom-action.html.twig
tibi2303: added it in the main template

src/components/site-header/demo/data--ec.js
tibi2303: yes, added the href.

src/components/site-header/eu-site-header.story.js
tibi2303: hidden by default now

src/components/site-header/site-header.html.twig
tibi2303: made it to use ecl link

src/components/site-header/site-header.js
tibi2303: i made the arrow position and the overlay width using css. Removed not needed things in js

src/components/link/link.html.twig
```diff
@@ -8,6 +8,7 @@
         type: '' (can be '', 'standalone', 'cta', 'primary', 'secondary')
         inverted: false (boolean),
         no_visited: false (boolean),
+        button_behaviour: false (boolean),
```
emeryro: I would prefer to keep this as an exception in the site header (just adding the button classes there), and not a real variant of the link
This is a hacky thing, and we should not make it too visible to users

src/components/site-header/README.md
```diff
@@ -88,6 +94,20 @@ npm install --save @ecl/site-header
         ... 
       ], 
     }, 
+  },
+  custom_action: { 
+    href: '/notifications',
```
emeryro, line 100: this should be updated with the new link structure

emeryro (changes requested):
That's already cleaner, thanks
I added a few extra remarks
Also I noticed that the dropdown is not centered for the custom action, while it is the case for the language switcher
image

image

src/components/site-header/_site-header-dropdown.scss
```diff
@@ -335,3 +334,10 @@ $language-list: null !default;
     width: calc(#{map.get($language-list, 'item-width')} - (2 * var(--s-s)));
   }
 }
+
+@include breakpoints.up('l') {
+  .ecl-site-header__custom-action-overlay {
+    right: 50%;
```
emeryro: this will not work if the custom action is on the right of the screen (no search or language switcher)
image
I don't really understand: why do we have a custom behavior here, instead of reusing what we have for the language switcher?
The language switcher overlay has some custom content for sure, but the way it is placed on the screen is generic, if I'm not wrong

emeryro (changes requested):

src/components/site-header/_site-header-dropdown.scss
tibi2303: i did not want to use it because it needed to be refactored because it was hardcoded for language list. In any case, i refactored and now supports also custom action. Also fixed some issues on eu.

src/components/site-header/site-header.html.twig
```diff
@@ -259,6 +270,65 @@
             {# Header actions #}
             <div class="ecl-site-header__action">
             
+            {# Custom action #}
+            {% if _custom_action is not empty %}
+              <div class="ecl-site-header__custom-action">
+                {% set custom_action_link = {
+                  link: _custom_action.link|merge({
```
emeryro: we should clean up things here, we don't need to force the icon position, and the button_behaviour is no longer used

src/components/site-header/site-header.html.twig
```diff
@@ -259,6 +270,65 @@
...
+                    icon_position: 'before',
+                    type: "standalone",
+                    button_behaviour: true
+                  }),
+                  icon: {
+                    path: _icon_path,
+                    name: _custom_action.link.icon, 
+                    extra_classes: 'ecl-site-header__icon'
+                  },
+                  extra_attributes: [
```
emeryro: we will lose any potential icon parameter and custom attributes set by the user; we should merge them first and override what is needed

src/components/site-header/demo/data--ec.js
```diff
@@ -22,6 +22,27 @@ module.exports = {
     label: 'Log out',
     href: exampleLink,
   },
+  custom_action: {
+    link: {
+      path: exampleLink,
+      label: 'Custom label',
+      icon: 'brochure',
```
emeryro: I just noticed, but the data structure for the icon is not correct; it should be defined as an object outside of the link

emeryro (changes requested):
Thanks, the display seems to be fine now.
Before we send the preview to COMM, could you do the following extra steps?
- update the demo data to hide the link label and display the indicator, as it is the foreseen use case (EC and EU)
image
- add aria-live="polite" on the ecl-site-header__custom-action wrapper 

There may be some other accessibility feedback later, to be followed on the comments https://citnet.tech.ec.europa.eu/CITnet/jira/browse/ECL-1278

src/components/site-header/site-header.html.twig
tibi2303: cleaned up things, also now the icon params should not be forced

src/components/site-header/site-header.html.twig
tibi2303: indeed, now i merge them and improved readability

src/components/site-header/demo/data--ec.js
tibi2303: true, i moved it outside now of link

PR #3905, opened 2025-03-14
feat(divider): Renaming separator to divider, updating EC color - FRONT-4839

planctus:


emeryro (changes requested):
We should update netlify CMS configuration with the new usage page names

Also, not related to this change, but I noticed an issue in the divider story: there is an extra `<ol>`, triggering accessibility issue
image

emeryro (approved):

src/components/separator/separator.scss
```diff
@@ -12,5 +12,5 @@ $separator: null !default;
 %ecl-separator {
   border-width: 0;
   border-top: 1px solid map.get($separator, 'color');
-  margin: var(--s-xs) 0;
+  margin: 0;
```
emeryro, line 15: was it a request to remove this spacing?
display is a little different now, I wonder if that would not cause issues
image
image

planctus: yes, it was requested by the designers, i've asked ewpp what impact this would have been before applying the change, they told me that they set a bigger margin on their own on the concerned elements, so it would not been an issue. I have been warning also the designers about this but they insisted.

emeryro (approved):

PR #3913, opened 2025-03-21
feat(grid): Different columns gutter based on breakpoints - FRONT-4849

planctus:


emeryro (changes requested):
Change seems fine for the grid, but the navigation list is broken (still using the old variables)
image
To be checked if this is the only one

emeryro (approved):

PR #3916, opened 2025-03-24
feat(color): update component A-C - FRONT-4844

emeryro:


planctus (commented):
I see that we miss the category filter from the list of components to be handled in this ticket, checking in figma is one of those that should react to the color mores so i guess we have forgotten it.

emeryro:
yes, I noticed it.
I handle the category filter in https://github.com/ec-europa/europa-component-library/pull/3923

src/themes/ec/variables/_checkbox.scss
```diff
@@ -2,32 +2,37 @@
...
     content: '',
     display: block,
-    border-radius: 50%,
     position: absolute,
     top: 50%,
     left: 50%,
     transform: translate(-50%, -50%),
-    background-color: var(--c-p-20),
-    z-index: -1,
+    background-color: color-mix(
```
planctus: but..wasn't this the chance to simplify these styles..? I understand that they use layers to get the desired color, but on our side can't we just set a color instead of reproducing this madness in css..? Also because doing it like this we reduce the visibility of the white stroke of the check, if we keep this we can fix this issue setting a z-index for the icon, but to me this should be ultra simplified.

planctus:
To me it's good, just that i remember that the designers didn't want this element to change with the color modes, it might simply be that they changed their mind, just a confirmation..
image

planctus (approved):

emeryro:
that's how it is in Figma currently, so I assume that it is correct (I didn't change this value in the PR)

PR #3917, opened 2025-03-25
feat(color): update component N-R. - FRONT-4847

tibi2303:


emeryro:
@tibi2303 I didn't check everything here yet, but for the radio the design has changed for the hover/focus, so we have to adapt (no more outer circle).
I did it for the checkbox too here https://github.com/ec-europa/europa-component-library/pull/3916

emeryro (changes requested):
We don't need to add a parameter for color mode if the component is not using one of the color affected by it
https://github.com/ec-europa/europa-component-library/blob/v5-dev/src/themes/ec/maps/color.scss#L337
As far as I can see, in this set of components, none is affected

src/themes/ec/variables/_radio.scss
```diff
@@ -14,19 +14,17 @@ $radio: (
...
-    content: '',
-    display: block,
-    border-radius: 50%,
-    position: absolute,
-    top: 50%,
-    left: 50%,
-    transform: translate(-50%, -50%),
-    background-color: var(--c-p-20),
-    z-index: -1,
+  hover-effect: (
```
emeryro: we can probably simplify this a little; here we are using a single nested element, and only for one thing
This may be a simple parameter like 'box-color-hover'. If we define it in EC and EU, we won't have to test for it in the scss

emeryro (changes requested):
Just a small cleanup, but the display if fine

PR #3918, opened 2025-03-26
feat(header): Added custom-action V4. - FRONT-4819

tibi2303:
The same as v5.

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -27,14 +27,14 @@
   },
   {
     "path": "dist/packages/ec/scripts/ecl-ec.js",
-    "limit": "240 KB",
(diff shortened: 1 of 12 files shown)


emeryro (changes requested):
we should just apply the last modification made for v5 (move indicator to the left), and fix the test / size
The rest is fine

tibi2303:
Thanks @emeryro. updated tests and increased limits by minimal, also the indicator is on the left.

emeryro:
on EU I still see the indicator on the right (same on v5)
image

tibi2303:
@emeryro thanks romain, indeed on eu variant it was not moved, i moved it also on v5

PR #3920, opened 2025-03-27
feat(colors): Update components from S to Z

planctus:


src/themes/ec/_custom-properties.scss
```diff
@@ -374,6 +374,7 @@
   --cm-surface-low-2: var(--c-l-300);
   --cm-surface-medium: var(--c-l-400);
   --cm-surface: var(--c-d-900);
+  --cm-surface-high: var(--c-d-500);
```
emeryro: this token is not in the [excel file ](https://eceuropaeu.sharepoint.com/:x:/r/teams/GRP-COMMDesignteam-Designonly/_layouts/15/doc2.aspx?sourcedoc=%7B56441792-388B-4F00-9C3F-27B954A053EE%7D&file=Token%20list%20overview.xlsx&action=default&mobileredirect=true&wdOrigin=TEAMS-MAGLEV.teams_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1738686722102&web=1)
I see it used in the table, but maybe they forgot to update this value

src/themes/ec/variables/_table.scss
```diff
@@ -2,26 +2,26 @@
 @use '../index' as *;
 
 $table: (
```
emeryro, line 4: as the table is now reacting to the color mode, we should add a control for that

emeryro (changes requested):

PR #3921, opened 2025-03-28
feat(site-header): Style updates - FRONT-4854 

planctus:


emeryro (changes requested):
According to the ticket, the site header actions (login, language, ...) should use font S

emeryro (approved):

PR #3925, opened 2025-04-01
feat(social-media): Removed social-media-share and color variants from social media follow. - FRONT-4853

tibi2303:


emeryro (approved):
This is fine
Could you just ask for designer review, on Teams?

PR #3932, opened 2025-04-02
feat(color): remove old EC color, small fixes - FRONT-4830

emeryro:
Also fix a few missed spot in component colors


src/themes/ec/_custom-properties.scss
```diff
@@ -2,52 +2,6 @@
 @use 'index' as *;
 
 :root {
-  // old colors, to be removed
```
planctus, line 5: i don't think we can do this thing without warning in advance ewpp and giving them the time to "migrate" to the new colors and related utilities. This might actually be pretty tricky, considering that these utilities could be nested into chunks of html as the wysiwyg value, but we can try..

emeryro: In fact, I warned them a few weeks ago already that the utilities for the colors will change.
So I guess that they already did the change on their side

planctus (approved):

PR #3933, opened 2025-04-02
feat(menu): Bug fix tablet position. - FRONT-4756

tibi2303:


src/components/menu/menu.js
```diff
@@ -514,7 +514,7 @@ export class Menu {
     const isTablet = browser.getPlatformType() === 'tablet';
 
     // Detect mobile devices
-    if (isMobile) {
+    if (isMobile && window.innerWidth < this.breakpointL) {
```
planctus, line 517: what is this supposed to fix..? :) The issue in the ticket is about the vertical positioning of the menu, it can happen that the calculations fail and the menu is wrongly positioned in desktop, but in very peculiar conditions.
Here you are changing the logic we use in this component to detect whether to use the "desktop display" or not, in the current logic if it's not a mobile and not a tablet we still return false if the viewport is less than the "l" breakpoint we have, with what you did we would return true always, if it's not a mobile or tablet.
You can see one of the consequences in mobile in this pull request, the position of the menu is wrong.
I told you that the issue in the ticket is really difficult to reproduce in the real world, it's easier using the inspector, but i believe it's not worth doing it, i can explain you what the issue is exactly, anyway, if you still want to work on it.

planctus (commented):
Mmhh..not sure what you wanted to do here, in case you want to work on this issue i can explain to you what it really is about otherwise i think we can close this.

planctus:
can we close this @tibi2303 or do you plan to do something about it..?

tibi2303:
no, i am going to close it.. sorry

PR #3936, opened 2025-04-07
feat(featured-item): Removed description margin first-child. - FRONT-4860

tibi2303:
Seems to works fine with only this change. I checked rtl, print, position left-right


emeryro (commented):
This seem to work indeed, but I was wondering why we have this issue only when the media is on the right.
turns out that we are defining a wrong margin in that situation
image
Maybe we could fix this margin instead of having an additional override

emeryro (approved):

PR #3938, opened 2025-04-07
feat(typography): Responsive font. - FRONT-4861

tibi2303:


src/themes/ec/variables/_featured-item.scss
```diff
@@ -9,8 +9,7 @@ $featured-item: (
   max-width: 918px,
   title: (
     color: var(--cm-on-surface-brand),
-    font-mobile: var(--f-2xl),
-    font-desktop: var(--f-3xl),
+    font: '3xl',
```
emeryro: do you think it would be possible to still use `var(--f-3xl)`, and have the mixin handle it correctly?
that way we would not have to figure out if we should put a var or a direct value

src/themes/ec/mixins/_typography.scss
```diff
@@ -0,0 +1,36 @@
...
+  'xl': 'l',
+  'l': 'm',
+  'm': 's',
+  's': 'xs',
+);
+
+/// Responsive font
+/// @param {String} $desktop-size - Font size for desktop (e.g. '3xl')
+/// @param {String | null} $mobile-size - Optional override for mobile font size
+@mixin responsive-font($desktop-size, $mobile-size: null) {
```
emeryro: we should define this mixin in EU too (even if empty) or it will break

src/themes/ec/mixins/_typography.scss
```diff
@@ -0,0 +1,36 @@
...
+    map.get($font-size-map, $desktop-size)
+  );
+
+  @if not $mobile {
+    $mobile: $desktop-size;
+  }
+
+  font: var(--f-#{$mobile});
+
+  @include breakpoints.up('m') {
```
emeryro: the breakpoint could also be a parameter, with a default value

emeryro (commented):
overall it seems like a valid approach, just a few remarks

src/components/featured-item/featured-item.scss
```diff
@@ -29,9 +30,10 @@ $featured-item: null !default;
 
 .ecl-featured-item__title {
   color: map.get($featured-item, 'title', 'color');
-  font: map.get($featured-item, 'title', 'font-mobile');
   font-weight: map.get($featured-item, 'title', 'font-weight');
   max-width: map.get($featured-item, 'max-width');
+
+  @include responsive-font(map.get($featured-item, 'title', 'font'));
```
emeryro: I see a possible issue here: the font is defined after the font weight, so it will override it.
We may consider changing the order, or having the font weight handled by the mixin

tibi2303: thanks romain, i included also the font-weight... also rewrote some stuff because the mixin was not getting the mobile properly.

src/themes/ec/mixins/_typography.scss
```diff
@@ -0,0 +1,70 @@
...
+  font: $resolved-mobile;
+
+  @if $font-weight {
+    font-weight: $font-weight;
+  }
+
+  // Only add desktop font-size at breakpoint if it differs from mobile
+  @if $resolved-desktop != $resolved-mobile {
+    @include breakpoints.up($breakpoint) {
+      font: $resolved-desktop;
```
emeryro: we should have the font weight for mobile too

src/components/featured-item/featured-item.scss
```diff
@@ -75,7 +79,6 @@ $featured-item: null !default;
   }
 
   .ecl-featured-item__title {
-    font: map.get($featured-item, 'title', 'font-desktop');
     font-weight: map.get($featured-item, 'title', 'font-weight');
```
emeryro: we can remove this, as the font weight is handled by the mixin

src/themes/ec/mixins/_typography.scss
```diff
@@ -0,0 +1,70 @@
+@use 'sass:map';
+@use 'sass:string';
+@use '@ecl/grid/mixins/breakpoints';
+
+$font-size-map: (
```
emeryro, line 6: if I'm not wrong, there is a link between the font size and the font weight, for the designers.
If that is the case, we could have a default value for the font weight based on another mapping ('3xl': 300, for instance). 
We would still be able to override the value in the mixin, but for standard case we would not have to define it each time

emeryro (changes requested):
A few things to be improved

While we are at it, we could also add the letter spacing in that mixin, for compact/extended. It is related to the font size, so that shouldn't be an issue. We should be able to override it too, similar to what is done for the font-weight
https://github.com/ec-europa/europa-component-library/blob/v5-dev/src/themes/ec/maps/typography.scss#L84

src/themes/ec/mixins/_typography.scss
tibi2303: added font-weight

src/components/featured-item/featured-item.scss
tibi2303: removed

src/themes/ec/mixins/_typography.scss
tibi2303: also added letter spacing too, if there is another map for letter spacing, that can be overwritter in the param

emeryro (approved):

PR #3941, opened 2025-04-08
fix(mega-menu): mega menu html validation - FRONT-4859

planctus:
Testing the snapshot we get from jest there is one remaining error:
Error: Attribute an-extra-attribute-for-the-featured-link not allowed on element [a](https://html.spec.whatwg.org/multipage/#the-a-element) at this point
and a few warning for the unneded trailing slash in void elements.
Both are twing specific behaviours, this is not happening in twig, php.

A space has been added in the subitems between classes and attributes, this is the other way around, in twing it is ok but in twig it generates invalid markup, this is visible in the validator's report on the commission's website: https://validator.w3.org/nu/?doc=https%3A%2F%2Fcommission.europa.eu%2Findex_en

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -914,6 +914,8 @@ importers:
 
   src/layout/grid: {}
 
+  src/layout/headings: {}
+
   src/page-example/page-fact-sheet:
     dependencies:
       '@ecl/banner':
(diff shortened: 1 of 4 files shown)


emeryro (approved):
not a big fan of these added `&nbsp;`...
it does not seem to brake anything so this is fine, but this looks like an old hack

planctus:
i know, not a fan myself, but it turns out that both twig and twing have this quirk so that an empty container will be rendered without closing tag even if it's there in the twig template, generating invalid html, this way we fix this wrong behaviour...

PR #3948, opened 2025-04-14
fix(menu): fix current item - FRONT-4868

emeryro:
Fix display of current item on hover and focus

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -914,8 +914,6 @@ importers:
 
   src/layout/grid: {}
 
-  src/layout/headings: {}
-
   src/page-example/page-fact-sheet:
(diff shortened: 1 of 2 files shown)


planctus (commented):
This thing is incredibly confusing, with this code we fix the bug, but at the same time we introduce even more code realted to the styles of these items, it seems to me that we could fix this removing the exclusion for the current item in this rule: 
image
and maybe the same happens for the focus..? If it's not the case just discard this comment.

emeryro:
that was my first intuition indeed, but it didn't work just by adding this exclusion.
I can have a second look at it, but I fear that if we start changing things in the menu, we end up refactoring half of it

planctus (approved):

PR #3950, opened 2025-04-16
feat(playground): Storybook plugin to enable/disable stylesheets on the fly - FRONT-4867 

planctus:


emeryro (changes requested):
This is nice!
Just one thing: we should not override the text color, as it does not work for storybook dark mode
image

emeryro (approved):

PR #3952, opened 2025-04-16
fix(fonts): Fixing fallback on local font for Inter family - FRONT-4875

planctus:


emeryro:
I'm a little concerned by this. With the multiple font face as proposed, won't it always take the local font in the end, and so skip the one from the webtools cdn?
I would rather see it the other way around...
Also, what if we keep the multiple url in a single font face, but just switch them?

planctus:
right, separating the declaration was not the correct way to handle this, in reality the order is the right one, because it uses the first good value in src and only if the first fails it tries the second and so on.
This can be tested, currently the webtools url works also in our preview, but we can pretend it doesn't by changing the url, the browser would try the url, get a 404 and then it would load the local font as we want.

emeryro (approved):

PR #3960, opened 2025-04-24
fix(website): String replacement in getting started page - FRONT-4871

planctus:


emeryro (commented):
I still see them as n/a on the preview, is it supposed to work there too?
image

planctus:
well, the fact that you see n/a already means that the replacement is happening, but those sri are generated only when releasing, so to test this you would have to make a release locally, or i can create a "fake" release here in github for then deleting it.
To do this locally you would need to set two env variables:

- CI to true
- GITHUB_REF=refs/tags/test

Then you need to create a file in /scripts, named europa-component-library-test-sri.json with some content like: 
{
  "ecl-ec.css": ["sha384-abc123css"],
  "ecl-ec-print.css": ["sha384-def456printcss"],
  "ecl-ec.js": ["sha384-ghi789js"],
  "ecl-esm-ec.js": ["sha384-jkl012esmjs"]
}

Then you run dist and if you access the website in dist/website (you would need a vhost with this root) in getting started you should have these sri replaced.
It seems probabily easy to make a release here in github, though.. :)

PR #3967, opened 2025-04-29
feat(footer): new EC footer standardised/harmonised - FRONT-4851

emeryro:
Full rewrite of the EC footer
See migration doc for more information


planctus:
This is great, but i wonder about a couple of things, this is the first time we are introducing different templates for EC and EU, they are not just rendering a different HTML, they expect a different data set, on our side this is because we were requested to create new versions of EC while maintaining EU as it is. ASsuming that this is fine for our users, maybe we could still refactor EU now to use these new templates, maintaining the current look and feel, just with a different markup, if that is not overcomplicated. Otherwise let's include in the readme two separated sections for the two systems (the include's example should still be only one otherwise the replacement with the current data passed would not work)

emeryro:
I had a second look for EU, for that would be really complex to match the EC data structure, as the sections are quite different.
Also the EU footer would be updated at some point, so we would have to change the structure again

planctus (approved):

PR #3968, opened 2025-04-29
feat(typography): Changed to responsive-font approach. - FRONT-4870

tibi2303:
I expected more changes to be done.. but these were the use cases I could find. Too bad we cant use the default size map too much because mostly the ec/eu themes uses a lot of different values than the ones we have in the default map.


emeryro (approved):

src/mixins/README.md
```diff
@@ -0,0 +1,110 @@
...
+  @include mixins.responsive-font($custom-map);
+}
+```
+
+## Notes
+
+- `$token` can be either:
+  - A string (like `'heading1'` or `'body-m'`) that resolves inside the `$typography` map.
+  - A direct SCSS map matching the expected structure.
+- Font family is applied from `$typography['font-family']`.
```
planctus, line 114: it should, but you have it harcoded in the mixin instead

src/mixins/typography/mixins.scss
```diff
@@ -0,0 +1,84 @@
...
+
+    @if $variant {
+      $token-map: map-deep-get($typography, $base-token, $variant);
+    } @else {
+      $token-map: map.get($typography, $base-token);
+    }
+  }
+
+  // Font family
+  $font-family: var(--ff-d);
```
planctus: this should be retrieved from the map, no..?

src/mixins/README.md
```diff
@@ -0,0 +1,110 @@
...
+);
+```
+
+## Usage Examples
+
+### Using a token string
+
+```scss
+.title {
+  @include mixins.responsive-font('heading1');
```
planctus, line 69: this syntax is specific of a certain way of importing the mixins, not so important maybe but it would be better to also include the @use syntax in these examples

src/mixins/typography/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "5.0.0-alpha.10",
+  "description": "ECL typography",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@ecl/grid": "5.0.0-alpha.10",
+    "@ecl/theme-ec": "5.0.0-alpha.10"
```
planctus: why this dep here..?

src/mixins/README.md
tibi2303: i added the use in these syntaxes

src/mixins/typography/package.json
tibi2303: leftover, removed

src/mixins/typography/mixins.scss
tibi2303: i changed it to get from typography map.. romain said to use vars in components and i though it would be good here too, cause it would be easier to override

PR #3972, opened 2025-04-30
feat(mega-menu): New iteration - FRONT-4883

planctus:
In this pull request:

1. moved the "info link" in the "second" panel in desktop, in mobile it goes after the list of items
2. Made the featured panel visible even in the absence of children for second level menu items
3. Added a "promotional" variant for the item providing styles through custom properties that are supposed to be overridden directly by the users
4. Added the possibility of setting the featured panel as part of the first level items, when this is set it will appear in every child. If both the first level item and the second level item are defining a featured panel what happens by default is that the one from the second level will be shown after clicking on the item, a parameter has been added to invert the process and keep showing the panel from the first level item hiding the one from the clicked item.


src/components/mega-menu/mega-menu.scss
```diff
@@ -12,6 +12,29 @@
 $theme: null !default;
 $mega-menu: null !default;
 
+:root {
```
emeryro, line 15: we may define these specific variables at the root of the mega menu component (.ecl-mega-menu), instead of the page root

docs/Migrating-v5.md
```diff
@@ -207,6 +207,13 @@ featured: {
 
 The image will be clickable and will act as the associated link.
 
+It is now possible to highlight a menu link (first level) associated to a special event or page passing the "promotional" paramter as part of the item's data.
+The related styles can be customized defining:
+--ecl-mega-menu-item-promotional-bg
```
emeryro, line 214: I wonder if we will be able to cover the different use cases with these variables only.
The text color on hover may have to be different, same for the focus

src/themes/ec/variables/_mega-menu.scss
```diff
@@ -8,6 +8,10 @@ $mega-menu: (
     greysh-background: var(--cm-surface-neutral-lowest),
     shadow: none,
     font: var(--f-s),
+    promotional-item-bg: var(--c-p-400),
```
emeryro: are these colors provide by the designer? The contrast ratio is too low here

src/components/mega-menu/mega-menu.scss
planctus: mmh..i chose to put them at root to simplify the life of those who might override them, the variables names are specific to the component and to the element they refer to, no need to nest them into a component selector

docs/Migrating-v5.md
planctus: oh, gosh..i noticed only now that they set a different color on focus including a border of a different color, i can't believe it..

src/themes/ec/variables/_mega-menu.scss
planctus: they were, then they changed them and i'm pretty sure they will change them again although these are only the default colors, they are supposed to be overridden..

docs/Migrating-v5.md
planctus: but in the end i haven't applied the border, i recalled the issues with retina displays we had, to be able to add the border safely we should apply the same to all the elements and it doesn't seem reasonable to do so for this minor style on focus for this specific item.

src/components/mega-menu/mega-menu-item.html.twig
```diff
@@ -6,6 +6,12 @@
   - "item" (object)
     - "label" (string) Label of the menu link
     - "external" (boolean) External link
+    - "promotional" (boolean) (default: false) Promotional menu item, styles can be custommized defining:
+        --ecl-mega-menu-item-promotional-bg
+        --ecl-mega-menu-item-promotional-hover-bg
+        --ecl-mega-menu-item-promotional-outline-color
+        --ecl-mega-menu-item-promotional-text-color
+        --ecl-mega-menu-item-promotional-hover-text-color
```
emeryro, line 15: missing focus variables here

emeryro (changes requested):
just a comment to be updated, the rest is fine

emeryro (approved):

PR #3976, opened 2025-05-05
feat(banner): mobile and tablet display - FRONT-4892

emeryro:
Fixed aspect ratio for mobile and tablet


planctus:
A few issues, only visible in the usage of the banner in the carousel, though:

1.
image
this is not 3/1, it seems that this is caused by this rule: 
image
removing it doesn't seem to break anything, but to be checked carefully
2.
image
we have an align-items: center that moves the text of the info area in the middle, this is probably ok in desktop, but tablet and mobile should not have it, i think.

planctus:
image

Maybe this is why we had that height: 100%..?

emeryro:
I don't have issues on my side, the display seems correct (in the preview and local)
Maybe this was fixed when merging v5-dev?
image

planctus (approved):

PR #3978, opened 2025-05-06
feat(typography): Adding values for tablet font styles - FRONT-4895

planctus:
I'm not entirely sure about what breakpoints matches the "tablet" and "desktop" display, in the typograhpy utilities i have used "m" for tablet and "l" for desktop.

It's not particularly clear the situation regarding the letter spacing, astrid m said to me that she is going to generate a list of the different letter spacing needed, for now i set the values directly in the maps.

To be able to generate utilities for EU some variables (font-size and line-height) have been added also there.

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -15,7 +15,7 @@
   },
   {
     "path": "dist/packages/eu/styles/ecl-eu.css",
-    "limit": "290 KB",
+    "limit": "295 KB",
     "webpack": false,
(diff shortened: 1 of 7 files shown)


emeryro (approved):
from what I've seen so far, it seems that tablet covers breakpoints m and l, desktop starts at xl
But anyway, this will be handled by the mixin in the end, same for the letter spacing hopefully

PR #3980, opened 2025-05-07
feat(typography): Updating letter spacing scale in EC - FRONT-4896

planctus:


emeryro (commented):
I'm surprised that they removed the 'extended/compact' thingy, but fine, for once it make things easier
Still I'm confused by the values. Letter spacing go from 4xs to 5xl, while font size go from 2xs to 10xl. How are we supposed to do the mapping here?

Also we should make sure to keep the migration and website doc up to date with all these changes

planctus:
i don't think that they removed anything, there was no scale before on their side, we built it that way in order to be able to use it in relation to the font size, but i think it was arbitrary, the request by the designers is indeed to separate  from the font sizes, so it's a standalone scale now.
I don't know how we will "map" it, but it might be that there is nothing to map, currently only a few typography items have a "compact" or "extended" version, so in the mixin we will retrieve the value from the typography map and for the moment i wouldn't generate any utility for the letter spacing.

emeryro (approved):

PR #3981, opened 2025-05-12
fix(select): Fixing accessibility report - FRONT-4897

planctus:
In this pull request:

1. removed role attribute from the form group wrapper
2. removed the customisation of the for attribute for the label in the multiple select in order to associate it with the button replacing the select, not sure why this has been done, considering that it was triggering an error in our accessibility test in storybook about a missing label for the form element.

Maybe this second point is also relevant regarding the other request in the ticket, although i did not fully understand it since it mentions ids of a div element that i could not identify.


emeryro (commented):
Seems good for me, but I asked confirmation on the initial ticket

emeryro (approved):

emeryro (approved):

PR #3988, opened 2025-05-19
feat(typography): Changed to responsive font approach in components. - FRONT-4900

tibi2303:


src/themes/ec/variables/_banner.scss
```diff
@@ -44,47 +44,91 @@ $banner: (
...
+          'size': var(--fs-7xl),
+          'line-height': var(--lh-6xl),
+        ),
+        'desktop': (
+          'size': var(--fs-10xl),
+          'line-height': var(--lh-10xl),
+        ),
       ),
-      'description': (
+      'description-s': (
```
emeryro: description font weight should be regular for small font (it is light currently)

src/themes/ec/variables/_site-header.scss
```diff
@@ -12,6 +12,16 @@ $site-header: (
   search-width-l: 255px,
   search-width-xl: 295px,
   toggle-color: var(--cm-on-surface-brand),
+  banner-typography: (
```
emeryro: line height seems different (mobile and desktop)

src/themes/ec/variables/_language-list.scss
```diff
@@ -15,8 +15,17 @@ $language-list: (
     inline-start: var(--s-xl),
     inline-end: var(--s-xl),
   ),
-  header-font-mobile: var(--f-2xl),
-  header-font-desktop: var(--f-3xl),
+  header-typography: (
```
emeryro: font seems different (mobile and desktop)

emeryro (changes requested):
As expected, this is a great simplification for the banner
I noticed a few differences between the preview and v5-dev.
Also, can't we use the predefined font directly (h2, h3, ...)? We can check case by case if there are differences, but as the mixin supports it, the easiest the better

src/themes/ec/variables/_language-list.scss
tibi2303: changed, thanks

src/themes/ec/variables/_site-header.scss
tibi2303: changed.. should be the same as before.

src/themes/ec/variables/_banner.scss
tibi2303: changed.

src/themes/ec/variables/_language-list.scss
tibi2303: I would have changed more, to simple heading approach as param, but the heading in ec have font-weight light as default.. and would not be the same.

emeryro:
I feel like it should be possbile to simplify more.
If I take the banner for instance, the title could be "display" or "H1" (depending on the variant), and the description could be "H2" or "body 2XL". Can't we use the font directly (with specific font-weight when needed)?
To me, the custom definition should be an exception. If there are cases where the font does not correspond, we will check case by case

src/themes/ec/variables/_site-header.scss
planctus: please don't apply any change to the site header, we've got a request for style changes here and there about the site -header including the site name, these changes would surely conflict with those. Thanks

tibi2303:
Hey @emeryro using your new param added for font-weight, i simplified more on banner and header using what the token you said. Thanks

src/themes/ec/variables/_site-header.scss
tibi2303: removed the differences here

emeryro (changes requested):
@tibi2303 I did another review and noticed a few font size issue, when comparing to the current v5-dev. Could you double check the values?
Then we will be able to merge this task

Issues:
- banner EU, when selecting large or small font size
- site header EC/EU, language switcher (title)

Note: for EU you can still use the previous font variables (var--f-*), instead of defining the font size and line height. The font update was mainly for EC

tibi2303:
@emeryro hello romain,

regarding your point here Note: for EU you can still use the previous font variables (var--f-*), instead of defining the font size and line height. The font update was mainly for EC

because of the different sizes on viewports on eu, the mixin was not supporting this kind of behaviour
`
        'mobile': var(--f-2xl),
        'desktop': var(--f-5xl),
 `
 direct font tokens, so i needed to enhancce the mixin and also simplify the responsive breakpoint part. The checking needed an improvement in order to make it work and also keep the previous behaviour.
 
 and for 
 
 banner EU, when selecting large or small font size
site header EC/EU, language switcher (title)

there are both the same as on v5 dev now

emeryro:
Thanks @tibi2303, the display is fine now, remaining points have been fixed.
I just notice that with the recent changes made on the mixin, there are some deprecation warning raised. This is not the case on v5-dev, so it seem to come from this PR. Could you check?
https://github.com/ec-europa/europa-component-library/actions/runs/16315659991/job/46080845851?pr=3988#step:7:56

tibi2303:
@emeryro done, i fixed the deprecation warnings, thanks.

emeryro (approved):

PR #3992, opened 2025-05-21
feat(footer): add custom sections - FRONT-4904

emeryro:


src/components/site-footer/README.md
```diff
@@ -9,7 +9,7 @@ npm install --save @ecl/site-footer
 ### EC footer parameters
 
 - **"section_xx"** (object) (default: {}): Specific section of the footer, all using the same structure
-  Available sections: "section_site_info", "section_core", "section_contact", "section_about", "section_more", "section_related" and "section_common"
+  Available sections: "section_site_info", "section_core", "section_contact", "section_about", "section_more", "section_related", and "section_common"
```
planctus:
```suggestion
  Available sections: "section_site_info", "section_core", "section_contact", "section_about", "section_more", "section_related" and "section_common"
```

src/components/site-footer/site-footer-ec.html.twig
```diff
@@ -120,6 +122,21 @@
...
         </div>
         {%- endif -%}
+
+        {# Custom sections #}
+        {%- if _sections_custom is not empty -%}
+          {% for key, pair in _sections_custom|batch(2) %}
+            <div class="ecl-site-footer__row">
+              {% for custom in pair %}
+                {% include '@ecl/site-footer/site-footer-ec-section.html.twig' with {
+                  section: custom|merge({ name: 'custom-' + ((key * 2) + loop.index) }),
```
planctus:
```suggestion
                  section: custom|merge({ name: 'custom-' ~ ((key * 2) + loop.index) }),
```
It works as it is because twing is coercing the number we add to be a string and most likely this just works the same in twig php, but we normally and rightfully use tilde ~ for string concatenation, so i would do also here

planctus (approved):
There is just a minor suggestion, see if you want to apply it or not.. ;)

PR #3996, opened 2025-05-22
feat(typography): update microcopy token and utilities - FRONT-4905

emeryro:
added utilities for microcopy, and a demo to review them


src/themes/ec/variables/_typography.scss
```diff
@@ -199,20 +199,60 @@ $typography: (
   ),
   microcopy: (
     '2xs': (
-      'size': var(--fs-2xs),
-      'line-height': var(--lh-3xs),
+      'mobile': (
```
planctus: i don't get it, microcopy have the same values for all the viewports, why would we generate media queries to duplicate the same values, i wonder...

planctus:
At this point i would add also a microcopy section in the website, in the typography utilities, and in storybook it would make probably more sense to have this order: headings, paragraph, microcopy, text color, text style

PR #3997, opened 2025-05-26
feat(site header): Styles iteration - FRONT-4906

planctus:


emeryro (commented):
Almost good, I just have one question:
When the control "site name mobile only" is on, we still have the black (empty) bar on tablet. Shouldn't we display the site name here, or remove this bar?
image

emeryro (approved):

PR #3999, opened 2025-05-28
chore(page-examples): Fixing rendering - FRONT-4912

planctus:


emeryro (changes requested):
EU home page is still broken, but the other ones are fine
https://ecl-preview-3999--europa-component-library.netlify.app/playground/eu/?path=/story/page-examples-home--default&globals=viewport:responsive

emeryro (approved):

PR #4000, opened 2025-05-28
fix(select): Fixing position of the multiple select counter - FRONT-4915

planctus:


emeryro (commented):
to have the counter always centered, we could use something like that
image

emeryro (approved):

PR #4001, opened 2025-06-03
feat(timeline): headline and timeline set - FRONT-4908

emeryro:
Add possibility to have headlines and set of timelines


src/components/timeline/package.json
```diff
@@ -13,7 +13,8 @@
   "dependencies": {
```
planctus, line 13: we released alpha.14 in the meantime, can we align these deps..?

src/components/timeline/timeline-print.scss
```diff
@@ -51,70 +52,86 @@ $timeline-print: null !default;
...
-.ecl-timeline__item--toggle {
-  display: none;
-}
-
 // Border adjustment
-.ecl-timeline__item:first-child::before {
+// stylelint-disable no-descending-specificity
+.ecl-timeline__item:not(.ecl-timeline-set *):first-child::before,
+.ecl-timeline-set
+  .ecl-timeline:first-child
```
planctus, line 79: i foresee issues with the webcomponent, but ok.. :)

src/components/timeline/timeline-print.scss
```diff
@@ -51,70 +52,86 @@ $timeline-print: null !default;
...
-  );
+  font-weight: map.get($timeline, 'label', 'font-weight');
   margin-bottom: map.get($timeline-print, 'label-margin');
-  position: map.get($timeline, 'label-layout');
-  text-align: map.get($timeline, 'label-alignment');
-  width: map.get($timeline, 'left-spacing');
+  position: map.get($timeline, 'label', 'layout');
+  text-align: map.get($timeline, 'label', 'alignment');
+
+  @if map.has-key($timeline, 'left-spacing-desktop') {
```
planctus, line 118: oh gosh..

src/website/src/pages/ec/components/timeline/docs/code.mdx
```diff
@@ -11,3 +11,15 @@ import { Playground } from '@ecl/website-components';
...
+## Timeline set
+
+You can group multiple timelines by using a timeline set
+
+<Playground
+  system="ec"
+  selectedKind="components-timeline"
+  selectedStory="default"
+  selectedArgs="nb_timeline:2"
+  showFrame
```
planctus, line 24: we can remove all the usages of showFrame, basically, it's not used anymore, we always have an iframe in the demos now, no need to do it right now cause it's harmless, just saying this to let you know it.

planctus (commented):
too good, just overcomplex, the css is really difficult to read, i understand that in the maps we have a nice organization this way, but who looks at the maps, instead when using a nested property with map.get it gets really tricky, you cannot copy paste anything, i think that we would better keep the maps as flat as possible, but i see that it might be already too late, we have objects with nested properties in many components already.

I would just align the dependencies and then we merge it ;)

PR #4006, opened 2025-06-10
fix(mega-menu): Taking into account the minHeight, recalculating also on collapse - FRONT-4922

planctus:
This fixes two issues noticed after a report by ewpp on an example with only one item per panel.

- we set a min Height for the dropdown that was not taken into account, resulting in a white space below the panel's content
- When closing a previously opened third panel the height of the panels was not correctly recalculated so that it would not take into account the real heights of the panel being shown (see screenshots in the ticket)


emeryro (approved):
It is quite hard to check all the possible use cases.
Everything seems fine with this change

PR #4014, opened 2025-06-12
feat(highlight): add utilities - FRONT-4916

emeryro:


planctus (commented):
Ah, wow...not really what i thought we would do...
When we mentioned "component" to me that only meant something "more" than a utility, not anything that would require an include of a twig template.

It seems really weird to have to include a template inside a block of text, not even sure it would be feasible in the integration side, but it just feels it's not the appropriate implementation, to me this is still a utility, just that it might support in the future something more than a background color.

emeryro:
well, this is not just a utility, as there may be more css rules or variant added in the future, but this is not a real component either, I agree...
To me that was similar to the divider / separator (a single css class), and we ended up having a component for that, that's why I implemented it that way

planctus:
the difference to me is that the "separator" is a layout element, here we are dealing with a "text style", so something to be used in the text flow, that's why to me it would be hard to use it this way, but it's not really a philosophical choice to me, if for ewpp moving from a "selector" to a twig include would be fine, no issue about keeping it like this.

emeryro:
as discussed, element has been moved to utilities

src/utilities/highlight/highlight.story.js
```diff
@@ -0,0 +1,39 @@
...
+  decorators: [withCode],
+  parameters: {
+    EclNotes: { disable: true },
+  },
+};
+
+export const Highlight = (args) => {
+  let classes = 'ecl-u-highlight';
+  if (system === 'ec') {
+    classes += ` ecl-color-mode--${args.color_mode}`;
```
planctus: doing this adds by default a ecl-color-mode-- class, i think yo could simply check for the color mode arg itself, being that only in EC without even needing to detect the system here.
But to me, we shouldn't add the color mode class to the element itself, that is not really the right way to use them, it makes you think that the color mode is a sort of a duplicate of the background utilities, it should be set on a parent element i think, to mean that if a color mode is being used then this background is going to react to it.

src/website/src/pages/ec/utilities/highlight/docs/code.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: Showcase
+order: 2
+---
+
+import { Playground } from '@ecl/website-components';
+
+## Highlighted text
+
+<Playground
```
planctus: To me this should be part of the typography utilities, we have:

- text color
- text style

we could add text highlighted

The ratio is that this utility is just setting a background color (for the moment) if we detach it from the typography is a way to make the background utilities work with the color modes, which might make sense, maybe, but then it should not be named "highlghted text" which clearly relates to the text

planctus (changes requested):
There are two small things to be improved in the story but also the possibility to demo this in a different way, i let you evaluate my considerations and the resulting proposal.. ;)

src/website/src/pages/ec/utilities/highlight/docs/code.mdx
emeryro: my idea was to have a section grouping the different highlight ("highlighted text" currently, and the three variant of "enhanced paragraph" once done
But we could also put it in the typography...

planctus (approved):

planctus (approved):

PR #4019, opened 2025-06-17
feat(form): update form element display - FRONT-4924

emeryro:


planctus:
Just an issue in the EU multiple select: 
image

Then a super minor thing, in the textarea the default value in the controls panel for the width is "object object"
image

planctus (approved):

PR #4021, opened 2025-06-19
feat(accordion): Minimal tweaks to simplify the webcomponent implementation - FRONT-4937

planctus:
…- FRONT-4937

### DIFF
--- a/docs/Migrating-v5.md
+++ b/docs/Migrating-v5.md
@@ -125,6 +125,8 @@ Markup of accordion title has been updated to use a simple div instead of a head
 
 Corresponding twig parameter `level` has been removed.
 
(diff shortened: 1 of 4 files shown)


emeryro (commented):
I'm a little surprised by this. I did expect some issues when we have selectors like `.a + .b` or `.a > .b`, but here couldn't we solve this by using `:first-of-type` and `:last-of-type`?

planctus:
unfortunately no, because in the accordion the items are nested into a custom element, so the .ecl-accordion__item elements are not siblings

emeryro:
Could you just add this change to the migration document? Users not relying on the twig template would have to update on their side too.

emeryro (approved):

PR #4022, opened 2025-06-19
fix(icon): fix flag icon display - FRONT-4931

emeryro:
also a small fix for the icon story

### DIFF
--- a/src/components/icon/icon.story.js
+++ b/src/components/icon/icon.story.js
@@ -22,7 +22,7 @@ const getArgs = (data) => ({
   description: '',
 });
 
-const getArgTypes = (data, icons) => getIconControls(data, icons, iconMapping);
+const getArgTypes = (icons) => getIconControls(icons, iconMapping);
 
 const prepareData = (data, args) => {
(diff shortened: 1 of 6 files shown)


planctus:
Since we are touching this, i would also fix something in the story, we get "ecl-icon--none" because of the transformation control. 
In the website instead we have this text in the icon's showcase:
```
You can pick an icon from the library we provide
Note: display of icons on Internet Explorer is made possible by the use of svg4everybody
```
In the first sentence we could specify that the icons shown are not provided by us, the second is simply outdated.

Then, all the examples are just the same, since there is no argument passed to the story.

planctus:
mmh..ok for the text but we haven't touched the different examples, they are still just the same, while we could/should  showcase what we are describing, Icon color, Rotate icons, Flip icons, Accessibility, no..?

emeryro:
Examples updated. There is an issue on the website: the code snippet is not correct when using the same story multiple times (even if the example itself is correct). This is not related to the icons, as it occurs also on other components (file for instance), so I will open another ticket for that

planctus (approved):

PR #4026, opened 2025-06-24
fix(website): Fixing html source in showcase when using args from storybook - FRONT-4945

planctus:
tricky...we are not rendering the templates anymore in the website, so these snippets with the source code are coming directly from the stories, well, not so directly..We intercept messages sent by storybook when a story is loaded and rendered, but it's a lot of messages that we have to filter to identify the right story where to get the source code from.
Now it works, but there are cases where we have to specify more arguments than needed, because otherwise we might still get the source code from a different story.
So, if in the future we will experience this mismatch we know that specifying the args values can fix that.


emeryro (changes requested):
I spotted a few cases where the content is not correct on the website:
- button ghost inverted (missing background)
- content item left image (image is on right in the snippet)
- horizontal list with illustration - image square (no code snippet)
- some form elements have a wrong mix of optional/required in the code snippet

PR #4028, opened 2025-06-25
fix(mega-menu): overriding paragraph styles in featured panel description - FRONT-4949

planctus:


src/components/mega-menu/mega-menu.scss
```diff
@@ -494,9 +494,14 @@ $mega-menu: null !default;
...
 
   .ecl-mega-menu__featured-list__item-description {
-    font: map.get($mega-menu, 'global', 'featured-description-font');
     margin-inline-start: var(--s-l);
   }
+
+  .ecl-mega-menu__featured-list__item-description,
+  .ecl-mega-menu__featured-list__item-description p {
+    font: map.get($mega-menu, 'global', 'featured-description-font');
+    margin-block: 0;
```
emeryro, line 503: do we really want to remove all the vertical margins here?
if there are multiple paragraphs, there would no longer be spacing between them. Maybe we could just target the first and last paragraph

src/components/mega-menu/mega-menu.scss
```diff
@@ -494,9 +494,14 @@ $mega-menu: null !default;
   }
 
   .ecl-mega-menu__featured-list__item-description {
-    font: map.get($mega-menu, 'global', 'featured-description-font');
     margin-inline-start: var(--s-l);
   }
+
+  .ecl-mega-menu__featured-list__item-description,
```
emeryro, line 500: it is to have higher specificity that you didn't merge this block with the previous one?

planctus: no, it's because the margin-inline would have been applied to both..

src/components/mega-menu/mega-menu.scss
planctus: the idea is that there shouldn't be paragraphs inside that, or no..? to me this was resetting unwanted styles, otherwise we should ask the designers to determine how a more complex block of text would look like...
There is no space or better, vertical space there is precious, imagine with what they might come out with empty paragraphs or a paragraph for each sentence, to me here we are preventing this, not styling some unforeseen markup. But yeah, not happy about doing this anyhow, if you think we should handle this use case and not just delete any effect of the presence of this html tag,  we can do it, sure..

PR #4029, opened 2025-06-27
fix(carousel): Not preventing vertical scroll when swiping vertically - FRONT-4951

planctus:
…


emeryro (commented):
It works fine on the initial page load: I can swipe up and down without issues
But if I swipe left or right on the carousel, to go to another slide, the next swipe up/down also moves the carousel horizontally.
I don't know if you have the same behavior...

planctus:
yeah, i see that, let's try to see if we can improve this.

emeryro (approved):

PR #4030, opened 2025-06-30
feat(inpage): Styles iteration - FRONT-4948

planctus:


src/components/inpage-navigation/inpage-navigation.html.twig
```diff
@@ -61,7 +61,17 @@
   data-ecl-inpage-navigation="true"
   aria-labelledby="{{ _id }}"
 >
-  <div class="ecl-inpage-navigation__title" id="{{ _id }}">{{ _title }}</div>
+  <div class="ecl-inpage-navigation__title" id="{{ _id }}">
+    {% include '@ecl/icon/icon.html.twig' with {
+      icon: {
+        name: 'hamburger',
```
emeryro: the icon does not seems to be correct. In figma it is the list icon (and hamburger would not make much sense as it is not clickable)
image

emeryro (commented):
It is a little strange to have the border always blue on desktop, and picking color mode on mobile. If it has been approved like that, it's fine, but I don't see mockups in figma
image
image

planctus:
Right, changed the icon and used border-primary also in mobile, there are only examples in desktop but there isn't any reason why the behavior should be different in mobile.

emeryro (approved):

PR #4032, opened 2025-07-01
feat(content-block): Font styles - FRONT-4950

planctus:
I thought that adding a mixin would have been simplifying things, i'm more confused than ever instead, anyway it should be correct like this, both system had regular font weight so it is now passed directly in the component css, the thing is that i'm not sure the font-weight we set in the EC map for headings is still valid, it is 300 but i don't see any font-weight for the headings in the "font-token" in figma anymore.


src/components/content-block/content-block.scss
```diff
@@ -57,11 +58,14 @@ $content-block: null !default;
...
 .ecl-content-block__title {
   color: map.get($content-block, 'text-color');
-  font: map.get($content-block, 'title-font');
-  font-weight: map.get($content-block, 'title-font-weight');
   margin: map.get($content-block, 'title-margin');
   max-width: var(--max-w);
 
+  @include mixins.responsive-font(
+    map.get($content-block, 'title-font'),
+    map.get($theme, 'font-weight', 'regular')
```
emeryro, line 67: the font weight in the mixin is optional. If you are using regular, then there is no need to define it, you can just pass the font and that's it

emeryro (approved):
The title is quite small on mobile now, but if that's what they want...

I just have a small remark for the mixin use, but overall that's fine for me

src/components/content-block/content-block.scss
planctus: it might be in the next version, currently for the headings you get 300 by default

PR #4037, opened 2025-07-03
chore(grid): Removing XXL breakpoint, keeping container max width at 1368px - FRONT-4958

planctus:


emeryro (approved):
This is fine for me.
Let's just send a preview to the designers, so they can see that nothing has changed

PR #4042, opened 2025-07-08
feat(typography): add mobile-xs fonts- FRONT-4963

emeryro:
add definition of fonts for xs screen


src/mixins/typography/mixins.scss
```diff
@@ -5,17 +5,25 @@
...
 
-@mixin responsive-font-emit($tier-map) {
+@mixin responsive-font-emit($tier-map, $important: false) {
   @if meta.type-of($tier-map) == 'map' {
     $size: map.get($tier-map, 'size');
     $lh: map.get($tier-map, 'line-height');
 
     @if $size {
-      font-size: $size;
+      @if $important {
```
planctus, line 14: oh wow, why do we need to handle !important, that seems to be the case for an exception that might fall out of the usage of this mxin, it's already pretty tricky to deal with this sass, why adding complexity to it..?

emeryro: it is needed for the utilities. Without that, the "default" css takes over
image
We initially decide to use important for the utilities, so I kept the end result as it is

src/mixins/typography/mixins.scss
```diff
@@ -84,11 +97,14 @@ $font-breakpoints: (
...
-    @if map.has-key($token-map, 'mobile') {
-      @include responsive-font-emit(map.get($token-map, 'mobile'));
+    // Base (mobile-xs)
+    @if map.has-key($token-map, 'mobile-xs') {
+      @include responsive-font-emit(
+        map.get($token-map, 'mobile-xs'),
+        $important
+      );
     } @else {
-      @include responsive-font-emit($token-map);
```
planctus, line 91: there is a problem, if mobile-xs is not present then we don't get any font-size and line-height before the m breakpoint, this can be seen in the site header site name.

image

planctus (changes requested):
To be honest, i regret having launched the initiative to create this mixin, the more i look at it the more confused i am, i have the impression that before this was working as expected by chance, i wanted something simple, sass becomes a mess if you want to handle to much complexity.

src/mixins/typography/mixins.scss
planctus: well, i see that using mobile-xs instead of mobile when the first one is not defined makes it work like before, not sure how, but it does, so maybe we can simply update the existing maps defined for the components

emeryro: yes, it makes sense. The smallest display is used by default. Previously it was mobile, now it is mobile-xs.
what I could do is rename it to "mobile" (default), and "mobile-s" (breakpoint s), but then it would not be aligned anymore with the design, so I guess that it is better to check in the component the few cases where we define the mobile font

planctus (approved):

PR #4045, opened 2025-07-09
feat(file): Icon and link attributes overridable.

julien-:
See https://github.com/openeuropa/bootstrap-component-library/pull/600 for the same issue

### DIFF
--- a/src/implementations/twig/components/file/__snapshots__/file.test.js.snap
+++ b/src/implementations/twig/components/file/__snapshots__/file.test.js.snap
@@ -951,6 +951,7 @@ exports[`File With translation renders correctly 1`] = `
         <a
           aria-labelledby="ecl-file-with-translation-link-label ecl-file-with-translation-title ecl-file-with-translation-lang"
           class="ecl-link ecl-link--standalone ecl-link--icon ecl-file__download"
+          extra-download-attribute=""
           href="/example"
           id="ecl-file-with-translation-link"
         >
@@ -963,6 +964,7 @@ exports[`File With translation renders correctly 1`] = `
           <svg
             aria-hidden="true"
             class="ecl-icon ecl-icon--fluid ecl-link__icon"
+            extra-download-icon-attribute=""
             focusable="false"
           >
             <use
@@ -1079,6 +1081,7 @@ exports[`File With translation renders correctly 1`] = `
               aria-labelledby="ecl-file-with-translation-translation-1-link-label ecl-file-with-translation-title ecl-file-with-translation-translation-1-lang"
               class="ecl-link ecl-link--standalone ecl-link--icon ecl-file__translation-download"
               download=""
+              extra-translation-item-attribute=""
               href="/example#es"
               hreflang="es"
               id="ecl-file-with-translation-translation-1-link"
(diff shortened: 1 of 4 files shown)


planctus:
Hello @julien- , you might have noticed that ECL is not exactly as BCL, we don't use drupal attributes here, for instance.
But in any case at the moment is already possible to provide extra_attributes for the main download link in the file template and its icon as well as for the ones for the translation items, when you see this:
`      {% set _download_link_attributes = _download.extra_attributes|default([])|merge([
        { name: 'id', value: _download_id },
        { name: 'aria-labelledby', value: _download_label_id ~ ' ' ~ _title_id ~ ' ' ~ _lang_id },
      ]) %}`

we are taking the extra_attributes defined in the download object and we merge ours with those. 
The same logic it's used for the main download link and the ones for the translations item.

Same for the icon that this link contains:
`        icon: _download.icon|default(_download.icon|default({}))|merge({
          name: 'download',
          size: 'fluid'
        }),`

I created a pull request for you to see this, attributes are added in the main download link and icon and in one of the translation items link and relative icon, you can see those rendered in the test snapshot of the file:
https://github.com/ec-europa/europa-component-library/pull/4046/files#diff-3431da0e0c1f19eb23af458d9bcf13ab5498d53888a9f4c64541b377d3ca83ec

julien-:
Hello @planctus
Thank you for this quick answer. Effectively no Drupal attributes, sorry for that. I will update my PR.

My needs are:
- Open a file in a new tab : `download="false" target="_blank"` 
- Be able to change the icon
- Be able to change the label

So it should be appreciable to be able to override the default attributes, not be overridden by them.

planctus:
@julien- , now i understand better what you are trying to achieve.
So, some the things you want to customise have been changing in ECL during time, what you have in v4 is the result of decisions made that would aim to standardise the behavior in all the websites using the library, for instance the use of a certain icon or the opportunity of opening a link in a different tab or not.

But let's focus on the technical aspects, you don't really need to patch the ECL templates or even ask for ECL to use your patch to do some of the things you need to do, i'm talking about the latest version in v4 as you created this pr against the v4-dev branch.

1. Open a file in a new tab : download="false" target="_blank"

There is a parameter, named `download_attribute` which is set by default to `false` and we use this to conditionally render the download attribute:
`      {% if _download_attribute %}
        {% set _download_link_attributes = _download_link_attributes|merge([
          { name: 'download' },
        ]) %}
      {% endif %}`

so if you're getting it is because your application is explicitely passing this parameter with a TRUE value.
For adding the target attribute you can pass it as an extra_attribute as it is shown in the pr i mentioned earlier.

2. Be able to change the icon

This is not currently possible via the template in v4 as it is not in v5 either, you see that the logic is that ECL is only forcing certain properties that are relevant for the UI and are not supposed to be changed, in this case the icon size and its name.

3. Be able to change the label

Labels, as any text, are not part of ECL, so it's again your application setting a certain label, the template is expecting that from you.
Thanks.

julien-:
@planctus 

Big THANK YOU for this detailed answer.

Unfortunately due to short delay I had to do the quick & dirty solution : override locally the templates. I let you investigate on a flexible ECL solution. Hope you will for the future. Thank you again

PR #4053, opened 2025-07-15
feat(fact-figures): Small tweaks to simplify the webcomponents implementation - FRONT-4972

planctus:


emeryro (commented):
This is fine, could you just add these changes to the migration doc, as it was done for the accordion?

emeryro (approved):

PR #4057, opened 2025-07-17
feat(spotlight): add component - FRONT-4962

emeryro:
New component to display text on an image
This is also the first ECL component using container queries


src/components/spotlight/spotlight.html.twig
```diff
@@ -0,0 +1,146 @@
...
+{% set _color_mode = color_mode|default('') %}
+{% set _picture = picture|default({}) %}
+{% set _video = video|default({}) %}
+{% set _font_size = font_size|default('m') %}
+{% set _full_width = full_width|default(false) %}
+{% set _header = header|default('') %}
+{% set _title = title|default('') %}
+{% set _path = path|default('') %}
+{% set _credit = credit|default('') %}
+{% set _has_anchor = has_anchor|default(true) %}
```
planctus: this needs to use the null coalescence operator, otherwise it will always be true
```suggestion
{% set _has_anchor = has_anchor ?? true %}
```

src/components/spotlight/spotlight.html.twig
```diff
@@ -0,0 +1,146 @@
...
+    - "value" (optional) (string) Attribute value, eg: 'data-test-1'
+#}
+
+{# Internal properties #}
+
+{% set _css_class = 'ecl-spotlight' %}
+{% set _extra_attributes = '' %}
+{% set _color_mode = color_mode|default('') %}
+{% set _picture = picture|default({}) %}
+{% set _video = video|default({}) %}
```
planctus: why do we have a video param if we don't use it..? :)

src/components/spotlight/spotlight.html.twig
```diff
@@ -0,0 +1,146 @@
...
+{% set _sr_play = sr_play|default('') %}
+{% set _sr_pause = sr_pause|default('') %}
+
+{# Internal logic - Process properties #}
+
+{% if _color_mode is not empty %}
+  {% set _css_class = _css_class ~ ' ' ~ 'ecl-color-mode--' ~ _color_mode %}
+{% endif %}
+
+{% if _font_size is not empty %}
```
planctus: we set a default value for this, so it cannot be empty..

src/components/spotlight/spotlight.html.twig
```diff
@@ -0,0 +1,146 @@
...
+  {% if _header is not empty %}
+    <div class="ecl-spotlight__header">
+      <div class="ecl-container">
+        {{- _header -}}
+      </div>
+    </div>
+  {% endif %}
+
+  {# Link #}
+  {% if _path is not empty %}
```
planctus, line 79: is this really how it is intended..? Fully clickable but not on the "header..? Anyway, it's just a personal feeling, but i don't like the idea of an <a> tag wrapping the whole thing, i would prefer defining it as an element and then extend its area to the full width and height needed, probably, but it's just a preference, you can ignore it.

src/components/spotlight/spotlight.scss
```diff
@@ -0,0 +1,185 @@
...
+    map.get($spotlight, 'credit', 'font'),
+    $container: 'ecl-spotlight'
+  );
+}
+
+.ecl-spotlight__anchor {
+  display: none;
+}
+
+// Full width
```
planctus, line 129: we need to invert this when in RTL

src/components/spotlight/spotlight.scss
```diff
@@ -0,0 +1,185 @@
...
+    height: map.get($spotlight, 'header', 'height-tablet');
+  }
+
+  .ecl-spotlight__anchor {
+    align-self: flex-end;
+    background-color: map.get($spotlight, 'anchor', 'color');
+    display: block;
+    height: 73px;
+    justify-self: flex-end;
+    mask-image: url("data:image/svg+xml;utf8,<svg width='161' height='73' viewBox='0 0 161 73' xmlns='http://www.w3.org/2000/svg'><path d='M147.353 50.6247L0.389648 62.8649V67.9293V72.9936L160.999 62.8649L159.447 0.709961H149.526L147.353 50.6247Z' fill-rule='evenodd' /></svg>");
```
planctus, line 169: nice, but i checked on browserstack and in safari the anchor is not visible at all. Ah, no, actually it's not initially visible because it's white, it's only the position that is wrong, is right below the banner instead of being at its bottom right.

src/components/spotlight/spotlight.story.js
```diff
@@ -0,0 +1,217 @@
...
+  return argTypes;
+};
+
+const prepareData = (data, args) => {
+  correctPaths(data);
+  const clone = JSON.parse(JSON.stringify(data));
+  Object.assign(clone, args);
+
+  if (!args.show_header) delete clone.header;
+  if (!args.show_anchor) clone.has_anchor = 0;
```
planctus: ahh, nice, this way you bypass the problem of the template using default(true), once you'll use the ?? you will be able to set this to false ;)

src/components/spotlight/spotlight.story.js
```diff
@@ -0,0 +1,217 @@
...
+      table: {
+        category: 'Test content',
+      },
+      control: {
+        type: 'boolean',
+      },
+    },
+  };
+
+  if (data.picture) {
```
planctus: guess that you initially worked on the idea of having also a spotlight showing a video. Anyway, an idea, what if we put in this control a select with all the images we have, or some of them, more than a text field, this component is not usable without an image, i'd say that we could also enforce this in the template, like no image nothing rendered...

src/components/spotlight/spotlight.story.js
```diff
@@ -0,0 +1,217 @@
...
+      mapping: {
+        0: false,
+        1: true,
+      },
+      table: {
+        type: { summary: 'boolean' },
+        defaultValue: { summary: 'false' },
+        category: 'Display',
+      },
+    },
```
planctus: we should make this conditional, only when sidebarContent is false, otherwise we reset the value in the data we pass but the control is still there showing "true".

src/components/spotlight/spotlight.test.js
```diff
@@ -0,0 +1,72 @@
...
+  renderTwigFileAsNode,
+  renderTwigFileAsHtml,
+} from '@ecl/test-utils';
+import { axe, toHaveNoViolations } from 'jest-axe';
+
+import bannerDataImage from './demo/data--image';
+
+expect.extend(toHaveNoViolations);
+
+describe('Banner inpage', () => {
```
planctus: maybe we can rename also the test for consistency

src/website/src/pages/ec/components/spotlight/docs/code.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+import { Playground } from '@ecl/website-components';
+
+## Spotlight
+
+<Playground
+  system="ec"
+  selectedKind="components-spotlight"
+  selectedStory="default"
+  showFrame
```
planctus: no harm, but unneeded..

src/components/spotlight/spotlight.scss
```diff
@@ -0,0 +1,185 @@
...
+    $container: 'ecl-spotlight'
+  );
+}
+
+.ecl-spotlight__anchor {
+  display: none;
+}
+
+// Full width
+.ecl-spotlight--full-width {
```
planctus, line 130: i guess that it's not intended to have the "credit" like this: 
image

planctus (changes requested):
Nice, there will be issues because of the title color and because of its length, for sure, we know this from the beginning but the implementation is nice, just found a couple of visual problems and some minor things in the template ;)

src/components/spotlight/spotlight.scss
emeryro: in fact, the css is correct, but the padding is "lost" because of the full screen hack (taking more than the visual space)

planctus: eh, ok, but we have to do something, or do we want to use this as a demonstration of the unreliability of this solution..? :) Being an absolutely positioned element it seems we could use horizontal position to make this work, like right: 1% or so...

src/website/src/pages/ec/components/spotlight/docs/code.mdx
```diff
@@ -0,0 +1,24 @@
...
+  system="ec"
+  selectedKind="components-spotlight"
+  selectedStory="default"
+></Playground>
+
+## Spotlight width
+
+Spotlight placed inside a grid container (`.ecl-container`) will expand to this container width.
+
+To have a full width spotlight, the recommanded way is to place it outside of the grid container.
```
planctus: this is very nice, but if we demo also this component with the "full width" hack by default this sentence here will appear for what it is, something that we say for the sake and the respect of our job but that we don't really expect anyone to follow. To me we should demo the component by default without the wrapping container and only wrap it when "full-width" is chosen or even better it would be not to show the full-width toggle at all, the hack and the parameter are there but we don't necessarily have to advertize them.

emeryro: I agree, we should not have the css hack for this component, or at least not in the demo.
But I see one issue: the aspect ratio is different if the component is "full width" or if it is placed alongside a sidebar. Currently this is checked with the full-width class.
So we could keep that class only for that (but then with no possibility to have the css hack), or we would have to introduce a new parameter for the "sidebar" display

emeryro: ok maybe we don't have to.. I'm confusing myself here, with all the different possibilities and the container queries...

planctus (approved):

PR #4064, opened 2025-07-22
feat(select): rotate select icon - FRONT-4977

emeryro:
Rotate the icon when opening the single select.

Note: this is based on a fairly new css selector, so it may not be visible on some older browsers


planctus:
according to can i use this is not working in safari, any version, but since it works in chrome, firefox and edge, let's use it.. :)

planctus (approved):

PR #4066, opened 2025-07-23
feat(checkbox, radio): put invalid and help text in label - FRONT-4698

emeryro:
Duplicate helper and invalid text in the fieldset label, to make it readable by screen readers


planctus (commented):
Madness, pure madness, but i think that if we hide the elements with display: none it's not just crazy what we are doing, but also useless, no..? So here we should include the screen reader compliant way of visually hiding something, i guess...
We should also put something in the migration doc about this but honestly, we are hacking the screen readers, this is overengineering not accessibility...

emeryro:
This is a shame that this issue is not fixed by the screen reader, indeed. They have it open for years...
These extra labels are correctly hidden on screen, and have a display none on print (as the whole form element is not displayed anyways)

planctus (approved):

PR #4070, opened 2025-07-24
fix(demo): fix inverted demo - FRONT-4994

emeryro:
Some elements were displayed white on white


planctus (approved):
I think it would be better to use color values in these stories, otherwise we have to keep an eye on them everytime we have to update the colors

PR #4074, opened 2025-07-28
chore(content-block): Moving attribute to identify title with links to the link itself - FRONT-5000

planctus:


src/components/content-block/content-block.html.twig
```diff
@@ -88,12 +88,11 @@
   {% endif %}
 
   {% if _title is not empty %}
-    {%- if _title.link is defined -%}
-      {% set _title_attrs = 'data-ecl-title-link' %}
-    {%- endif -%}
     <div class="ecl-content-block__title" {{ _title_attrs }}>
```
emeryro: I guess that we don't need the _title_attrs variable anymore

emeryro (changes requested):
Just some cleanup

emeryro (approved):

PR #4075, opened 2025-07-28
fix(news-ticker): fix padding on EU - FRONT-4996

emeryro:


src/components/news-ticker/news-ticker.scss
```diff
@@ -144,7 +144,7 @@ $news-ticker: null !default;
     background: map.get($news-ticker, 'container-background');
     border-inline-end: map.get($news-ticker, 'border');
     height: map.get($news-ticker, 'desktop-only-height');
-    padding: var(--s-s) var(--s-6xl) var(--s-s) var(--s-m);
+    padding: map.get($news-ticker, 'container-padding-desktop');
```
planctus: funny, looking at the demo it seemed to work, but this should be `padding-container-desktop` based on the name of the key in the maps. :)

planctus (commented):
Uff, there is a small glitch in EC, not introduced by this pr but maybe we can fix it here: 
image

planctus (approved):

PR #4087, opened 2025-08-05
feat(table): small style update - FRONT-5008

emeryro:


planctus:
super minor thing, but i noticed this: 
image which is a bit weird.
Basically the last cell has a margin bottom, we could reset that for the last tr

planctus (approved):

PR #4090, opened 2025-08-07
chore(website): maintenance of typography page - FRONT-4995

emeryro:


planctus (commented):
Oh, wow, is this easier to maintain compared to what we had before..? The tables are completely unreadable and still we would need to manually update any value that has changed, in a js object instead of a markdown table yeah, but i don't know...anyway, you have been working on updating this page, so if you think this is an easier way to do it, let's go...

Well, i'm also thinking that on a general basis, we should avoid the usage of document in the website, we are now in SSR, document is not available there, now this works and doesn't throw errors, so apparently it is only running in the browser, but it would be probably better to get the css values from the source files instead of the DOM.

PR #4093, opened 2025-08-11
fix(carousel): fix tablet padding - FRONT-4993

emeryro:
Fix padding on tablet, when displaying the carousel "full width"

### DIFF
--- a/src/components/carousel/carousel.html.twig
+++ b/src/components/carousel/carousel.html.twig
@@ -121,6 +121,7 @@
       >
         {% include '@ecl/banner/banner.html.twig' with item|merge({
           size: _size,
+          full_width: _full_width,
         }) only %}
       </div>
     {% endfor %}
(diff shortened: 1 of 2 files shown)


src/components/carousel/carousel.html.twig
```diff
@@ -121,6 +121,7 @@
       >
         {% include '@ecl/banner/banner.html.twig' with item|merge({
           size: _size,
+          full_width: _full_width,
```
planctus: mmh..it seems to work fine, but we are nesting an hack inside another hack, is this the way to solve a "non issue", just an inconsistency, i wonder? (also, are we sure that the situation in the banner is correct, that this override in the padding is really needed?) I would probably prefer only having a css rule as we have in the banner that we would possibly remove in the future, since this has no other scope that applying that rule also when the banner is used in a carousel.

emeryro: I proposed another solution, with css. It is not fully clean either, as we have to add a carousel related rule in the banner, but it is either that or we have to redefine the spacings for the different breakpoints in the carousel (and probably forget to update them later)

planctus (approved):

PR #4097, opened 2025-08-13
feat(featured-item): add id and aria-describedby - FRONT-5021

emeryro:
Add aria-describedby on the "Read more" link

### DIFF
--- a/src/components/featured-item/featured-item.html.twig
+++ b/src/components/featured-item/featured-item.html.twig
@@ -2,6 +2,7 @@
 {#
   Parameters:
   - "color_mode" (string) (default: '') Name of the color mode
+  - "id" (string) (default: random)
   - "type" (optional) (can be 'highlight'),
   - "title" (string) (default: ''),
   - "description" (string) (default: ''),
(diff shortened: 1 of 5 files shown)


src/components/featured-item/featured-item.html.twig
```diff
@@ -84,14 +86,22 @@
...
     <div class="ecl-featured-item__item">
     {% if _title is not empty %}
-      <div class="ecl-featured-item__title">{{- _title -}}</div>
+      <div class="ecl-featured-item__title" id="{{ _id ~ '-title' }}">{{- _title -}}</div>
     {% endif %}
     {% if _description is not empty %}
       <div class="ecl-featured-item__description">{{ _description }}</div>
     {% endif %}
     {% if _link is not empty and _link.link is defined and _link.link.label is not empty %}
+      {% set _link_attributes = _link.extra_attributes|default({}) %}
```
planctus, line 95: extra attributes is an array, but fine..

planctus (approved):

PR #4100, opened 2025-08-19
feat(card): update border - FRONT-5023

emeryro:


planctus (commented):
Really really confusing, what is the sense of having tokens only in selected color modes, i wonder, why in the others they are not defined instead. Using the control in the story might give you the impression that is there for no reason...

PR #4105, opened 2025-08-25
feat(right-to-left): move rtl specific css to components - FRONT-5007

emeryro:
Text direction should be specified in the markup, see the "getting started" page or the migration doc for more information


planctus (commented):
Fine for me to remove the standalone css, i just opened https://github.com/ec-europa/europa-component-library/pull/4106 to add an addon that adds the dir attribute instead of using the styles addon. I removed the code that was adding it from there, so if you merge that pr you can then remove the ecl-rtl configuration in the preview.js of the storybook instances.

PR #4111, opened 2025-09-02
feat(form): move helper text, multiple feedback status - FRONT-5042

emeryro:
- Helper text moved below the form element
- Feedback message can now have different status (error, warning, success).

Make sure to read the migration documentation, as the data structure had to change a little


src/components/checkbox/checkbox-group.html.twig
```diff
@@ -4,27 +4,27 @@
...
 #}
 
 {# Internal properties #}
 
 {% set _items = items|default([]) %}
 {% set _name = name|default('') %}
-{% set _invalid = invalid|default(false) %}
-{% set _invalid_icon = invalid_icon|default({}) %}
-{% set _sr_invalid_icon = sr_invalid_icon|default('') %}
+{% set _feedback_type = feedback_type|default('') %}
```
planctus, line 25: i would keep a boolean parameter to "activate" the feedback, to keep consistency with what we used to have, otherwise the users would have to completely change their approach if they were sending the parameter by default and relying on the boolean to make them visible or not.
At this point i would also set a default for the feedback_type "error", in the end this will be the basic usage if any other will ever be used, so this way we make the transition easier, the parameter names will change but their logic will not.

src/components/checkbox/checkbox-group.html.twig
```diff
@@ -4,27 +4,27 @@
...
+{% set _feedback_icon = feedback_icon|default({}) %}
+{% set _sr_feedback_icon = sr_feedback_icon|default('') %}
```
planctus, line 27: why don't we make this just a string with the icon name now that we don't need anymore a path..? And following what i said before i would set the error icon as default for this.

planctus (commented):
Given the fact that no one really felt any need about moving the helper text from where it was, the overall result is that now it looks ugly when you have the helper text together with a feedback, fantastic, really..
I mean, we could replace the helper text when there is a feedback, couldn't we..? I know that there have been surely discussions already on this topic but i cannot really say that the final decision was the right one... 
I would maintain anyhow the logic that handles the user data as close as possible to what we had, also in the story this would mean having a boolean to show/hide the additional feedback message.

planctus (approved):

planctus:
It seems that we are missing the icon for the success feedback message, now..

PR #4112, opened 2025-09-02
fix(site-header): fix menu button overlap - FRONT-5044

emeryro:
This can't be tested directly on ECL, but removing the z-index here seems to solve the issue.
I didn't notice any unwanted side effects on the example pages

### DIFF
--- a/src/components/site-header/site-header-ec.scss
+++ b/src/components/site-header/site-header-ec.scss
@@ -62,11 +62,6 @@ $language-list: null !default;
   }
 }
 
-.ecl-site-header__header {
-  position: relative;
-  z-index: 54;
-}
-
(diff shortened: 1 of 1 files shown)


planctus (approved):
Cool, i checked and this z-index was needed for the shadow set on the element which was removed here: https://github.com/ec-europa/europa-component-library/commit/bf0d0e2f05f073fb89c95d7f5ee26f9d03101ec6#diff-708d42c77ff51689b095019cc9da0f76306a34f6a99bd7f7a798854172d2024a
so fine to remove it, but i wouldn't exclude that we will be asked to reintroduce some similar styling in the future and therefore also have the need to set a z-index on this element.

PR #4113, opened 2025-09-02
feat(icons): use ecl family - FRONT-5047

emeryro:
Use the new ecl family for Webtools icons

Attention points for some icons:
- corner-arrow: renamed to corner-arrow-up
- error: renamed to error-filled
- information-round: removed
- information: renamed to information-filled
- notification-active: removed
- solid-arrow: renamed to solid-arrow-up
- sort-carets: removed
- spinner: removed
- success-filled: new icon
- success-outline: new icon
- warning-round: removed
- warning: renamed to warning-filled


emeryro:
Note: components modal, notification and form elements are updated in dedicated pull requests

planctus:
do we have a plan about what to do with the removed icons..? Normally we cannot do it since they might and they probably are already being used, so in this case we should have a mapping to an existing icon that ewpp could use to make the old one still work.

emeryro:
As this is a new family, all the previous icons are still available in the "default" family.
So users / EWPP should be able to continue using them by setting the family to '' for these specific icons. 
All this thing hasn't really been discussed with EWPP so far, so we can propose this idea...

planctus:
and who is going to maintain the icons not in the ECL family..? I mean, our icons will evolve with time, isn't this going to create a lot of redundancy and inconsistencies..? Anyway, one thing is to map a value to a different one, another is to conditionally set a value for a different field, let's agree this with them to prevent increasing the mess that obviously resulted from this irresponsible decision to give away our icons.

PR #4114, opened 2025-09-03
feat(banner): underline link on hover - FRONT-5046

emeryro:


planctus:
Can we show this to Regine before merging it..? I don't get it, it's clear that we create an ambigous experience, clicking on the banner doesn't produce a redirection to the link page so why is the link getting the same style as when it can actually be clicked and followed..? It clearly gives the impression that the whole banner can be clicked and since it's not the case, the next step to me will be to receive complains about the fact that this is not possible but this would not be doable since the video can be clicked to pause it.

PR #4124, opened 2025-09-11
feat(list-illustration): Adding icon-list and number-list variants - FRONT-5062

planctus:


src/components/list-illustration/README.md
```diff
@@ -18,6 +18,9 @@ npm install --save @ecl/list-illustration
   - "icon" (associative array) (default: {}): predefined structure for Icon component
   - "media_size" (string) (default: 'm'): size of the media (square picture or icon). Could be 's', 'm' or 'l'
 - **"column"** (number) (default: 1): number of columns (1 column means vertical display)
+- **"icon_list"** (boolean) (default: false) Renders as icon list
+- **"number_list"** (boolean) (default: false) Renders as number list
+- **"divider"** (boolean) (default: false) Seprator for items of an icon list
```
emeryro:
```suggestion
- **"divider"** (boolean) (default: false) Separator for items of an icon list
```

src/components/list-illustration/list-illustration-item.html.twig
```diff
@@ -96,5 +100,9 @@
   {% endif %}
   </div>
 </li>
-
+{% if divider and not last %}
+  <li class="ecl-list-illustration__divider">
```
emeryro: Couldn't we set a css border, instead of rendering a full `<li>`  for this?
The `last` could be used to add a css class to the last item, instead of relying only on twig

src/components/list-illustration/list-illustration.story.js
```diff
@@ -513,3 +559,48 @@ VerticalIcon.argTypes = getArgTypes(dataListIllustrationIcon, 'vertical-icon');
...
+};
+IconList.argTypes = {
+  ...getVariantArgTypes(),
+  icon: {
+    control: {
+      type: 'select',
+    },
+    options: ['check-bold', 'arrow-right-bold', 'close-bold'],
+  },
+  divider: {
```
emeryro, line 591: I'm quite sure that we will receive requests to have divider also on the counter...
At least we would be able to do it easily

src/components/list-illustration/list-illustration.scss
```diff
@@ -290,3 +298,146 @@ $list-illustration: null !default;
...
+.ecl-list-illustration__divider {
+  padding: 0;
+  margin: 0;
+
+  .ecl-separator {
+    max-width: map.get($list-illustration, 'divider-max-width');
+    padding-block-end: var(--s-s);
+  }
+
+  @include breakpoints.up('m') {
```
emeryro, line 328: Open question here: shouldn't we use container queries instead? List with illustration could be placed in smaller container

src/themes/ec/variables/_list-illustration.scss
```diff
@@ -4,9 +4,13 @@
 $list-illustration: (
```
emeryro, line 4: There are still fonts using static size definition in the component. If we have time, it may be worth using the responsive fonts for them

src/components/list-illustration/list-illustration-item.html.twig
planctus: well, yes, in the end i had to, but i wanted to use once this silly component we introduced.. :)

src/components/list-illustration/list-illustration.scss
planctus: So, i didn't because in the only example available in figma with the component placed in a layout with columns, a media query is used, i agree that we are progressively switching, but let's do it consistently in all the component here, in case.

src/themes/ec/variables/_list-illustration.scss
planctus: true, but better not to introduce other changes, i actually restored the previous font for the decription, only set the responsive one in the new variants.

planctus:
New functionalities have been requested in the meantime, now it is possible to:

- have the numbering of the list continue from the previous one
- set a specific number as the start of the counter

In the old variants with icons the icon can be placed on the left instead of being on top

src/components/list-illustration/list-illustration.scss
```diff
@@ -5,11 +5,17 @@
...
 @use '@ecl/grid/mixins/breakpoints';
+@use '@ecl/mixins-typography/mixins';
 
 // Exposed variables
 $theme: null !default;
 $list-illustration: null !default;
 
+// Intiialize the counter
+body {
+  counter-reset: item-counter;
```
emeryro: we should use a namespace for the counter, just in case

src/components/list-illustration/list-illustration-print.scss
```diff
@@ -165,3 +165,40 @@ $list-illustration-print: null !default;
     }
   }
 }
+
+/* stylelint-disable no-descending-specificity */
+.ecl-list-illustration--icon-list {
+  .ecl-list-illustration__icon {
```
emeryro, line 184: on print, the icons are placed above the content, not sure if this is on purpose

emeryro (changes requested):
With the parameter to put the icon "inline", we are able to recreate something very similar to the icon list, by using the vertical icon. 
This will surely not help to bring some clarity to the page building...

Anyways, I just have a small remark for the counter and print

src/components/list-illustration/list-illustration-print.scss
planctus: ah, yes, that was due to the latest changes in the markup. About the general confusion related to this component it's crazy, it's like we have all the matrix of all the different possibilities made available in the most confusing way..i know, this was a late request by Vlad, coming from a request by the editors when the "icon list" was not available, it made sense before now probably much less...

emeryro (approved):

PR #4134, opened 2025-09-18
feat(modal): put icon in header content - FRONT-5076

emeryro:
Requested to improve accessiblity


planctus (commented):
Do we want to take this chance to fix also the position of the background in the warning icon: 
image
Hopefully no one is really using this option, but we demo this..

planctus (approved):

PR #4140, opened 2025-09-25
fix(spotlight): container queries for picture - FRONT-5098

emeryro:
The picture tag, used by EWPP to configure images set, do not support container queries.
This PR contains a way to mimic the container queries behavior for this situation


planctus:
mmh..no, i mean...like this you see everything, meaning the first image and the fact that it gets replaced, it's really bad.
At least we should hide the image as long as it is not replaced or "confirmed" by the script but to me what we are doing is simply wrong, it's unfortunate that we tried to use container queries on this component in particular, being this based on the picture tag, but better to have some unpredictability in the image selection by the browser than a situation where we know we are going to significantly delay the image loading, especially because we talk about big images, and so a big delay as a consequence..

emeryro:
The thing is, that would be more than "some unpredictability in the image selection by the browser". Every image has a different aspect ratio (design...), going from 1:1 to 3:1. 
So we cannot really allow a wrong image to be loaded, or it would be badly cropped...
I updated the script to use a placeholder (plain color) until the image is loaded

planctus:
Not sure what has changed, Romain, i still clearly see the first image and then the loading of the replacement one. I guess that we could set a visibility none via css and then reset it via js but to me we can push this back to where the issue is coming from, we cannot fix this with a workaround, it's an html limitation, but pretty understandable, i see two possibilities: 

1. that the design changes limiting the aspect ratio differences
2. that we move back to media queries entirely, this component is not supposed to be in any random container, most of the times it will be using the whole layout and potentially the whole viewport, so maybe this is not the component where container queries might offer the highest benefit, and actually, for how unfortuntate that is, if you cannot provide different images in a reliable way then container queries do not seem to be usable in relation to elements that are mainly containing a media.

PR #4150, opened 2025-10-02
feat(indicator): add screen reader label - FRONT-5092

emeryro:


planctus (commented):
Beside the relevance of this pseudo component and the real chance that this parameter will ever be used, we demo it also in the link story, so maybe we can add the control also there

planctus (approved):

PR #4153, opened 2025-10-03
feat(page-header): Page header not expandable - FRONT-5099

planctus:


src/components/page-header/page-header-expandable.html.twig
```diff
@@ -35,13 +35,19 @@
...
 {% set _separator = separator|e('html_attr')|default('–') %}
 {% set _extra_attributes = '' %}
 {% set _toggle_extra_attributes = toggle_extra_attributes|default([]) %}
 {% set _css_class = "ecl-page-header-expandable" %}
 
 {# Internal logic - Process properties #}
 
+{% if _lists is empty and _panel_content is empty %}
+  {% set _expandable = false %}
+  {% set _css_class =  _css_class ~ ' ecl-page-header-not-expandable' %}
```
emeryro: just a remark for the naming. We end up with things like `ecl-page-header-expandable ecl-page-header-not-expandable`, which is quite confusing. 
We may find something cleaner, using BEM, for instance `ecl-page-header-expandable--static`,  `ecl-page-header-expandable--simple`, `ecl-page-header-expandable--no-child`, ...

planctus: ok, i will change it, to be honest i did it on purpose, they requested an expandable, they chose the name expandable and then it turns out that it's only expandable if the user wants it like that.. :)

emeryro (approved):
this is fine for me.
If we want to be really consistent, e could rename the control `expandable` to `toggle`, to match the css class, but this is internal so it doesn't affect the patch release

PR #4154, opened 2025-10-03
chore(vertical-flow): Generic solution using grid display - FRONT-4957

planctus:
Ways to imlement vertical flow:

- separated lists of items, side by side
- css columns
- grid, with vertical fow

The first two are not usable, there is no way to get a structured layout, with equal height for the items, css columns is also potentially distributing the items in a kind of unpredictable way

Css grid is nice, it does everything right if you tell it how many rows it should create, it gives you even the chance to choose whether to have equally tall rows or not but the way it fills the columns is not "balancing" them, it is filling them one by one so you have four use cases where you get an empty column

-  3 columns: 4 items (2 rows)
- 4 columns: 5 items, 6 items (2 rows) 9 items (3rows)

I've tried to hack this, i positioning absolutely the last item, it worked nice but you loose the alignment with the other items in the row, the items can be taller or smaller than the others, so i haven't implemented it, in the end.


emeryro:
Nicely done!
For the problematic cases, Im not sure what is expected exactly. The browser is doing more or less what it should do (filling column first).
Worse case, if these are only a few identified situation we could handle them manually with the mixin

emeryro:
If we move forward here, I guess that we should add the controls for the column and flow for the "icon" and "number" list

planctus:
Done, i've also updated the template to apply by default the vertical sort in case of a number list with columns.

PR #4155, opened 2025-10-06
feat(fact-figures): group title and value - FRONT-5091

emeryro:


planctus (commented):
We changed the layout when printing the component, let's add the display block to the title also there

PR #4156, opened 2025-10-06
feat(menu): update simple menu styles - FRONT-5073

emeryro:


src/components/menu/menu.scss
```diff
@@ -525,12 +553,14 @@ $menu: null !default;
...
-      font: map.get($menu, 'desktop', 'item-font');
-      line-height: map.get($menu, 'desktop', 'link-line-height');
-      padding: map.get($menu, 'desktop', 'link-padding');
+      min-height: map.get($menu, 'desktop', 'link-min-height');
+      padding-block: map.get($menu, 'desktop', 'link-padding-block');
+      padding-inline: map.get($menu, 'desktop', 'link-padding-inline');
       width: auto;
       z-index: 0;
 
+      @include mixins.responsive-font(map.get($menu, 'desktop', 'item-font'));
```
planctus: I find this incredibly confusing, when we use this mixin we should never retrieve a variable meant or named as a specific one for a certain viewport or group of viewports, in this case we have it twice, one using the value from "mobile" and the other from "desktop" and both are set on the .ecl-menu__link selector, i was expecting to see a media query clash in the inspector but there is none, so i am even more confused.. If this is done on purpose might be worth clarifying this.

src/components/menu/menu.scss
```diff
@@ -349,11 +350,12 @@ $menu: null !default;
   color: map.get($menu, 'mobile', 'item-color');
   display: inline-flex;
   flex-grow: 1;
-  font: map.get($menu, 'mobile', 'item-font');
   justify-content: space-between;
   padding: map.get($menu, 'mobile', 'item-padding');
   position: relative;
 
+  @include mixins.responsive-font(map.get($menu, 'mobile', 'item-font'));
```
planctus: so this is the first usage of the mixin on ecl-menu__link using the value for "mobile".

src/components/menu/package.json
```diff
@@ -18,6 +18,9 @@
     "bowser": "2.12.1",
     "focus-trap": "7.6.5"
   },
+  "devDependencies": {
+    "@ecl/mixins-typography": "5.0.0-alpha.16"
```
planctus: just as a reminder, we released alpha.17 in the meantime (the others are patches so not to be used as deps)

planctus:
a question, with the short version of the menu now we can see something that has probably always been there, but i wonder if it shouldn't be corrected: 
image
We set min-width: 100% but with the padding of the element it will always be wider of the menu item, shouldn't we subtract this padding so that it aligns with it, instead?

emeryro:
The menu items themselves have a min-width of 13.56rem currently, so I doubt that we could align that with the first level item.
That seem to be just a coincidence if the dropdown has almost the same width as the item in your screenshot
For smaller or larger labels, that seem fine
image
image

planctus (approved):

PR #4158, opened 2025-10-09
feat(footer): update data markup - FRONT-5105

emeryro:
Use a span instead of a br, to improve user experience with screen readers


src/components/site-footer/__snapshots__/site-footer.test.js.snap
```diff
@@ -32,8 +32,11 @@ exports[`Site Footer EC Core renders correctly 1`] = `
             class="ecl-site-footer__description"
           >
             This site is managed by:
-            <br />
-            [name of the manager of the site]
+            <span
```
planctus, line 35: mmh..i understand this might be better for the screen readers but this way we introduce a dependency, since we are adding a span can't we do the same but using a selector defined in the component styles, instead of using an utility..?

planctus (approved):

PR #4159, opened 2025-10-09
feat(storybook): Addon to show the width of the preview - FRONT-5110

planctus:


emeryro (commented):
the width reported here is the one of the iframe. There is some padding coming from storybook, so it may give wrong result when the designer want to check some specific sizes
image

planctus:
but it's the width of the viewport of the iframe, the one that will trigger the media queries, regardless of the inner padding, that is what it's meant for, checking the different look and feel in the different breakpoints. if we would subtract the padding the width would not match anymore the media queries, instead.

emeryro (approved):

PR #4160, opened 2025-10-09
feat(table): dynamic sort label - FRONT-5093

emeryro:


planctus:
Oh my god, and it wasn't enough the aria-sort on the heading..? To me if we add this we should remove the other, i don't see any aria-label on the button in this page, to me it's a duplicate: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-sort

emeryro:
This was my first impression too, and what I reported in the ticket with Regine. 
The argument is that the labels serve a different purpose.
The aria-sort indicates the current sorting order. The new aria-label indicates what will happen if the user click on the button (so the value is different to the current aria-sort). This is mostly useful because of the aria-sort="none" option, but I agree that the usage is limited

planctus (approved):

PR #4164, opened 2025-10-14
feat(page-header): add new variants - FRONT-5049

emeryro:


planctus (commented):
I've played with the controls, there might be some edge cases, not sure whether we should consider them but basing on our controls it seems there is nothing mandatory:
50/50 with only title
image
50/50 with description only (bottom position)
image
with breadcrumb and no title or description
image

- In the "news" if you show the expandable it has the same background color, i'm not sure the designer ever tried it, i see no example like that on their side.

- It seems that the pages in the website have not been updated yet

PR #4170, opened 2025-10-22
feat(inpage-nav): icon and spacing - FRONT-5122

emeryro:
Remove icon and update spacing


planctus (commented):
we used to have some spacing on top of the title, it was removed here, was that intentional..? When scrolling the title gets attached to the top of the viewport.

emeryro:
This is how it is in Figma, but I will double check

emeryro:
Spacing is back

PR #4172, opened 2025-10-23
feat(footer): add co-owner section - FRONT-5121

emeryro:


src/components/site-footer/site-footer-ec.scss
```diff
@@ -5,15 +5,82 @@
...
+  );
+}
+
+.ecl-site-footer__co-owner-list {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.ecl-site-footer__co-owner-item:not(:first-of-type) {
```
planctus: for the sake of the webcomponents, couldn't we add a class here instead of looking for the first of type..? In cases like this we used .is-first in other components.

src/components/site-footer/site-footer-ec.scss
```diff
@@ -5,15 +5,82 @@
...
+  &:hover {
+    color: map.get($site-footer, 'co-owner', 'link-color-hover');
+  }
+}
+
+@include breakpoints.up('l') {
+  .ecl-site-footer__co-owner-item {
+    display: inline;
+
+    &:not(:first-of-type) {
```
planctus: to be used also here..

src/components/site-footer/site-footer-eu.scss
```diff
@@ -5,15 +5,76 @@
 
 @use 'sass:map';
 @use '@ecl/grid/mixins/breakpoints';
+@use '@ecl/mixins-typography/mixins';
```
planctus, line 8: do we have responsive typography in EU..? Or is it just for "consistency"..?

planctus (commented):
Just a small request for applying a webcomponent "safe" approach...

src/components/site-footer/site-footer-eu.scss
emeryro: this is for consistency, and to be able to copy/paste the css between EC and EU.

planctus (commented):
I guess that the EU css hasn't been updated after making the changes, i am a bit surpised by the color of the links on hover, isn't that too close to the background color?

planctus (approved):

PR #4174, opened 2025-10-23
feat(indicator): update markup and demo - FRONT-5128

emeryro:


planctus (commented):
We have the same controls now in three stories, why don't we put these in the story utils package..?

planctus:
Wouldn't it be better to set a default for the indicator value..? It's not nice that when you enable it it's just a red circle, you have to notice that there is an additional control that appeared where to set a value.
Also, shouldn't we set overflow:hidden on the indicator, or alternatively shouldn't its size increase based on the content it has..?

planctus (approved):

PR #4175, opened 2025-10-27
feat(gallery): cursor, description and aria-label - FRONT-5129

emeryro:
Improve accessibility:
- zoom cursor on hover
- extra zoom icon on focus
- description available to screen readers, except for long caption (still available in the overlay)
- extra aria-label for link


src/components/gallery/gallery-item.html.twig
```diff
@@ -49,6 +49,14 @@
   {% set _css_class = _css_class ~ ' ' ~ extra_classes %}
 {% endif %}
 
+{# Process image alt #}
+{% set _image_alt = _thumbnail.img.alt|default('') %}
```
planctus: this is probably not an issue on our side, but in php i fear that this might throw an error in case _thumbnail.img is not defined, the default protects only the alt, but i see that it was already the case an no one complained, anyway we could still make it "safer" adding a check for for thumbnail.img..

planctus (approved):

PR #4177, opened 2025-10-27
feat(featured-item): inverse markup order - FRONT-5131

emeryro:
Change markup order to improve accessibility. No visual modification


planctus (commented):
Totally unrelated, can we flip the read more icon in RTL..? :)

planctus (approved):

PR #4179, opened 2025-10-28
feat(page-header-expandable): Removing unused params and related styes, not using flex in the header anymore - FRONT-5113

planctus:


src/components/page-header/page-header-ec.scss
```diff
@@ -196,118 +170,59 @@ $page-header: null !default;
...
-    font-size: var(--fs-s);
-    line-height: map.get($theme, 'line-height', 's');
-  }
-
-  .ecl-page-header-expandable__list .ecl-icon {
-    height: var(--s-m);
-    width: var(--s-m);
-    margin-inline-end: var(--s-3xs);
-  }
+[data-ecl-page-header-expandable-header-link],
```
emeryro: we should avoid relying on custom data attributes in the css. eUi is not putting them in their markup for instance
this is for EC, EU, and print

src/components/page-header/page-header.story.js
```diff
@@ -89,26 +40,6 @@ const expandableArgTypes = () => {
...
-      description: 'Separator for strings in the header',
-      table: {
-        type: { summary: 'string' },
-        defaultValue: { summary: '' },
-        category: 'page header expandable',
-      },
-      if: { arg: 'show_page_header_expandable' },
-    },
     header_content: {
       name: 'content of the header (additional)',
```
emeryro: this is no longer an "additional" or "alternative" element

emeryro (changes requested):

src/components/page-header/page-header-ec.scss
planctus: but it's only one way of trying to target those link, the full selector is: 
`[data-ecl-page-header-expandable-header-link],
[data-ecl-page-header-expandable-header-link]:visited,
.ecl-page-header-expandable__header .ecl-link,
.ecl-page-header-expandable__header .ecl-link:visited`

i don't know, i can remove it but it's meant to enhance the possibilites of applying the requested styles to some random markup, alternatively we could style the <a> tag and that would be the best in terms of effectiveness, but we never do that in a component style.

emeryro (approved):

PR #4189, opened 2025-11-05
feat(styles): rich text spacing and example - FRONT-5145

emeryro:
Remove first and last margin of rich text sections.
Also add a page example for rich text content (/story/page-examples-rich-text--default)


planctus:
So we are saying "fine, keep adding an .ecl wrapper on every ckeditor instance"..? :)

Are we sure we really want to put our hands into the inner styles of whatever they might generate inside an .ecl wrapper..?
Even in this example, the first h2 should actually have the margin-top, we reset that and then you need to add a utility on the container, why is it assumed that the heading, if it's the first element in an .ecl wrapper would not need a margin-top? In reality it would probably be better the other way around, the heading being used to set the margin between sections, so wherever an heading appears there would be space on top and at the bottom.

The contradiction is then that we have  huge margin top for headings, so in this example we are basically resetting them all except for the h3s that indeed are creating a misalignment with the 2xl margins set on the wrappers.

I don't know, maybe i understand a bit more why resetting the margin-bottom on the last-child... 
i guess this was based on examples of markup available in ewpp and therefore in the mess of the usage of random html this can be beneficial, but it feels really sinking more and more into quicksand..

src/page-example/page-rich-text/README.md
```diff
@@ -0,0 +1,7 @@
+# ECL Example page - Home
+
+npm package: `@ecl/page-home`
```
planctus: for what it counts, since we don't release these packages, this should be updated :)

src/presets/eu/src/eu-default.scss
```diff
@@ -227,3 +227,24 @@
 .ecl blockquote img:is([class*='ecl-u-']) {
   @extend %ecl-blockquote__image;
 }
+
+// Remove some spacing
+.ecl p:first-child,
```
planctus: well, not that this would change anything, it would just look less repetive in these sass files, up to you, but something like this..? `.ecl :is(h1, h2, h3, h4, h5, h6, p):first-child { }`

emeryro:
To limit the risks, we could have this as an option. By default .ecl would not reset the margins, but we can add .ecl--no-margin, manually added where needed by EWPP/editors...

PR #4190, opened 2025-11-05
feat(tabs): Adding basic tabs functionality - FRONT-5144

planctus:


src/components/tabs/tabs.js
```diff
@@ -330,6 +436,66 @@ export class Tabs {
...
+      }
+      // Toggle active styles and attributes
+      if (tab.id !== tabId) {
+        tab.link.classList.remove(this.activeSelector);
+        tab.link.setAttribute('aria-selected', 'false');
+      } else {
+        tab.link.classList.add(this.activeSelector);
+        tab.link.setAttribute('aria-selected', 'true');
+        if (
+          !tab.link.closest('div').classList.contains('ecl-tabs__item--hidden')
```
emeryro: it may be safer to look for "ecl-tabs__item" instead of just div

emeryro (changes requested):
It works fine, but there is an accessibility warning now that each tab has an id: as the tab items are duplicated in the dropdown, we end up with duplicated ids

emeryro (approved):

PR #4193, opened 2025-11-10
feat(mega-menu): Optional first level only panel with multi column layout - FRONT-5149

planctus:
i added three examples, one with info and featured panels, one with info only and one without any other panel, the number of columns goes from 4 to 2.


src/components/mega-menu/mega-menu.js
```diff
@@ -816,31 +830,37 @@ export class MegaMenu {
...
+            if (remaining > 0) {
+              seeAll.style.top = `${remaining}px`;
+            }
+          }
         }
         if (infoPanel && this.isLarge) {
           infoPanel.style.height = `${height}px`;
         }
-        if (secondPanel && this.isLarge) {
+        if (secondPanel && this.isDesktop) {
```
emeryro: not sure if this is linked to this change, but on tablet display I am not able to see all the content when there is not enough space; last item stays hidden (item 3.5 here)
image

That seems to work fine on v5-dev

src/components/mega-menu/mega-menu.scss
```diff
@@ -1666,8 +1687,60 @@ $mega-menu: null !default;
...
+          column-gap: var(--s-2xl);
+          column-rule: 1px solid var(--cm-border-low);
+          display: block;
+        }
+
+        // Without info
+        &.ecl-mega-menu__mega--no-info {
+          width: 100%;
+
+          .ecl-mega-menu__sublist {
```
emeryro, line 1716: do we really need to have a css selector that specific?
image
couldn't we target this element without all the ">"? Or maybe check the item content with js and add a single class. I usually prefer to rely on css only, but here that looks troublesome for maintenance

emeryro (commented):
a few questions/remarks.
Also, it is worth mentioning this in the migration doc

src/components/mega-menu/mega-menu.scss
planctus: eh, well..the main issue here is that being a menu structure has nested similar chunks of html, so that > becomes a way to limit the items targeted by a selector.
I agree that this doesn't necessarily justify the complexity of certain selectors used by this component, but the problem is that we added styles on top of existing styles and the more we did this the more complex the selector became.
Here in particular the selector that has been added is overriding the style set for the generic element which is already hyper nested, so it had to be likewise specific.
I'd say that maybe one day, out of an iteration on this component, if there will ever be such a situation we could reserve time to try to refactor this css, as well as the js that suffered from the same continuos addition of features and behavioral changes.

src/components/mega-menu/mega-menu.js
planctus: it didn't work in v5-dev, this was due to a wrong logic in limiting the dropdown to the 70% of the available space, now it should be ok.

emeryro (approved):

PR #4197, opened 2025-11-18
feat(spotlight): update focus and gradient - FRONT-5154

emeryro:
- update focus display
- update gradient behind text
- css class added when there is an anchor
- fix print


planctus (commented):
This element was flawed from the beginning, we have underlined the problematics of its design with no avail,  regarding the main scope of this change, to me this is not necessarily improving the readibility and therefore the accessibility, but it's surely looking bad, worse than before.

Maybe one thing that we haven't highlighted enough is also the fact that it's crazy than we might force the user to scroll because we impose a 1/1 aspect ratio in small screens for an element that has text at the bottom, so you might see an image taking the full screen and then only after scrolling you would finally reach the text that it contains, it's insane.

About the pr, the control that has been added should specify that it's only related to the demo, nothing to do with the component itself, and i don't know, maybe we could actually add the same image in aws with different aspect ratios to be used to demo this component, it looks bad when it completely switches image.

emeryro:
The demo images can be improved, for sure. We should receive a proper aws storage to replace the old one we have, so we can change it then.

planctus (approved):

planctus:
We can upload images if we really want to, i can create a dedicated workflow for doing that, i don't think this is something introduce in this pr but there are cases where the text can overlap the graphic sign: 
image
image
Not sure if we want to do something about it or live with this.

emeryro:
Thanks, the overlap was coming from this change indeed.
I also fixed the print

planctus (approved):

PR #4199, opened 2025-11-19
feat(content-block): allow horizontal meta - FRONT-5147

emeryro:
Applied to content block, content item and card


src/components/content-block/content-block.scss
```diff
@@ -96,10 +96,30 @@ $content-block: null !default;
 .ecl-content-block__secondary-meta-item {
   align-items: center;
   display: inline-flex;
-  padding: var(--s-xs) 0;
+  margin-block: map.get($content-block, 'secondary-meta-item-margin-block');
```
planctus, line 99: are we sure that this is the way it was intended..? This reduces the spacing from 1rem to 0.5rem in EC in the vertical sorting too. It is also not clear the reason why we were using and we are still using spacing on top and bottom, this clearly alters the spacing with the elements above and below, both in the vertical and in the horizontal sorting.

emeryro: This is how it is defined in Figma...
image
I didn't get feedback from the designers yet, so let's see what the say

planctus: yes, i saw that, but this is the new, horizontal display, what we are doing is impacting also the vertical one.

The other consideration  is indeed about a flaw in the design, i guess that we might have to bring that to their attention, if we want, otherwise they would just see the values matching the ones they used and they will be happy with them.

emeryro: I got the confirmation from the designers that the vertical is fine as is (with smaller spacing)

planctus (approved):

PR #4201, opened 2025-11-20
feat(link): remove deprecated 'no-visited' - FRONT-5158

emeryro:
Remove deprecated variant. 
Visited links are using the same color as default one in ECL 5.


planctus (commented):
What if we introduce a custom property for the link color..? This way we would not need to override the css to set that color but just the value of the custom property..

emeryro:
You mean adding custom property in the component, like `--ecl-link-color`, `--ecl-link-color-hover`, `ecl-link-color-visited`, ...?
If I recall we considered doing that previously ("component level variables"). 
We can do it, but then it would probably have to be done in other components too

planctus:
mmh..i would define them globally as the others, in the end it's not going to change much, but ideally they should be available in the default css also when the ecl main css is not loaded, so i would place them in custom-properties.scss.
About the impact of this, i guess that we would only have to update the styles overriding the default, which mostly is going to be the new dark variant to be added. 
As long as :visited is going to be the same as the default i would just add two variables, default and hover

planctus:
It seems we still have styles for the no-visited in the inverted variant of the link, line 153 of the link scss, i guess we can get rid of it as well.
So the idea is to have a follow up ticket to update the overrides of the default styles in the other components, right..?

emeryro:
Other link variant are now correctly using the variables.
For the other components, I would rather wait to have the "black link" first, as it would already cover most of the cases. And the black link would need this PR, so let's start with that

planctus (approved):

PR #4202, opened 2025-11-20
fix(description-list): fix separator alignment - FRONT-5157

emeryro:
Center separator vertically for taxonomy.

Note: we couldn't use percent based position, as it would fail when an item is wrapped in two lines


planctus (commented):
Wouldn't be easier using:
 top: 50%;
 transform: translateY(-50%);

we would not have to hardcode any value that might have to be changed if something changes in the typography..

emeryro:
I tried, but it doesn't work when an item is wrapped on two lines
image

planctus:
true, but i wonder if we shouldn't prevent the single taxonomy from breaking in two lines, it's not really readable..

PR #4208, opened 2025-11-24
fix(range): Fixing bubble position when in a container - FRONT-5161

planctus:


emeryro:
This works fine, we should just add this to the migration doc, as the markup/js needs to be modified

emeryro (approved):

PR #4224, opened 2025-12-03
feat(page-info): Add page information component - FRONT-5164

emeryro:
Very simple component to display page information


planctus:
Lots of files and code for a string of text, eh.. :)
The example of the twig include in the README doesn't include the "with { } only", it seems that without it the script to replace the data breaks the content in the storybook panel.

PR #4225, opened 2025-12-03
fix(Inpage-navigation): focus heading programmatically when relative anchor is selected

### LINKED ISSUE #4218: [Accessibility] `Inpage Navigation`: heading should be programmatically focused when scrolled into view
# Description:
When using keyboard navigation with the Inpage Navigation component, focus should move to the heading the user navigates to. Currently, focus is lost and returns to the body element instead.

## Impact:
Creates a significant barrier for keyboard users, particularly blind users with screen readers, as they lose their navigation context and become disoriented.

## WCAG Impact:
Violates 2.4.3 Focus Order and affects user orientation (3.2.3 Consistent Navigation)

## Solution:
Programmatically set focus to the target heading element after navigation completes.

claurennt:
FIxes #4218

### DIFF
--- a/src/implementations/vanilla/components/inpage-navigation/inpage-navigation.js
+++ b/src/implementations/vanilla/components/inpage-navigation/inpage-navigation.js
@@ -516,6 +516,11 @@ export class InpageNavigation {
     // Untrap focus
     this.focusTrap.deactivate();
 
+    if (heading) {
+      heading.setAttribute('tabindex', '-1');
+      heading.focus();
+    }
+
     const eventData = { target: heading || href, e };
     this.trigger('onClick', eventData);
   }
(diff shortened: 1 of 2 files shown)


claurennt:
Hi @emeryro and @planctus ,

it's the first time I open a PR on the project. My code changes do not introduce the issues that make the audit pipeline fail.
Any idea why this happens? maybe I missed something!
Thanks a lot in advance

planctus:
Don't worry @claurennt, the audit step failing is due to dependencies having some vulnerabilities, nothing related to your code, about your pull request we will probably duplicate it on our side in order to be able to get a preview url, for accessibility matters we would involve our accessibility expert asking her to review your solution and if approved we will review this pr technically and merge it if we are happy with it. 
The only thing is that you created it against v4, we are not releasing anymore with regularity on that version, if the solution is good we will apply it to v5 as well. Thanks for contributing  ;)

claurennt:
thanks @planctus for clarifying! 
I have created another PR that points to `v5-dev` 
https://github.com/ec-europa/europa-component-library/pull/4226

PR #4226, opened 2025-12-04
fix(Inpage-navigation): focus heading programamtically when relative …

claurennt:
…anchor is selected


planctus:
i'm closing this since we extended this work in https://github.com/ec-europa/europa-component-library/pull/4238

PR #4228, opened 2025-12-08
feat(datepicker): Reverting styles to match the duet ones - FRONT-5134

planctus:


src/components/datepicker/datepicker.js
```diff
@@ -158,18 +158,18 @@ Datepicker.defaults = {
       'December',
     ],
     monthNamesShort: [
-      'January',
-      'February',
-      'March',
-      'April',
+      'Jan',
```
emeryro, line 161: this change should be mentioned in the migration doc. This would probably require some translation work

emeryro (changes requested):
Just some documentation update

emeryro (approved):

PR #4229, opened 2025-12-08
feat(link): add brand link - FRONT-5138

emeryro:
- Add new variant for link (`branded`), using dark text instead of blue
- Use this variant where relevant in existing components
- Use recently added css variable for links to simplify css


.stylelintrc.cjs
```diff
@@ -18,5 +18,7 @@ module.exports = {
     'declaration-block-no-redundant-longhand-properties': null,
     // Disable checks because automated fixes would be breaking
     'color-function-notation': null,
+    // Allow map.get() and other functions inside interpolation
+    'scss/dollar-variable-no-missing-interpolation': null,
```
planctus, line 22: Well, maybe completely disabling the rule is too much, unfortunately it seems that it can't be configured to prevent the false positives we have when we use map.get to retrieve the value for a custom property we set, though..

planctus (approved):

PR #4231, opened 2025-12-10
docs(ECL): update global documentation - FRONT-5120

emeryro:
Update technical and user-facing documentation for ECL 5


src/website/src/pages/ec/getting-started/index.mdx
```diff
@@ -178,71 +196,185 @@ For example, if you want to auto initialize a `Message`:
...
+2. **Copy the HTML markup** from the component documentation or <Link to="/playground/ec/" label="playground" />
+3. **Paste it into your page** where needed
+4. **Customize** the component, if needed, by modifying data attributes, classes, or content
+
+### Twig Templates
+
+For server-side rendering, ECL provides **Twig templates** for all components, available on npm:
+
+```bash
+npm install @ecl/twig-component-{component-name}
```
planctus:
```suggestion
npm install @ecl/{component-name}
```

planctus: Then we can specify the convention for the template's names, {component-name}.html.twig

src/website/src/pages/ec/getting-started/index.mdx
```diff
@@ -178,71 +196,185 @@ For example, if you want to auto initialize a `Message`:
...
+```
+
+Browse available templates at [npmjs.com/org/ecl](https://www.npmjs.com/org/ecl).
```
planctus:
```suggestion
Browse available packages on [npm](https://www.npmjs.com/org/ecl).
```

src/website/src/pages/ec/getting-started/index.mdx
```diff
@@ -178,71 +196,185 @@ For example, if you want to auto initialize a `Message`:
...
+
+---
+
+## Advanced Usage
+
+### Individual Component Packages
+
+While presets are the easiest way to get started, ECL also publishes individual packages for each component, tool, and resource. This is useful for:
+
+- Custom builds with only the components you need
```
planctus: mmh..this is not really supported, we do not compile the scss source of the component itself, so you cannot grab it from the package folder directly and we have now global styles with the custom properties so in order to use components individually the users would have to fetch also the theme packages and get the custom properties css (for EC also the color modes one)

docs/conventions/components.md
```diff
@@ -34,10 +34,90 @@ Please refer to the [dedicated conventions section regarding SCSS](./scss.md).
...
+export const Default = () => Demo;
+export const Primary = () => /* ... */;
+```
+
+## Tests
+
+Each component should have a `.test.js` file with:
+
+- **Snapshot tests**: Ensure HTML output remains consistent
+- **Unit tests**: Test JavaScript behavior (if component has `.js` file)
```
planctus: we do not have unit tests, actually..

docs/conventions/linting.md
```diff
@@ -1,19 +1,18 @@
 # Linting
 
-ECL uses [ESLint](https://eslint.org/) to ensure standartization within JavaScript and JSX files within this repository.
+ECL uses [ESLint](https://eslint.org/) to ensure standardization within JavaScript and JSX files within this repository.
```
planctus:
```suggestion
ECL uses [ESLint](https://eslint.org/) to ensure standardization within JavaScript and JSX code within this repository.
```

docs/javascript.md
```diff
@@ -6,12 +6,36 @@ Components do not depend on jQuery and provide consistent APIs which can be mana
 
 ## How to use
 
-First, you need to include the JavaScript file of `ecl-ec.js` or `ecl-eu.js` provided in the [latest release package](https://github.com/ec-europa/europa-component-library/releases). This file contains a JavaScript module called `ECL` which is an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) built by the [`ecl-builder` utility](https://www.npmjs.com/package/@ecl/builder).
+There are two ways to use ECL JavaScript: via global IIFE bundle or ESM imports.
 
-This means that when you include the `ECL` library in your pages, you will have a global called `ECL` which contains the components' factory functions.
+### Option 1: Global IIFE (Browser)
```
planctus: This section is confusing, we have two "global" exports, one in common js and one in ESM, we should mention both here, the fact that the common js export is an IIFE is pretty irrelevant in this context and can be ignored, probably.
In both cases those are meant to be used by the browser with a script tag (the only difference is that you have to use type="module" for the ESM script) and they set an ECL object in the window, we should recommend the use of the ESM one which is also the one we use in storybook.

Then we can also mention the use of the modules for the single component's class in other scripts but for how we define the preset pacakge, there is no export of the single classes in the esm file so they would now need to be imported from the component's package, not from the preset.

docs/javascript.md
```diff
@@ -20,34 +44,129 @@ You can get the ECL version you are using running `ECL.version` in the console o
...
 ```js
-var instance = ECL.components.get('yourElement');
+const element = document.querySelector('[data-ecl-accordion]');
+const instance = ECL.components.get(element);
+```
+
+**ESM:**
+
+```javascript
+import { components } from '@ecl/preset-ec';
```
planctus: mmh..this doesn't exist, the way to retrieve the map of instances is still from the ECL object, `ECL.components` so this differentiation between "global" and ESM to mean individual here is not really needed.

docs/javascript.md
```diff
@@ -20,34 +44,129 @@ You can get the ECL version you are using running `ECL.version` in the console o
...
+});
+
+accordion.on('onClose', (event) => {
+  console.log('Accordion closed', event);
+});
+```
+
+**Available events** vary by component. Common events include:
+
+- `onInit` - Component initialized
```
planctus: we don't have onInit or onDestroy, i think this list can be removed, we do not have "common" events used by all the classes at the moment, but it might make sense to do it..

docs/presets.md
```diff
@@ -11,14 +11,133 @@ Additional preset (included in the main preset):
...
+### Option 3: Direct Download
+
+Download the [latest release package](https://github.com/ec-europa/europa-component-library/releases) and include the files in your project:
+
+```html
+<!-- CSS -->
+<link rel="stylesheet" href="/path/to/ecl-ec.css" media="screen" />
+
+<!-- JavaScript -->
+<script src="/path/to/ecl-ec.js"></script>
```
planctus: the ESM should be added to as an option, probably the first one.

planctus (commented):
Thanks, very much needed update, beside some minor tweak only the Js section needs maybe some refactoring.

emeryro:
Thanks for the feedback. I updated the javascript documentation, and removed duplicated information

src/website/src/pages/ec/getting-started/index.mdx
```diff
@@ -178,71 +226,162 @@ For example, if you want to auto initialize a `Message`:
...
+2. **Copy the HTML markup** from the component documentation or <Link to="/playground/ec/" label="playground" />
+3. **Paste it into your page** where needed
+4. **Customize** the component, if needed, by modifying data attributes, classes, or content
+
+### Twig Templates
+
+For server-side rendering, ECL provides **Twig templates** for all components, available on npm:
+
+```bash
+npm install @ecl/{component-name}.html.twig
```
planctus: mmh..no, the package to be installed is @ecl/{component-name}, then {component-name}.html.twig is the name of the twig file

docs/javascript.md
```diff
@@ -20,34 +20,70 @@ You can get the ECL version you are using running `ECL.version` in the console o
...
+const element = document.querySelector('[data-ecl-accordion]');
+const accordion = new ECL.Accordion(element);
 accordion.init();
 ```
 
-In both cases the instance will be available in the main ECL object, in the components Map, from here an existing instance can be retrieved to further update it, for instance running `destroy()` and `init()` again.
+If the component is manually initialized, make sure to also clean up component instances when they're no longer needed:
 
-```js
-var instance = ECL.components.get('yourElement');
```
planctus, line 42: this like is pretty much needed, otherwise the user has no idea where to retrieve the instance from, maybe we can use a more modern `const` instead of `var`, though
Better also to use yourHtmlElement, might be confused for a selector otherwise

planctus (approved):

PR #4233, opened 2025-12-10
feat(forms): Aligning height of some form elements - FRONT-5173

planctus:


src/components/datepicker/datepicker.scss
```diff
@@ -43,10 +43,10 @@ $text-input: null !default;
   * duet styling
   */
 .duet-date__input {
-  height: 2.75rem;
+  height: map.get($text-input, 'height');
```
emeryro: we could put the form input height in a css variable, and reuse it everywhere

emeryro (approved):

PR #4237, opened 2025-12-11
feat(list-illustration): Using ordered list for number list variant - FRONT-5178

planctus:


emeryro (changes requested):
Now that we have an ordered list, we should also convey the starting number to it

planctus:
Well, i did it, but in reality it will not really change much, we do offer the possibility to set the starting number manually but i don't expect that to be used, instead what might be used is the continuity offered by the css counter, so that if you have two lists the second will continue the first, automatically, but that is not happening in the ordered list so in that case there will be a different numbering.

emeryro (approved):

PR #4238, opened 2025-12-11
fix(inpage): Focus target element when relative anchor is selected - FRONT-5183

planctus:


src/components/inpage-navigation/inpage-navigation.scss
```diff
@@ -368,6 +368,12 @@ $inpage-navigation: null !default;
   }
 }
 
+.inpage-navigation__heading--active {
```
emeryro: The focus works fine, but I wonder if that's really necessary to visualize it with an outline

planctus: i saw that regine reviewed this, so i made some changes, including the remvoal of the outline on the focused heading:

- using arrows in the list now goes into a loop, arrow down on the last brings to the first, arrow up on the first brings to the last
- when entering the list, if there is an active item, that one gets the focus
- when using ESC, the dropdown is closed, the focus goes to the toggle
- when using TAB, if the focus is going to go out of the inpage, the dropdown is going to be closed

This is more or less following what demoed here: https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-links

emeryro (approved):

PR #4243, opened 2025-12-18
fix(breadcrumb): fix display of home on mobile - FRONT-5171

emeryro:
Fix potential empty breadcrumb on mobile.
If the current page is a direct child to the home, the "home" link would be displayed


src/components/breadcrumb/breadcrumb.scss
```diff
@@ -128,13 +128,29 @@ $breadcrumb: null !default;
   display: none;
 }
 
+// Show first link and icon on mobile when there are only 2 items (Home + Current Page)
+// stylelint-disable-next-line no-descending-specificity
+.ecl-breadcrumb__segment:first-of-type:nth-last-of-type(2) {
```
planctus, line 133: nice selector.. ;)

planctus (approved):
Just a question, we do show also the icon, inverting its direction, shouldn't we hide it instead? I mean,  in any other case that arrow is probably meaning that there are items before the one being shown, in this case we are showing the first so there is none.

emeryro:
To me the icon is more to indicate that we are going "back" one level, which would be the case also for the home.
And without it we would just have a floating link, not very explicit that it is a breadcrumb

PR #4246, opened 2026-01-07
feat(news-ticker): new design - FRONT-5194

emeryro:
New design including 
- more content displayed on mobile (5 lines instead of 2)
- handle of single news 
- default icon displayed if none is provided
- color mode awareness

Usage of small images for news is deprecated (still working though), using icons is the recommended way.


planctus (commented):
The font size in EU increased a lot, is that intentional..?
Because of that, an issue with the showcase page in the website became more visible, since we use overflow: hidden iframe resizer gets fooled and it considers the list full height, we should set a fixed height for the iframe to fix this.

src/components/news-ticker/package.json
```diff
@@ -19,6 +19,9 @@
   "devDependencies": {
     "@ecl/grid": "5.2.0"
   },
+  "devDependencies": {
+    "@ecl/mixins-typography": "5.0.0-alpha.22"
```
planctus: let's align this to the latest version before merging this

planctus (commented):
Let's just align that dependency and maybe we could make the height in mobile dynamic, based on the content, it seems weird to have it that tall if the content is short but i guess it was designed like this and in this case fine as it is...

PR #4252, opened 2026-01-12
feat(accordion): use improved markup; no js - FRONT-5192

emeryro:
Change accordion markup to use `<details>` and `<summary>`.
Javascript is no longer needed and has been removed.
New optional parameter `name` to group items (only one item open at the time)


planctus:
Frankly it seems crazy to me to change the markup of one of the components that ECL has always shipped in this moment, couldn't we propose to postpone this to ECL 6..?

planctus (approved):

planctus:
I was about to merge it but i checked on mac, safari and i get this: 
image
but only with the latest safari on tahoia, the most recent version of MacOs and in iphone and ipad still with safari

PR #4255, opened 2026-01-13
docs(storybook): Add source HTML markup - FRONT-5072

emeryro:
Add tabs in storybook to display both the initial markup and the one after Webtools' icon injection

### DIFF
--- a/src/playground/addons/storybook-addon-code/HTMLMarkup.jsx
+++ b/src/playground/addons/storybook-addon-code/HTMLMarkup.jsx
@@ -51,19 +51,39 @@ const StyledSyntaxHighlighter = styled(SyntaxHighlighter)(({ theme }) => ({
 function HTMLMarkup({ active, markup, originalMarkup }) {
   const [activeTab, setActiveTab] = useState('source');
 
-  const beautifyHtml = (html) => {
+  const beautifyHtml = (html, options = {}) => {
     return beautify(html, {
       indent_size: 2,
-      max_preserve_newlines: -1,
-      preserve_newlines: false,
+      indent_char: ' ',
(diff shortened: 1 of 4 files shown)


src/playground/addons/storybook-addon-code/HTMLMarkup.jsx
```diff
@@ -16,26 +16,84 @@ const TitleBar = styled.div`
...
+      ...options,
+    });
+  };
+
+  // For rendered HTML with SVG icons, apply more aggressive formatting
+  const beautifiedRendered = beautifyHtml(markup, {
+    unformatted: [],
+    inline: [],
   });
+  const unescapedRendered = decode(beautifiedRendered);
```
planctus, line 78: doing this causes the svg opening tags not to be highlighted, maybe you noticed it, it seems that the highlighter is expecting escaped html, so this is not needed for the storybook panel but only for the codepen export. So we can "decode" only when passing the html to codepen instead of here.

planctus (commented):
Cool, it makes a lot of sense to have this, i'd say even more than for what webtools done for our js manipulation of the DOM. 

For the first time i looked at the codepen related code, it's still untouched since v4, the thing is that we had a big problem and we still have it, if we pass the "rendered" html obviously js will not work, so we should refactor this now that we have the "source" html and export to codepen only this but as a consequence the icons will not work, we could ask webtools to allow https://codepen.io/pen, maybe.
We can also add the color modes css, do you want to do this in this pr or do we create a separate task..?

emeryro:
I fixed the svg highlight.
Let's handle the codepen part in a follow-up task

PR #4256, opened 2026-01-14
feat(modal): preventing body scroll when opened - FRONT-5197

planctus:
Apparently with css only we cannot properly apply this lock for the scroll on IOS, so here is a Js solution applied for the moment only in the modal, but we can reuse it wherever we need to lock/unlock the scroll.


src/tools/dom-utils/index.js
```diff
@@ -32,3 +32,28 @@ export const styled = (stylesObject) =>
...
+
+/**
+ * @param {boolean} Lock/Unlock
+ */
+export const handleScroll = (lock, context = document) => {
+  if (lock) {
+    scrollY = window.scrollY;
+
+    context.body.style.position = 'fixed';
+    context.body.style.top = `-${scrollY}px`;
```
emeryro: I'm a bit worried about using a top value here. It should be fine in most cases, but getting window values with javascript is always a little risky. For instance, [the documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollY#value) already reports some differences with Safari
I checked how [Bootstrap](https://getbootstrap.com/docs/5.0/components/modal/#live-demo) is doing it, and they mainly just handle the overflow of the body. Isn't it enough?

planctus: as far as i read, not in IOS, that's the reason why i went for a js solution with position: fixed, i can try a simpler solution and test it, indeed if it's ok for bootstrap should be ok for us as well..

planctus: Just tested and indeed handling the overflow via css is not working in IOS, but all the design systems i checked are using it, despite this. 
Now, i discovered that we already have a utility class doing exactly this, hiding the overflow, we use it somewhere but not everywhere we want to prevent the scroll, the question now is..
What do we do..? 

- One possibility is to keep using this approach, knowing its limitations and maybe applying it consistently where we need it. 
- The other is to explore the possibility of a more complete solution and i think we would end up with the one proposed in this pull request.

Personally i would probably keep using the overflow: hidden, if someone complains because of IOS we'll use boostrap or material or USWDS as a defensive argument :)

emeryro: I would go with the css overflow too. On iOS the modal would usually be full width anyways, so no real possibility to scroll the background. 
But using the utilities may be risky as we ship them as an optional file now. Maybe safer to just put the css rules directly.

planctus: right..done using inline styles for the body.

emeryro (approved):

PR #4257, opened 2026-01-19
feat(tooltip): add tooltip component - FRONT-5204

emeryro:
Add component Tooltip, which can be used on any kind of markup.
Its content comes from the ` title`  attribute


planctus:
Nice, it seems already good enough for a first implementation.  But tricky, i'm sure that we will receive reports for all the situations that we might not want to handle. 

- The positioning, for instance, it's fixed and if you scroll we don't reposition the tooltip, we don't necessarily have to do it, but at least we could hide the tooltip if scrolling or resizing is detected.

- An annoying thing is the use of html, i'm pretty sure that we would receive such a request, bootstrap handles that, but i agree not to do it for the moment.

- It cannot work inside our modal as long as the element is appended to the body, it seems, again from our point of view this might surely be acceptable, but most likely would be reported as a bug.

- About the choices for the implementation, i think that adding global listeners and checking for the presence of the element's attribute would be better than adding multiple dedicated event listeners for then having to rescan using an observer for new elements added.

- I'm still wondering about the opprtunity of using the title attribute, for elements that do not have it is simply a complication, you have to set two attributes, for elements that have it it make sense, but we could also detect that and use the title if present or directly our attribute in case it's not

- This is a design issue, i think the element should have a border, if placed inside a container with a dark background (the footer, for instance) only the text would be visible.

- On the same track, i fear that editors would come with use case where we need to alter the color of the tooltip based on the underlying element's background, in their perspective.

A few points, but it already works nicely and it's really neat

emeryro:
Thanks for the feedback!

Here are the updates:
- add scroll/resize event to close the tooltip
- I discovered that there is a native [popover api](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover), fixing the display in the modal
- use global listeners
- take content of data-ecl-tooltip, if title is not present
- I will check with designer for the border

planctus (approved):

PR #4270, opened 2026-02-05
feat(page-header): new parameters, remove variant - FRONT-5221

emeryro:
- Several new display option added to change the display (background, color_mode, description position, picture_position)
- Remove global variants
- Remove parameter to change title font size


planctus (approved):
It's weird that the color used in the 50/50 variant in some of the color modes is simply white..like "red tomato": 

Image

emeryro:
It is strange indeed, but this is following the figma specs
image

emeryro:
I'll post the preview to the designers, so we will have confirmation

planctus:
Cool, basically i have 1 question and 1 remark:

- shouldn't the background toggle be applied also to the description when placed in the bottom..?
- we miss some styles in print,  spacing for the meta for instance when the image is at the bottom, color background and image "beside"
- Ah, also..we haven't renamed the "thumbnail" param to "featured image", not sure if anything has changed about this request

PR #4273, opened 2026-02-09
fix(datepicker): fix display right to left - FRONT-5220

emeryro:
Fix position of button and dropdown when using the datepicker right to left

### DIFF
--- a/src/components/datepicker/datepicker.js
+++ b/src/components/datepicker/datepicker.js
@@ -73,8 +73,6 @@ export class Datepicker {
         },
       };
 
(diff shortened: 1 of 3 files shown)


planctus:
There is still a small issue in RTL: 
image
but beside this i'm not sure that keeping the width of the dropdown fixed also in mobile is a good idea, i guess that the intention in making it full width was to increase the tappable area, simplify the usage in mobile although i agree that it might look better like this.

emeryro:
The mobile rtl seem to be broken on the duet demo itself. I added a fix on our side

PR #4281, opened 2026-02-23
feat(tooltip): add inverted tooltip - FRONT-5230

emeryro:
Add the possibility to display tooltip as inverted, to be visible on dark background

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -5,6 +5,7 @@ import { queryOne } from '@ecl/dom-utils';
  * @param {Object} options
  * @param {String} options.tooltipSelector Selector for tooltip triggers (requires data-ecl-tooltip attribute and could use title attribute)
  * @param {String} options.tooltipPopupSelector Selector for tooltip popup element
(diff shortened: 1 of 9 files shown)


src/components/tooltip/tooltip.js
```diff
@@ -278,6 +281,15 @@ export class Tooltip {
     // Copy content to tooltip
     this.popup.textContent = content;
 
+    // Use inverted if needed
+    console.log(trigger.getAttribute(this.tooltipInverted));
```
planctus: i should check the eslint configuration, we should not be able to commit debug statements.

planctus:
Since you use "closest", couldn't we use only one attribute, alternatively data-ecl-tooltip or data-ecl-tooltip-inverted instead of using one for identifying the trigger and the other to decide the style..?

To me it's ok to add this "variant", but the problem remains, the user will not be aware everytime of the color of the background and then use the inverted instead of the standard one, imagine a button on an image background, so without a border there is still the possibility of th tooltip to blend with the background, but ok.

emeryro:
Thanks, now the tooltip inverted uses the same logic as the default one. I updated the examples to show the different possibilities.

planctus (approved):

PR #4283, opened 2026-02-24
feat(content-block): don't use list for single meta - FRONT-5236

emeryro:
Use simple div for primary meta when there is only one item


planctus (commented):
Why not doing the same for the secondary meta..? Ok not going through all the components but maybe in this component we could do this consistently

planctus (approved):

PR #4296, opened 2026-03-04
fix(mega-menu): Fixing icon size, icon position, scrolling in mobile, tablet display - FRONT-5243

planctus:
The problem was spotted on a "second panel" in mobile, it means the list of third level items, when there was no featured item set on any of the parent items we were resetting the list scrollability and so we had no scroll.
The example in the demo is item 2.3, i have removed the featured panel from news and media to reproduce a similar scenario to the one where the problem was identified.
I added some comments to explain a bit what we do in the script, it's not really easy to follow what happens. 

A fix has been added for CWS-448 as well, display issues with an Ipad, The fix was to remove the inconsistencies we had using device detection for "desktop" and screen resoulution for "large" displays, now only screen resolution is used, Ipad, behaves as a desktop display when in landscape mode and as a mobile when in portrait mode.


emeryro (changes requested):
This component's maintenance is becoming more and more problematic...
The fix seems fine for the reported issues, but there is a side effect for the first level items: the margin is missing for the icon

Image

planctus:
I didn't see your comment, but i noticed by myself the issue which is a direct consequence of the fix, the reality is that the css rule coming from the button that we are overriding here is too specific, if we fix that we can also remove the !important used in this component.

emeryro:
I'm glad that we are removing the specific device detection here, as it will make things easier to maintain and more consistent.
I just hope that it won't have side effects

emeryro (approved):

emeryro (approved):

PR #4306, opened 2026-03-10
feat(layout-wrapper): Universal grid display of components - FRONT-5242

emeryro:


src/layout/layout-wrapper/layout-wrapper.scss
```diff
@@ -0,0 +1,119 @@
...
+    grid-template-columns: 1fr 3fr;
+  }
+
+  // Vertical flow with 4 cols
+  .ecl-layout-wrapper--vertical.col-4 .ecl-layout-wrapper__content {
+    grid-template-rows: repeat(var(--ecl-layout-rows-4), auto);
+  }
+}
+
+// Set 100% height for all items
```
planctus, line 186: Isn't this the default in a grid display, meaning that the items assume the height of the tallest one and so the full height of the row?
And without this we could offer the option to "override" this default behaviour by setting align-items: start on the grid wrapper.
When i worked on this it seemed nice to have this option, cannot identify a clear situation where you would prefer the items not to be equally tall but there might surely be.

emeryro: right, I added a parameter to disable the full height if needed

PR #4307, opened 2026-03-11
feat(media-container): separate caption and credit - FRONT-5251

emeryro:
Add a new field to put the media credit, in a dedicated HTML tag.
Having the credit in the media caption, as it was the case so far, would still work, but using the new markup is better for accessibility


planctus:
Fine, but why don't we update also the demo of the featured item with this markup...?

planctus (approved):

PR #4309, opened 2026-03-11
feat(tooltip): logic change for accessibility - FRONT-5253

emeryro:
Use multiple tooltip instead of a single instance (one per element needing it). Each of the tooltip is then linked to the trigger in a more accessible way.

Also contains an internal improvement for storybook, to ensure that the resources are correctly loaded before initializing the tooltip

### DIFF
--- a/src/components/tooltip/README.md
+++ b/src/components/tooltip/README.md
@@ -8,36 +8,25 @@ npm install --save @ecl/tooltip
 
 ### Attributes:
 
-- **"title"** (recommended) (string) (default: ''): Tooltip content. Recommended for accessibility as it is exposed to assistive technologies
-- **"data-ecl-tooltip"** (string) (default: ''): Enables the tooltip. Can optionally contain the tooltip content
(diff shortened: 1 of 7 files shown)


planctus:
It seems all good to me, but i'm not sure at this point that it makes sense to retrieve an existing instance more than destroy and recreat it every time or simply update its content, what i mean is that you could use this tooltip to "highlight" an information that you update with or without use interaction, if we reuse the same instance then it will not update the content of the tooltip.

A question i forgot to ask you before, are the designers aware of the fact that we only support plain text currently..? Because it seems very likely to happen that we will have some requests about it coming already from the designers and only then from the users..

emeryro:
I just added an update to improve the loading order in storybook (making sure that the story is correctly loaded before triggering autoInit).
There were some cases where the tooltip was not loaded initially.

Regarding the content of the tooltip, this has not been directly discussed indeed. It may be possible to allow html tags in that string, if we really want to, but currently it offers the same feature as the default "title" tag.
Better make sure that EWPP doesn't put a rich text there...

PR #4313, opened 2026-03-12
feat(file): link meta to download button - FRONT-5252

emeryro:
Add id to meta and allows it to be found by screen readers


planctus (approved):
unbelievable amount of ids.. :)

PR #4314, opened 2026-03-17
feat(banner): add overlay - FRONT-5256

emeryro:
Add an optional overlay, to increase the text visibility when needed


src/components/banner/banner-print.scss
```diff
@@ -242,3 +242,26 @@ $banner-print: null !default;
...
+    background-color: map.get($banner, 'overlay');
+    content: '';
+    inset: 0;
+    mix-blend-mode: multiply;
+    position: absolute;
+    z-index: 0;
+  }
+
+  // Force light text
+  &:not(.ecl-banner--box-bg-light) {
```
planctus: Do we need additional css for this? Can't we use the existing styles for the  color-light variant..? I would associate the two, so that if the overlay is used the ecl-banner--color-light selector is added. Currently we demo the component with the color-dark selector in place but the text is white if the overlay is chosen.
Also, this is what already happen when the user selects the dark box option so there we don't even need this "override" here already.

planctus:
good but it seems we still have the css forcing the light color in the print css, let's remove it also from there.

planctus (approved):

PR #4340, opened 2026-03-25
fix(breadcrumb): hide on mobile for single item - FRONT-5270

emeryro:
Hide breadcrumb on mobile when there is a single item in it


planctus:
There is a misunderstanding, what this pr is doing is basically duplicating something that we already do, if there is one item only this is going to be the "current" item, indeed, and it is already hidden in "mobile".
The issue is about the spacing added by the empty breadcrumb due to styles (padding) set in the page header's css
image

emeryro:
Not really... The breadcrumb is not "empty", it is just that the first item is not displayed on mobile.
What the added css selector do is hide the full breadcrumb (and so the padding/background) on mobile, when there is only one item.
If you prefer, I can change only the padding instead of having a display none, but the css selector seems correct

planctus:
true, i simply misread the selector

planctus (approved):

PR #4341, opened 2026-03-26
feat(add-to-calendar): new component - FRONT-5261

emeryro:
Add new component `Add to calendar`


src/components/add-to-calendar/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "5.0.0-RC1",
+  "description": "ECL Add to calendar",
+  "publishConfig": {
+    "access": "public"
+  },
+  "style": "add-to-calendar.scss",
+  "dependencies": {
+    "@ecl/picture": "5.0.0-RC1"
```
planctus: this can b eremoved, right..?

src/website/public/admin/config.yml
```diff
@@ -951,7 +966,282 @@ collections:
...
+        file: 'src/website/src/pages/ec/components/accordion/docs/accessibility.md'
+        fields:
+          - { label: Body, name: body, widget: markdown }
+      - label: 'Add to calendar'
+        name: 'eu_add_to_calendar_a11y'
+        file: 'src/website/src/pages/ec/components/add-to-calendar/docs/accessibility.md'
+        fields:
+          - { label: Body, name: body, widget: markdown }
+      - label: 'Banner'
+        name: 'eu_banner_a11y'
```
planctus: mmh..i don't get it, why are we doing this..?  If we want to add the a11y pages to EU we can do it, but the idea is to copy all those files from EC so i would wait for them to be available and then we can add also the proper configuration here in the cms

src/components/add-to-calendar/add-to-calendar.story.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  return args;
+};
+
+const getArgTypes = () => {
+  const argTypes = {
+    ...getColorModeControls(),
+    show_title: {
+      name: 'Title',
```
planctus, line 32: we have always used lowercased labels for the controls, did something change about that..? Also here we have a mix since the color_mode label is lowercased

src/website/public/admin/config.yml
emeryro: I saw the accordion EU a11y page there already, so I thought that we didn't add the other pages yet...
I can remove the other pages, but this is odd to have only the accordion

src/components/add-to-calendar/add-to-calendar.story.js
emeryro: True, I just found it better capitalized, as this is a label and not a machine name (and the description is also capitalized). 
If I'm not wrong, there are other control label already capitalized. So let's just decide to go one way or another and harmonize this in a dedicated task

PR #4345, opened 2026-03-27
feat(components): Adding Quiz component - FRONT-5262

planctus:


src/components/quiz/quiz.html.twig
```diff
@@ -0,0 +1,153 @@
+{% apply spaceless %}
+
+{#
+  Parameters:
+    - "variant" (string) (default: reveal) poll otherwise
+    - "title" (string) (default: '')
+    - "description" (string) (default: '')
+    - "with_background" (boolean) (default: false)
+    - "full_width" (boolean) (default: false)
```
emeryro: full width seems to be used only for the background. Do we need to keep two parameters for that?

src/components/quiz/quiz.html.twig
```diff
@@ -0,0 +1,153 @@
...
+    - "slider" (boolean) (default: true)
```
emeryro: I don't really understand this "slider" parameter. What is expected here? 
We could hide the slider based on the number of items, but in any case that would be weird for small screens

src/website/src/pages/ec/components/quiz/docs/code.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+## Reveal
+
+<Playground
+  system="ec"
+  selectedKind="components-quiz"
+  selectedStory="reveal"
+></Playground>
+
+## poll
```
emeryro:
```suggestion
## Poll
```

src/website/src/pages/eu/components/quiz/docs/code.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+## Reveal
+
+<Playground
+  system="eu"
+  selectedKind="components-quiz"
+  selectedStory="reveal"
+></Playground>
+
+## poll
```
emeryro:
```suggestion
## Poll
```

src/themes/ec/maps/color.scss
```diff
@@ -76,6 +76,7 @@ $color-default: (
   'white-alpha-50': color-mix(in srgb, #fff 5%, transparent),
   'white-alpha-75': color-mix(in srgb, #fff 7.5%, transparent),
   'white-alpha-100': color-mix(in srgb, #fff 10%, transparent),
+  'white-alpha-150': color-mix(in srgb, #fff 15%, transparent),
```
emeryro: Is this new color used somewhere? It isn't even defined in Figma

Image

src/website/public/admin/config.yml
```diff
@@ -521,6 +526,11 @@ collections:
         file: 'src/website/src/pages/ec/components/navigation/pagination/docs/accessibility.md'
         fields:
           - { label: Body, name: body, widget: markdown }
+      - label: 'Quiz'
+        name: 'eu_quiz'
+        file: 'src/website/src/pages/eu/components/quiz/docs/usage.md'
```
emeryro: There is something wrong in this file, we should have usage and accessibility for both EC and EU

emeryro (changes requested):
That's nicely done. I like the clean print display.
I let a few remarks, but my main concerns are:
- The quiz cards are not reachable with keyboard, so no way to flip them or answer
- RTL display doesn't seem to be available

src/website/public/admin/config.yml
planctus: Right, but in reality we added a11y pages only in EC, the idea is to copy them in EU afterwards, but we will have to remember to ad them in the cms conf..

src/components/quiz/quiz.html.twig
planctus: Bah, the idea was to make the slider functionality provided by us optional, like giving the chance to use a different library, but it didn't make sense..

src/components/quiz/quiz.html.twig
planctus: mmh..yeah, the idea was to make it work properly with the background even when properly rendered outside of the container, now i joined the two actions, but the problem is that if you now try to use it in the right way it will break.. :)

planctus:
Wow, keyboard navigation kind of tricky, but now we handle it in both variants, you can flip the cards, browse the options, make your selection and see the result.
Navigation using Tab, click with Enter or Spacebar, Esc to move out of the slider form anywhere.

For the RTL you have to switch story, basically, if you enable it when you are in one of the two variants, go in the other and then back, it should work.

emeryro (approved):

PR #4346, opened 2026-03-28
Fix lang selector

rriemann:
fixing the lang selector when css framework apply by default overflow-wrap to all `li` elements.

image


rriemann:
There is an issue with the branches. I try again.

Issue #4333, opened 2026-03-21
5.0.0-RC1: CSS background for drop-down menu not long enough


rriemann:
Hi,

I integrate the ECL 5.0.0-RC1 site header in the default template of https://vitepress.dev . I noticed an issue with the header. I also had this with the latest version (4.11.0) that I tested before.

<img width="2361" height="1044" alt="Image" src="https://github.com/user-attachments/assets/19eac614-95c7-422d-b66c-815ad1777f07" />

I would expect that the box of (1) is longer than the box of (2).

planctus:
For sure what you expect is what the component should do, not sure in your case what is going wrong, the element is ecl-mega-menu__wrapper and you should see that its size is set via js as the result of the calculations relative to the different panels, it might be that this is failing for that item specifically, or do you have the same issue also in the other menu items?

rriemann:
You can check the website yourself at https://simpl.pages.code.europa.eu/pso/dg-test/ssg-website/

Code is at: https://code.europa.eu/simpl/pso/dg-test/ssg-website

I noticed that the wrapper height changes from one main menu item to another, but it is always to short. So the code to set the height runs, but the calculus is wrong.


<img width="2507" height="1744" alt="Image" src="https://github.com/user-attachments/assets/52223202-0150-4e3d-9ee8-ffc590b44632" />

planctus:
The reason for what you are seeing is in some css added by your application, not a wrong calculation:
```
    *,:before,:after {
        box-sizing: border-box
    }
```
On our side .ecl-mega-menu__wrapper is content-box, if you see indeed we set the same height value for the wrapper and the panels inside, that is why if you make the wrapper border-box it will be shorter, due to its padding.
We might add a box-sizing rule to that element to prevent unintentional overrides.

PR #4347, opened 2026-03-28
fix language selector items to avoid line breaks in lang codes

rriemann:
image

### DIFF
--- a/src/components/site-header/_site-header-dropdown.scss
+++ b/src/components/site-header/_site-header-dropdown.scss
@@ -169,6 +169,10 @@ $language-list: null !default;
   padding: 0;
 }
 
+.ecl-site-header__language-item {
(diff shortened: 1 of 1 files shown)


rriemann:
I realise that the language name for "NB" is also split and this comes from `box-sizing: border-box;` that was already a problem in https://github.com/ec-europa/europa-component-library/issues/4333

planctus:
sorry, @rriemann but you want to use our css to override the styles set by the applications you are using..? Do it on your side, or reset the default styles it gives you, like removing that css, we cannot set styles in ECL to fix the issues that it causes.

rriemann:
I understand your decision. My pull request was motivated through a goal that I have read somewhere (unable to find the link right now) that you want to namespace everything and be framework neutral to allow components to work in whatever environment.

PR #4349, opened 2026-03-30
feat(rtl): cleanup css for rtl display - FRONT-5271

emeryro:
use inline/block properties where possible instead of having specific rules for rtl


planctus (commented):
Not sure what criteria did you use to identify what to replace with logical properties, it could be that you looked for the :dir(rtl) identifier, but looking around i see still an insane amount of left and right, and i agree maybe not to replace them at once, although mixing physical and logical properties can easily lead to problems. 
Anyway there are a few cases where we probably never looked for the RTL display, because it is broken and using logic properties would fix them.
I noticed this in these components:
- gallery full width is broken in RTL
- navigation list with illustration is broken in RTL
- list with illustration zebra in mobile is broken in RTL

I would take the chance to fix those using logical properties

emeryro:
There were indeed still work to be done there... I updated most of the right/left rules.
The remaining ones are for specific display, including javascript calculus

planctus (approved):

PR #4353, opened 2026-04-01
fix(site-header): fixing white space on the right in mobile, fixing mega menu in EU - FRONT-5269

planctus:


src/components/site-header/site-header-ec.scss
```diff
@@ -268,19 +268,15 @@ $menu-top: calc(44px + 2 * var(--s-xs));
...
-  right: calc(-1 * var(--s-l));
+  // 44px as the other buttons
+  width: 2.75rem;
 
-  &:dir(rtl) {
-    left: calc(-1 * var(--s-l));
-    right: auto;
-  }
+  // container's padding inline end
+  right: var(--s-m);
```
emeryro, line 277: we should use `inset-inline-end` instead of right, to handle rtl display
This is what is done in https://github.com/ec-europa/europa-component-library/pull/4349

emeryro (changes requested):

src/components/site-header/site-header-ec.scss
planctus: i suggest to do this all in the same pull request, if this is ok for what was intended, fixing the issue of the oversized layout in mobile, i would first merge this and then update the properties in your pr, there are many others occurrences of right or left in this files, it wouldn't make sense to update only the ones i touched here.

emeryro (approved):

PR #4355, opened 2026-04-07
feat(text-media): Add new component: Text and media - FRONT-5277

emeryro:


planctus:
While testing i had often issues with the caption rendering, this is chrome on windows:
image
here you have the two issues together, one is js related, at first load i've seen many times the two lines with the caption and the copyright overlapping, in different browsers, in storybook or outside of storybook, that kind of issues that are entirely time related, a small resize would immediately fix it.
The other issue is that i don't see the transparent background here, not sure if it's really an issue, the snapshot was taken in browserstack, i haven't seen this neither locally nor in any other browser there.

emeryro:
I noticed the overlap from time to time too (especially on browserstack). The position is calculated with js, as we cannot add a wrapper around it (not allowed in a `<figure>`). I thought that we solved some of the js timing issues in storybook, but that's not the case here obviously...
It shouldn't happen outside of storybook though, so we can take time to check it again later

For the background, I cannot reproduce it in browserstack so far. The background color is set the same way as in other components, so I don't really see why it would be problematic here

planctus (approved):

PR #4357, opened 2026-04-07
fix(tabs): fix possible overlap - FRONT-5279

emeryro:
Fix several issues on the tabs:

- We were using some workaround to approximate the moment where the items become hidden below the dropdown (using a fixed "25px" value as margin).
It works fine as long as we don't have too many tabs; the more tabs, the more offset we get.
Now the real margin is taken into consideration in the calculus

- Use query selector to get the tab item, instead of childNode (less reliable)

- Fix items being partially hidden if the dropdown if there are many of them (padding was pushing the list out of the screen)

It can be tested by putting lots of tabs, for instance:
```
items: [
    { label: 'A', path: exampleLink, id: 'tabs-1' },
    { label: 'B', path: exampleLink, id: 'tabs-2' },
    { label: 'C', path: exampleLink, id: 'tabs-3' },
    {
      label: 'D',
      path: exampleLink,
      is_current: true,
      id: 'tabs-4',
    },
    { label: 'E', path: exampleLink, id: 'tabs-5' },
    { label: 'F', path: exampleLink, id: 'tabs-6' },
    { label: 'G', path: exampleLink, id: 'tabs-7' },
    { label: 'H', path: exampleLink, id: 'tabs-8' },
    { label: 'I', path: exampleLink, id: 'tabs-9' },
    { label: 'J', path: exampleLink, id: 'tabs-10' },
    { label: 'K', path: exampleLink, id: 'tabs-11' },
    { label: 'L', path: exampleLink, id: 'tabs-12' },
    { label: 'M', path: exampleLink, id: 'tabs-13' },
    { label: 'N', path: exampleLink, id: 'tabs-14' },
    { label: 'O', path: exampleLink, id: 'tabs-15' },
    { label: 'P', path: exampleLink, id: 'tabs-16' },
    { label: 'Q', path: exampleLink, id: 'tabs-17' },
    { label: 'R', path: exampleLink, id: 'tabs-18' },
    { label: 'S', path: exampleLink, id: 'tabs-19' },
    { label: 'T', path: exampleLink, id: 'tabs-20' },
    { label: 'U', path: exampleLink, id: 'tabs-21' },
    { label: 'V', path: exampleLink, id: 'tabs-22' },
    { label: 'W', path: exampleLink, id: 'tabs-23' },
    { label: 'X', path: exampleLink, id: 'tabs-24' },
    { label: 'Y', path: exampleLink, id: 'tabs-25' },
    { label: 'Z', path: exampleLink, id: 'tabs-26' },
    { label: 'Other', path: exampleLink, id: 'tabs-27' },
  ],
```


src/components/tabs/tabs.js
```diff
@@ -689,26 +690,68 @@ export class Tabs {
...
+              : item.getBoundingClientRect().right;
+          const fits =
+            direction === 'rtl' ? itemEdge >= threshold : itemEdge <= threshold;
+
+          if (fits && !hiddenItems.includes(i - 1)) {
+            // item fits
+          } else {
+            item.classList.add('ecl-tabs__item--hidden');
+            if (
+              item.childNodes[0].classList.contains('ecl-tabs__link--active')
```
planctus: reading the issue description it seems that this is the cause of the second problem they had, indeed with childNodes we get whatever it's the first child of the item, including an empty space. We should replace this with a query for the active class, i think.

src/components/tabs/tabs.js
```diff
@@ -666,12 +666,13 @@ export class Tabs {
...
         }
-        let listWidth = 0;
+        // Reset to auto so scrollWidth reflects the natural content width
+        this.list.style.width = 'auto';
         this.listItems.forEach((item) => {
           item.classList.remove('ecl-tabs__item--hidden');
-          listWidth += Math.ceil(item.getBoundingClientRect().width);
         });
-        this.list.style.width = `${listWidth}px`;
+        // scrollWidth includes item margins, unlike getBoundingClientRect().width
```
planctus: this is confusing, it refers to the previous implementation where the width was calculated using the single items width but not considering the margin, i think we can remove it or we can mention the fact that we are using scrollWidth to consider also the overflow

planctus (commented):
But..wasn't the previous approach better from an RTL support point of view..? I mean that we did not have to take into consideration the direction at all, because we were calculating the width basing on the single items + the button and that doesn't change inverting the direction.
To me we had a bug, because we were not considering the item's margin, it might be that in the past we did not have them, we could have simply fixed that, maybe.
Also, we had a problem, and we still have it, the width of the button is not reliable, i'm not sure the reason why, but if we calculate that in document.fonts.ready we get the right one, it might be because of the custom font we use.

emeryro:
I reverted the logic to what we had, so no more exception for rtl, still keeping the fixes added before

planctus (approved):

PR #4360, opened 2026-04-08
feat(featured-item): add micro title - FRONT-5281

emeryro:
Add an optional micro title on top of the title


src/components/inpage-navigation/inpage-navigation.story.js
```diff
@@ -12,7 +11,8 @@ const system = getSystem();
 if (system === 'eu') {
   demoData.icon_size = 's';
 }
-const lorem = loremIpsum({ count: 25 });
+const lorem =
```
planctus, line 14: is  this for chromatic..?  in theory we should be ignoring all the content in the main area

src/components/featured-item/featured-item.story.js
```diff
@@ -38,14 +39,26 @@ const getArgTypes = (data) => {
 
   argTypes.show_media = {
     type: 'boolean',
-    name: 'show media',
+    name: 'Show media',
```
planctus, line 42: i noticed this also in the add to calendar pr, if we want to change approach about using always lowercased labels for the controls let's do it consistently everywhere, but i don't necessarily see the reason why doing it..

src/components/inpage-navigation/inpage-navigation.story.js
emeryro: It is indeed. It was reported as a change every time

PR #4365, opened 2026-04-13
feat(components): Adding animated numbers - FRONT-5282

planctus:


src/components/animated-numbers/animated-numbers.js
```diff
@@ -0,0 +1,295 @@
...
+      );
+
+      this.itemsElements.forEach((element) => {
+        this.intersectionObserver.observe(element);
+      });
+    }
+
+    if (this.attachHoverListener && this.animateOnHover) {
+      this.itemsElements.forEach((element) => {
+        element.addEventListener('mouseenter', this.handleHoverStart);
```
emeryro: that was a request to trigger the animation again on hover?
That's unusual...

emeryro (commented):
It was reported already, but we don't have to push the boundaries very far to break the design... (this is "billion" in greek)

Image

src/components/animated-numbers/animated-numbers.js
planctus: i thought so, we can ask for a confirmation today in the meeting otherwise i will remove it.

planctus:
yeah, i can remove the handling of the white space and let it break in multiple lines to mitigate the issue, but it's three strings in a small box, it's clear that can only work they way it was designed if a severe limitation on the lenght of the strings is in place.

emeryro (approved):

PR #4370, opened 2026-04-16
fix(tooltip): Adding api pages - FRONT-5280 

planctus:


src/website/src/pages/ec/getting-started/index.mdx
```diff
@@ -263,9 +263,16 @@ https://webtools.europa.eu/showcase/demo/?comp=icons&section=about&demo=how_to_u
 
 ---
 
+### Tooltip
+
+ECL's tooltips can be enabled by adding a `data-ecl-auto-init="Tooltip"` attribute to a container in the page, like the body.
+All the elements containing a `data-ecl-tooltip` or a `data-ecl-tooltip-inverted` inverted will have their tooltip associated, it will show the content of the `title` attribute.
```
emeryro: that's true, but not the only way to use it in the end. The content of the tootltip can be provided directly to `data-ecl-tooltip` or `data-ecl-tooltip-inverted`.  Suprisingly, `title` is quite bad in terms of accessibility, so I considered it as a fallback
It is clarified a little in the [readme](https://github.com/ec-europa/europa-component-library/blob/v5-dev/src/components/tooltip/README.md)

emeryro (commented):
Just a remark for the doc.
I also noticed that the readme file is not properly rendered in the "notes" tab in storybook, we may want to fix that too

Image

emeryro (approved):

PR #4376, opened 2026-04-21
feat(menu, mega-menu): close on click - FRONT-5265

emeryro:
Close the menu and mega menu when clicking on any final link


planctus (commented):
For the mega menu, wouldn't it be easier to use the existing listeners for click instead of adding many new ones..? All these clicks are already handled by handleClickOnItem and  handleClickOnSubitem, i think.
It would be nice to add an anchor link so that this can also be tested,

emeryro:
Indeed, I updated the js to use existing listeners.
The "info" and "see all" links are also treated as any other sublink, as they share the same listeners

planctus (commented):
Mmmh..but why not simply add an handler for links, a-tags, and only execute the closeOpenDropdown() method..?  I mean, that's the idea, any click on a link should close the dropdown, so no logic whatsoever, why doing it instead as an "exception" for the existing methods to handle the click on the items that are meant to handle the Ui of the menu ..? I know i suggested to use them, previously, but probably because that seemed a simpler way in relation to the previous implementation, now i fail to understand why we haven't simply added one single generic click handler for doing this (am i missing something..? :) )

emeryro:
You're right, I thougt that the inner elements, opening sub sections, were also links. But they are buttons in fact, so we can target all links directly.
I just kept some code cleanup that were done here, and added an event on all links

src/components/mega-menu/mega-menu.js
```diff
@@ -272,6 +259,15 @@ export class MegaMenu {
       });
     }
 
+    // All <a> inside panels close the dropdown on click (covers leaf sublinks,
+    // see-all, info-links, and featured links)
+    this.endLinks = queryAll('.ecl-mega-menu__mega a', this.element);
```
planctus: Two questions:
- Why limiting this query to the links in the dropdown if then we are applying the same behavior also to the main items..?
- But instead of doing this, wouldn't it be better to simply add a click listener to the menu itself and just check if the target is a link? One event listener instead of potentially many, also when cleaning up when we remove them

emeryro:
Should be cleaner now

planctus (approved):

PR #4383, opened 2026-04-24
feat(quiz): Showing a box to explain how to skip the quiz when using the keyboard - FRONT-5301 

planctus:


emeryro (commented):
it works in the current demo, but what if there are only 3 questions (and so no pagination/next)?
maybe we could add an hidden item at the end of the quiz, and target it when needed

planctus:
right, let's simply make the next button visible in that case when we move the focus on Esc

emeryro (approved):

PR #4384, opened 2026-04-24
feat(quiz): Improving focus behavior and styles - FRONT-5311

planctus:


emeryro (commented):
Nice.
It works fine when tabing forwards, but I noticed some glitch when doing it backwards (jump of some elements)
Still that could be improved later if we want to merge this already

emeryro:
Also I noticed that the focus display is not the one presented in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/CWS-554 
Maybe the design specs came after...

emeryro (approved):

PR #4385, opened 2026-04-25
feat(quiz): Using fieldset and radio inputs in poll variant - FRONT-5300

planctus:


emeryro (commented):
I updated this PR to merge it with other markup changes (using section, article and lists).
I let you check

PR #4387, opened 2026-04-27
feat(quiz): remove aria-label, add aria-labelledby - FRONT-5309

emeryro:
update aria attribute for each quiz card


emeryro:
merging, as it is blocking the other tasks

PR #4388, opened 2026-04-28
feat(quiz): improve markup - FRONT-5302

emeryro:
use semantic HTML markup where possible


emeryro:
merging as it is needed for other tasks

PR #4395, opened 2026-05-04
feat(quiz): Making poll result accessible - FRONT-5296

planctus:


src/components/quiz/quiz-card.html.twig
```diff
@@ -167,8 +173,24 @@
         <li
           class="ecl-quiz-card__option"
           data-match="{{ _option.correct ? "true" : "false" }}"
+          aria-label="{{ _option.correct ? _correct_label : _incorrect_label }}"
```
emeryro: I think that this aria-label will prevent the rest of the list item to be read. 
Looking at the request that's not really clear indeed what should be done with aria-label. So let's see Regine's feedback here

src/components/quiz/quiz.js
```diff
@@ -479,118 +479,250 @@ export class Quiz {
     }
   }
 
+  /**
```
emeryro: I guess that most of this javascript is just some cleanup and improvement, as it doesn't seem related to the icon display

emeryro (commented):
Visually it looks fine.

src/components/quiz/quiz.js
planctus: i improved the keyboard navigation, it was not really handling all the use cases before, then it was becoming unreadable with a single method and i took the chance to reorganize it with small helpers and a simple method using them.

PR #4402, opened 2026-05-05
feat(quiz): update focus color for poll - FRONT-5297

emeryro:
Apply latest design


emeryro:
Done in https://github.com/ec-europa/europa-component-library/pull/4395

PR #4404, opened 2026-05-06
feat(fact-figures): add sources - FRONT-5308

emeryro:
add sources link to fact and figures, similar to what is done in animated numbers

This also introduce an easier way to use link variant in other components, when we cannot rely on the markup only (if it is different between EC and EU for instance)

Fix applied to animated numbers, regarding the source wrapping (same logic as here)


planctus (commented):
I see that when "centered" also the sources move to the center, not sure that's the intended behavior

emeryro:
yes, that's how it is designed
image

planctus (approved):

PR #4418, opened 2026-05-12
feat(animated-numbers): Using list markup and article - FRONT-5318

planctus:
fullfills tickets
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-5318
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-5322


emeryro (approved):
Fine for me. Let's send it to Regine too, just in case
We could also do the improvements from https://citnet.tech.ec.europa.eu/CITnet/jira/browse/CWS-585 before, as it is related to the markup too

PR #4420, opened 2026-05-13
feat(collapsible-forms): Accordion variant to handle content in the sidebar - FRONT-5294 

planctus:


planctus:
For the main use case is surely the right thing to hide the element when printing, but i wonder if we might ever have someone complaining because his collapsible sidebar is not visible when printing...i'm for keeping hidden unless that happens, by the way.. :)

src/components/accordion/accordion.js
```diff
@@ -0,0 +1,91 @@
...
+    accordion.init();
+    root.ECLAccordion = accordion;
+    return accordion;
+  }
+
+  constructor(
+    element,
+    {
+      sidebarItemSelector = '[data-desktop-open]',
+      sidebarBreakpoint = '(min-width: 996px)',
```
planctus, line 27: it would be nice to dynamically read this from somewhere, the problem is that no one will ever change this value setting an option in the js. What if we add a parameter with this default value and we read the media query from there..?

src/components/accordion/accordion.js
```diff
@@ -0,0 +1,91 @@
...
+   * Initialise component.
+   */
+  init() {
+    if (!ECL) {
+      throw new TypeError('Called init but ECL is not present');
+    }
+    ECL.components = ECL.components || new Map();
+
+    // Sidebar variant: keep items open on desktop, collapsed on mobile
+    const sidebarItems = this.element.querySelectorAll(
```
planctus: minor thing, but you reuse this so why not making it an element of the class so you don't have to repeat the query..?

planctus (commented):
super minor thing, but if we can do something to center a bit this i would...

Image

src/components/accordion/accordion.js
emeryro: would be nice indeed. There are other component already in that situation (banner for instance).
Not sure that putting it in the markup would make it more reliable
I'll create a follow-up task for that, so we take the time to evaluate the options

PR #4422, opened 2026-05-18
feat(animated-numbers): use link css directly - FRONT-5332

emeryro:
Use link css for source links, instead of redefining the colors in the component


planctus (commented):
I'm confused..this changes the current styles.
It seems to me we are using the "default" link styles here, not the standalone ones and the inverted link is currently having a different color in hover, compared to the one used in this component. 
Maybe this could/should be updated using the one that we currently have in the sources links but in this component i chose not to change the overall styles for the link also because the focus styles would involve a bigger change, probably, due to the "double border" used in the "inverted".

emeryro:
The styles used previously were coming from the designer? I looked at figma, and there it is using the default link, with no mention of specific color on hover
https://www.figma.com/design/WYHzeY3pB0U1avgX6qv9gI/ECL-2026%E2%80%A8New-components?node-id=2404-9912&p=f&t=sHYAYGdZ8MqLzJAF-0

planctus:
But indeed, currently we have the default (no standalone) styles, here you are applying the "standalone" ones.

For the inverted yes, not sure what you have in the component file,in figma but the designer wanted me to update the global styles of that link variant, using that grey on hover, i only applied this in the animated numbers, we can update it globally and then use it here, what i'm saying is that the current styles were just approved by the designer, here we are altering them.

emeryro:
ok, let's create a task to update the inverted links then. So we would be able to use them directly here.
For the underline, I'll ask her to update figma if it is not correct

planctus (commented):
Question, isn't it more logical to conditionally render the link as inverted if with_background is true instead of re-applying the related styles to a "generic" link..?

emeryro:
I think I tested that before, but it didn't work well. I'll double check

emeryro:
Done. It was not working before as I was defining styles for the default link too (so it did override the inverted link).
But now it works fine

src/components/animated-numbers/animated-numbers.scss
```diff
@@ -5,6 +5,7 @@
 
 @use 'sass:map';
 @use '@ecl/utility-screen-reader/screen-reader';
+@use '@ecl/link/link' as link-component;
```
planctus: This is a leftover i guess..

planctus (approved):

PR #4424, opened 2026-05-19
feat(quiz): improve keyboard navigation - FRONT-5316

emeryro:
Revamp the keyboard navigation, to allow jump from card to card with tab, browse options with arrow, and avoid flipped cards to be focused

### DIFF
--- a/src/components/quiz/quiz-card.html.twig
+++ b/src/components/quiz/quiz-card.html.twig
@@ -133,6 +133,7 @@
     <div
       class="ecl-quiz-card__back"
       hidden
+      inert
     >
       <div class="ecl-quiz-card__header">
         <div class="ecl-quiz-card__icon">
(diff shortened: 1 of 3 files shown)


src/components/quiz/quiz-card.html.twig
```diff
@@ -132,6 +132,7 @@
     </div>
     <div
       class="ecl-quiz-card__back"
+      aria-hidden="true"
```
planctus: But why do we do this..? It's already hidden, as far as i understand this means that it's already hidden to assistive tech and adding aria-hidden is only a redundant overhead.

emeryro: That was the request, to add aria-hidden. I was hoping that it would somehow fix the issue, but Regine confirmed that it does not.
I'll check for other solutions

planctus (approved):

PR #4431, opened 2026-05-26
feat(animated-numbers): control for icon title - FRONT-5319

emeryro:
Add a control in storybook to see the icon title


planctus (commented):
I would suggest to add a min height equal to the icon size for the item's header, now that we added a control to change the icon, setting it to none misaligns the content in relation to the other items, we don't have an indication by the designers but i guess this is the expectation.

planctus (approved):

PR #4434, opened 2026-05-26
feat(highlight-box): new component - FRONT-5340

emeryro:
New component: highlight box


src/themes/ec/maps/color.scss
```diff
@@ -356,6 +356,9 @@ $color-modes: (
         map.get($color-palette, 'green-lemon-50') 50%,
         transparent
       ),
+    '--cm-surface-color-mode-lowest': #{map.get($color-palette, 'blue-50')},
```
planctus, line 359: Funny, in a way, i have just realized that the slogan ticker was supposed to use a new token introduced only one year after the definition of the color modes, i assumed it was just a mismatch in the name of the token and the designer approved it like that.  So, i don't know, maybe we merge this first and then i'll update the token used in the other component, but how is it possible that we are still introducing new tokens for every new component...

planctus (approved):

PR #4435, opened 2026-05-27
feat(footer): add subscribe button - FRONT-5342

emeryro:
Add optional action button to the footer


planctus (commented):
This looked a bit weird so i checked on figma and the styles in EU are different, is it because of the pending updates on the button component..? 

Image

PR #4437, opened 2026-05-27
feat(components): Adding slogan ticker - FRONT-5341

planctus:


emeryro (commented):
There are a few things that could be improve, but not a blocker here. I let you check if it could be done easily, or if we need a follow-up task, to not block the release
- rtl display is not working well
- if we manually pause the component, it always restart when we scroll back to it (or resize the screen); we should keep track of the pause status
- print display seem wrong
- in some occasion, there seem to be some performance issue. I don't know if it is because of storybook, or my browser. For instance, if I let the component run a little, then try to switch to EU, or resize the window, the css may not load anymore, the browser tab would freeze, and my laptop's fan goes crazy. This doesn't happen every time, but I only notice that when displaying this component

planctus:
- i fixed RTL, but in storybook this is only visible in the page example and by moving it out and in the viewport after having enabled RTL. In a real RTL page it would work from the first load.
- fixed print
- not restarting the ticker if it has been stopped
- i see that there is something wrong with what happens in storybook, i see the component being re-initialized for instance without being destroyed, not sure where this comes from, worth an investigation but it shouldn't be harmful in the real world, your issue is scarier, will let it run for minutes and see what happens, but in that case i fear it might be an issue coming directly from embla + autoscroll

emeryro (approved):

PR #4439, opened 2026-05-27
feat(quiz-animated-numbers): Allowing background without full width - FRONT-5347

planctus:


src/components/animated-numbers/animated-numbers.html.twig
```diff
@@ -79,7 +85,7 @@
   data-ecl-auto-init="AnimatedNumbers"
 >
 {% if (_items is not empty and _items is iterable) or (_sources is not empty and _sources is iterable) %}
-  {% if _with_background %}
+  {% if _full_width %}
```
emeryro, line 88: we should update the test for the closing tag too

src/components/animated-numbers/animated-numbers.html.twig
```diff
@@ -49,11 +51,15 @@
   {% set _css_class = _css_class ~ ' ecl-animated-numbers--border' %}
 {% endif %}
 
-{% if _with_background %}
+{% if _with_background or _full_width %}
```
emeryro, line 54: this is quite strange... now we would have the possibility to have with background (full width or not), and full width, but only with background
What if they ask tomorrow to have the full width without background?

src/components/quiz/quiz.html.twig
```diff
@@ -79,7 +85,7 @@
   data-ecl-auto-init="Quiz"
   {{ _extra_attributes|raw }}
 >
-{% if _with_background %}
+{% if _full_width %}
```
emeryro, line 88: same thing for the closing tag

emeryro (changes requested):

src/components/animated-numbers/animated-numbers.html.twig
planctus: Mmhh..in the animated numbers full width makes only sense with the background, that is basically the only thing that would change enabling it, so this seems to me only a way to simplifiy the usage of this variant, otherwise you would have to use two parameters to do it.

emeryro (approved):
Good for me.
We should just remove the extra story

PR #4441, opened 2026-06-02
feat(featured-item, text-media): update link color - FRONT-5336

emeryro:
Make sure that the link used in the description of the featured item and text media follow the content color


planctus (commented):
What a struggle, i don't know, it feels there has to be something wrong in doing this, for how they are using the color modes the risk of having to introduce this in many different components, if not them all, is very high.
Couldn't we add a selector to the elements where we need to override the link color using the same color of the parent element and use inherit as the value of the color property for all the different  states so that we don't have to retrieve any value from any map and we have a single "utility" css..?

emeryro:
Thanks for the proposal.
I added a mixin, using currentcolor. It should be reusable in other cases.
The solution is not perfect though, as we don't handle specific hover or focus colors, but that would require to create new dedicated color token for each case

planctus (approved):

PR #4447, opened 2026-06-04
feat(button, link): improve css structure - FRONT-5344

emeryro:
Main ideas:
- Reduce the amount of css duplication between button and links; now the link-button would use the code from the button component directly (with some extra rules)
- Use css variables for link and buttons, for the properties that are often modified (like background or text color)
- Use small scss mixins to have a more modular approach; mixins are used in the button and link directly, and could be used in other components too (featured item for instance)


emeryro:
I tried to cleanup a little the code, and add comments in link and button.
Hopefully that would be easier to review (better check the final files rather than the diff)

PR #4451, opened 2026-06-08
feat(animated-number, fact-figures): add individual sources - FRONT-5352

emeryro:
Add sources for individual items in fact and figures, and animated numbers


planctus (approved):
If we get a third component using the "sources" i would propose to create a template for it, there is a lot of duplicated code otherwise. But even like this, couldn't we define a macro and reuse it instead of duplicating the code for the single and the global sources..?

planctus (approved):

PR #4463, opened 2026-06-15
docs(website): update display and navigation - FRONT-5364

emeryro:
Color and spacing updates for the website.
Also prevent navigation loss when switching between EC and EU


planctus:
Checking around i found some inconsistencies in the getting started page, maybe we can fix them here:
tooltip:
image

we are using the Message component as an example, but we don't have it anymore:
image

we are mostly referring to the esm export of our js, here we mention the "old" file instead:
image
If we change this we need to add type="module"

datepicker:
image

PR #4580, opened 2026-06-18
feat(gallery): update icons - FRONT-5170

emeryro:
Use phosphor icons


planctus (commented):
It's a bit confusing, why do we have now this mix of icons from phosphor and other from the old set, wouldn't it be better to use all the icons from the same set?

emeryro:
Even in figma they still use different icon families. I did the change here for the zoom icon (and play), as this is something that was discussed a few month ago with Daniela.
It is probably worth having a more global dedicated task to review the icons (design first), and update them were needed, knowing that not all the icon are available in phosphor (as webtools has it), and surely not with the same name.
This task could be a trigger to start the discussions

planctus (approved):

PR #4581, opened 2026-06-19
feat(accordion): Making media query configurable from the template - FRONT-5349

planctus:
I've only touched the accordion, because it seems the only case where the media query is used to determine a behavior that might be wrong if the responsiveness of the layout in which the collapsible sidebar will be used is using a different breakpoint for switching to a multi columns layout.

In order not to introduce a breaking change (although purely theoretical) i kept the js option priority and type of value, so a media query, i would have probably preferred to further simplify the usage by using just the breakpoint shortened code, like l, m, xl...


src/components/accordion/accordion.js
```diff
@@ -1,6 +1,6 @@
 /**
  * @param {HTMLElement} element DOM element for component instantiation and scope
- * @param {Object} options
+ * @param {Object} options.sidebarMediaQuerySelector Selector for the media query to match
```
emeryro:
```suggestion
 * @param {Object} options
 * @param {String} options.sidebarMediaQuerySelector Selector for the media query to match
```

emeryro (commented):
With that we increase the usage of magic numbers in our files.
If the value is already defined by default in the js, why do we need to also define it in the template? We could leave the parameter empty by default, and take it from the js in that case.

emeryro (approved):

emeryro:
That's already an improvement.
We may go a step further (here or in another PR), and centralize the breakpoint value used in the js.

Currently there are several component using hardcoded values in their js (accordion, carousel, menu, mega-menu, gallery, maybe others). We may put the values in a shared script (or in dom-utils). That would reduce the maintenance.

Alternatively, the breakpoint could be defined in css variables, and read from the js.

planctus:
This is not like other use cases, i agree that it would be better to have a mapping of our breakpoints in js and use "s" "l" "m" in the classes instead of hardcoding values but that's something else, here we are giving the users the chance to alter a default value (which could come from the mentioned mapping) that might not match the layout defined by their markup, so something that needs to be configurable  (as it is, indeed, in the first test or implementation done by ewpp) before they could only do it via js, now they can do it also via the template. that is the scope.

emeryro:
Ok I renamed the task to match what has been done, and created a clone to handle the more global improvement

PR #4582, opened 2026-06-19
fix(default-css): Not overriding color of text in typography - FRONT-5374

planctus:
More than fighting against the "default css" it seems to me logical to set the color of the generic paragraphs to be as the rest of the text elements in a container, why would we say indeed that the text inside a p tag has always to be of a certain color, it's not correct, the color should be taken from the context and be consistent with the rest of the text. 
I haven't added a paragraph in the demo not to alter its look and feel due to the margins.


emeryro:
Hmm, that would work only if there is a color defined at higher level (so when placed in a component for instance).
If we just put a paragraph at page level, the color would be black (and not "brand")
We could set the color by default in the body or html to fix that I guess

planctus:
I did not touch the link, in order to apply the right color in the examples we have in the default css demo i used the ecl-content-color class.

src/utilities/typography/typography.scss
```diff
@@ -96,7 +96,7 @@ $typography: null !default;
 }
 
 %ecl-paragraph {
-  color: map.get($typography, 'text-color');
+  color: currentcolor;
```
emeryro: the map value is correct, we should keep using it

emeryro (approved):

PR #4592, opened 2026-06-29
fix(list): fix bullet position - FRONT-5384

emeryro:
Fix bullet position when using a long standalone link

### DIFF
--- a/src/components/ordered-list/ordered-list-print.scss
+++ b/src/components/ordered-list/ordered-list-print.scss
@@ -15,6 +15,7 @@ $list-print: null !default;
   color: map.get($list, 'text-color');
   font: map.get($theme, 'font-print', 'm');
   list-style-position: outside;
+  list-style-type: map.get($list, 'ordered', 'marker');
   margin: 0;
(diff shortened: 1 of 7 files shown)


planctus (commented):
Isn't this a bit risky..? Don't see any obvious way this might create issues, but there are a lot of use cases involving a link, couldn't we set this style in the list to limit the impact and "surgically" tackle the issue that was identified..?

planctus (approved):

src/components/ordered-list/ordered-list-print.scss
```diff
@@ -29,8 +29,9 @@ $list-print: null !default;
   margin-top: map.get($theme, 'spacing-print', 's');
   max-width: var(--max-w);
 
-  .ecl-link {
```
planctus, line 32: it seems that we introduce a regression removing the padding, it was aligning the text with the marker: 

Image

emeryro:
Using the padding make things worse in fact.
For some reason, we have to use vertical-align: text-top on print (and not on screen).
It is still not perfect on EU when printing standalone links in lists (line height is larger), but I don't really see what could be done here.

PR #4594, opened 2026-06-30
chore(colors): Updating two colors - FRONT-5382

planctus:


emeryro (approved):

src/utilities/background/index.story.js
```diff
@@ -48,12 +48,17 @@ const getArgTypes = () => {
       options: [
         'surface-lowest',
         'surface-lowest-variant',
+        'surface-color-mode-lowest',
```
emeryro: only a few color mode tokens are converted to utilities (otherwise the css file size would explode). They are defined in background, border and typography css utilities files.
The list can be reviewed, but  these new options would not work currently

emeryro (approved):

PR #4596, opened 2026-07-01
feat(file): implement new style - FRONT-5193

emeryro:
Updated design for the file download component.
This comes with a few parameter update in the twig template (with backward compatibility), and markup update to accommodate the new element placement
Twig files have also been consolidated in a single file


planctus:
There is something fundamentally wrong in the behavior of this component now, beasically we force the user to re-click on the toggle (which is not anymore clearly identifiable, by the way. There is a language name, ok that we use the icon normally associated to an expandable, but it's not really clear) to close the translations overlay, it really doesn't make sense, also before it was not taking space unless the user wanted to see the translations, so now the only thing that changed is that we introduced a potential issue with z-index layer and a usability issue due to the fact that the element will hide what's underneath it, and that overlay will never close unless a user clicks again on the toggle.
Frankly, to me these are blocking issues, would there be any chance to go back to the previous behavior and not introduce these issues without any real advantage..?

PR #4600, opened 2026-07-03
feat(components): Adding story card component - FRONT-5379

planctus:


src/components/story-card/story-card.js
```diff
@@ -0,0 +1,683 @@
...
+      currentSelector = '[data-ecl-story-card-current]',
+      totalSelector = '[data-ecl-story-card-total]',
+      gridItemSelector = '[data-ecl-story-card-grid-item]',
+      gridButtonSelector = '[data-ecl-story-card-grid-button]',
+      gridDetailsSelector = '[data-ecl-story-card-grid-details]',
+      gridPrevSelector = '[data-ecl-story-card-grid-prev]',
+      gridNextSelector = '[data-ecl-story-card-grid-next]',
+      gridPlaySelector = '[data-ecl-story-card-grid-play]',
+      gridPauseSelector = '[data-ecl-story-card-grid-pause]',
+      gridAutoplayDelay = 10000,
```
emeryro, line 90: I think that the blue bar at the bottom is supposed to represent the timer before next slide (10s)

Image

src/components/story-card/demo/data--testimonial.js
```diff
@@ -0,0 +1,119 @@
...
+        sources: [
+          {
+            src: 'https://i.pravatar.cc/800?img=18',
+            media: '(max-width: 50rem)',
+          },
+        ],
+      },
+      teaser_label: 'RICHARD FELIX',
+      author: 'Richard Felix',
+      role: 'Lawier',
```
emeryro:
```suggestion
      role: 'Lawyer',
```

src/components/story-card/story-card-card.html.twig
```diff
@@ -0,0 +1,137 @@
...
+    <svg xmlns="http://www.w3.org/2000/svg" width="22" height="45" viewBox="0 0 22 45" fill="none">
+      <path d="M0 0H22V23.3143L0 44.5714V0Z" fill="currentColor"/>
+    </svg>
+    <svg xmlns="http://www.w3.org/2000/svg" width="22" height="45" viewBox="0 0 22 45" fill="none">
+      <path d="M0 0H22V23.3143L0 44.5714V0Z" fill="currentColor"/>
+    </svg>
+  </div>
+{% endif %}
+
+{% if _title %}
```
emeryro: is that enough to just test the variable? We usually add the "is not empty"

src/components/story-card/story-card.scss
```diff
@@ -0,0 +1,445 @@
...
+
+// Exposed variables
+$theme: null !default;
+$story-card: null !default;
+$col-9-width: 970px;
+$min-height-desktop: 548px;
+
+.ecl-story-card {
+  --story-card-grid-breakpoint: $col-9-width;
+  --story-card-min-height: $min-height-desktop;
```
emeryro: missing #{ ...} for the two variables

src/components/story-card/story-card.scss
```diff
@@ -0,0 +1,445 @@
...
+      .ecl-story-card__grid-container,
+      .ecl-story-card__grid-details {
+        box-sizing: border-box;
+        flex: 1;
+        margin-block-start: 0;
+      }
+
+      .ecl-story-card__grid-card {
+        height: 100%;
+        min-height: $min-height-desktop;
```
emeryro: shouldn't we use var(--story-card-min-height) instead?

src/components/story-card/story-card.js
```diff
@@ -0,0 +1,683 @@
...
+  resetGridDetailsHeight() {
+    this.gridDetails.forEach((details) => {
+      details.style.height = '';
+      details.style.visibility = '';
+      details.style.pointerEvents = '';
+    });
+  }
+
+  /**
+   * Set every grid details panel to the height of the tallest one.
```
emeryro, line 436: not sure if I got it wrong, but when I set a very long description in the first item, the other items keep being smaller

src/components/story-card/story-card.js
```diff
@@ -0,0 +1,683 @@
...
+      this.btnGridPause.addEventListener('click', () => {
+        this.toggleGridAutoplay(false);
+      });
+    }
+  }
+
+  /**
+   * Handle window resize.
+   */
+  onWindowResize() {
```
emeryro, line 660: we may use a debounce here, as we do in other ocmponents

src/components/story-card/story-card.js
```diff
@@ -0,0 +1,683 @@
...
+    this.scheduleGridDetailsHeight();
+  }
+
+  /**
+   * Destroy the component.
+   */
+  destroy() {
+    if (this.slider) {
+      this.slider.destroy();
```
emeryro, line 669: I assume that the different event added before are removed by this destroy() 
(not fully sure about this)

src/components/story-card/story-card.scss
```diff
@@ -0,0 +1,445 @@
...
+
+  .ecl-story-card__carousel {
+    margin-block-start: var(--s-xl);
+    position: relative;
+  }
+
+  .ecl-story-card__carousel {
+    display: block;
+
+    @container (width >= 600px) {
```
emeryro: wouldn't it be safer to define a name for the container queries, at component root level?

src/components/story-card/story-card.scss
```diff
@@ -0,0 +1,445 @@
...
+  .ecl-story-card__carousel-slide {
+    flex: 0 0 100%;
+    min-width: 0;
+  }
+
+  .ecl-story-card__card {
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+    background: var(--c-w);
```
emeryro: not sure if this is as expected, but just to be sure: the --c-w color is the semi transparent one, not a full white

src/components/story-card/story-card.story.js
```diff
@@ -0,0 +1,221 @@
...
+        type: { summary: 'string' },
+        defaultValue: { summary: '' },
+      },
+    },
+  };
+
+  return argTypes;
+};
+
+const prepareData = (data, args) => {
```
emeryro, line 79: some of the controls are not applied. I noted 'description', 'teaser label'  and 'author name', but there may be more

emeryro (changes requested):
I guess that there will be some accessibility improvements required. What I noted (to be confirmed by Regine):
- card list could be an HTML list
- when selecting a "card", focus could be put on the content area, to avoid having to go through the other items

src/components/story-card/story-card.js
planctus: although it's not the same thing it would be kind of redundant, the thing is that the function we execute on resize has cancelAnimationFrame and requestAnimationFrame which can only happen once per frame, so it felt smarter than adding a timeout and it's not anything that heavy what we do on resize, anyhow.

planctus:
Well, about accessibility, lots of things have been already done for that, in reality, but there is always a certain level of questionability in the choices you make about it, so we will see.
I use a list for the "tabs", so left column in desktop, now. 
We cannot move the focus to the "content" of the card directly, otherwise the user would not be able to reach the other cards and here is not like in the quiz, it's only 4 cards maximum and with not much to focus on, so i hope we will not have to handle a custom keyboard navigation at all, at least for the moment there is no need.

src/components/story-card/story-card.js
planctus: i think you got it right, but if you add text manipulating the dom, like with the inspector then you will have to do something like resizing the window to update the calculations and you should get all the elements with the same height. This was not in the requirements but i guess it was "implied"..

emeryro (approved):

PR #4602, opened 2026-07-07
feat(highlighted-search): add component - FRONT-5378

emeryro:


planctus:
I don't really understand why in mobile those tags are not taking full width, there might be a limit to their number in the backend, but it still looks pretty weird even with a few ones.

src/components/banner-search/banner-search.html.twig
```diff
@@ -0,0 +1,146 @@
...
+              extra_classes: 'ecl-banner-search__form-input' ~ ( _search_input.extra_classes ? ' ' ~ _search_input.extra_classes : '' ),
+            }),
+          } only %}
+
+          {% include '@ecl/button/button.html.twig' with _search_button|merge({
+            size: _search_button.size|default('m'),
+            variant: _search_button.variant|default('primary'),
+            style: _search_button.style|default('highlight'),
+            type: _search_button.type|default('submit'),
+            extra_classes: 'ecl-banner-search__form-button'
```
planctus: mmh..this way we would override any custom class coming from the user, not that crucial but we could always merge the two, but instead of doing this, why don't we create an object with our default values and then we simply merge the object from the data as it is, we need to process the extra_classes if we want to preserve the ones provided by the user but it would be cleaner, imho..

src/components/banner-search/banner-search.html.twig
```diff
@@ -0,0 +1,146 @@
...
+          } only %}
+
+          {% include '@ecl/form-group/form-group.html.twig' with {
+            label: _search_label|default(_search_input.placeholder|default('')),
+            hide_label: true,
+            extra_classes: 'ecl-banner-search__form-group',
+            input: _search_input|merge({
+              width: 'l',
+              input_type: 'search',
+              extra_classes: 'ecl-banner-search__form-input' ~ ( _search_input.extra_classes ? ' ' ~ _search_input.extra_classes : '' ),
```
planctus: i think we could avoid the ternary operator here: 
`extra_classes: ('ecl-banner-search__form-input ' ~ _search_input.extra_classes|default(''))|trim,`

src/components/banner-search/banner-search.html.twig
```diff
@@ -0,0 +1,149 @@
...
+      <div class="ecl-banner-search__suggestion">
+        {% if _suggestion_label is not empty %}
+          <span class="ecl-banner-search__suggestion-label" id="{{ _suggestion_label_id }}">
+            {{- _suggestion_label -}}
+          </span>
+        {% endif %}
+
+        {% include '@ecl/tag/tag-set.html.twig' with _suggestion|merge({
+          extra_classes: 'ecl-banner-search__suggestion-list',
+          extra_attributes: _suggestion_label is not empty ? [{ name: 'aria-labelledby', value: _suggestion_label_id }] : [],
```
planctus: mmh..here is delicate, i think that they might want to add an attribute for analytics, we should merge the ones potentially coming from the user data, less important for the classes..

src/components/banner-search/banner-search.scss
```diff
@@ -0,0 +1,142 @@
...
+  border-radius: map.get($theme, 'border-radius', 's');
+  color: map.get($banner-search, 'color');
+  padding: map.get($banner-search, 'padding');
+}
+
+.ecl-banner-search__content {
+  align-items: center;
+  display: flex;
+  flex-direction: column;
+  max-width: 600px;
```
planctus: We use a banner, then we have to limit the width of its content to 600px.. :)

src/components/highlighted-search/highlighted-search.html.twig
```diff
@@ -0,0 +1,149 @@
...
+            icon: {
+              name: 'magnifying-glass',
+              family: 'phosphor',
+              size: 'm',
+            },
+            extra_classes: 'ecl-highlighted-search__form-icon'
+          } only %}
+
+          {% include '@ecl/form-group/form-group.html.twig' with {
+            label: _search_input.placeholder|default(_search_button.label)|default(''),
```
planctus, line 100: Reading the "specs" it seems they will have a default value for the placeholder so this should never be a problem, but why using alternatively the label of the search button, that could be unexpected in case they could also send an empty value for the placeholder.

src/components/highlighted-search/highlighted-search.scss
```diff
@@ -0,0 +1,150 @@
...
+  align-self: flex-start;
+  margin-block-start: map.get($highlighted-search, 'helper', 'spacing');
+
+  @include mixins.responsive-font(
+    map.get($highlighted-search, 'helper', 'font')
+  );
+}
+
+// Suggestion
+.ecl-highlighted-search__suggestion {
```
planctus, line 131: This could open the door to a wider issue, we use border-primary for the focus color, but when used on a variable background using the color modes this might not always work, in the default color mode. at least, the focus on the tags is barely visible, i guess we need to do something about it.

src/components/highlighted-search/highlighted-search.story.js
```diff
@@ -0,0 +1,97 @@
...
+export const Default = (_, { loaded: { component } }) => component;
+
+Default.render = async (args) => {
+  const rendered = await HighlightedSearch(prepareData(defaultData, args));
+  return rendered;
+};
+Default.storyName = 'default';
+Default.args = getArgs(defaultData);
+Default.argTypes = getArgTypes();
+Default.parameters = { notes: { markdown: notes, json: defaultData } };
```
planctus: we changed the way we pass the data to the plugin, doing this way would not update the include example with the current data being used.

src/components/highlighted-search/highlighted-search.test.js
```diff
@@ -0,0 +1,53 @@
+import {
+  merge,
+  renderTwigFileAsNode,
+  renderTwigFileAsHtml,
+} from '@ecl/test-utils';
+import { axe, toHaveNoViolations } from 'jest-axe';
+
+import data from './demo/data';
+
+data.id = 'highlighted-search';
```
planctus: Why not adding this in the data for the demo, it's good that we get a default value with a random string from the template but it's not nice to "read" it in the markup

src/components/highlighted-search/highlighted-search.html.twig
emeryro: That was a request from Regine, in a figma comment. The form input should always have a filled label (hidden here), even if the placeholder is empty

planctus: ah, but i just misread what we were doing, i read placeholder and i thought we were dealing with it, forget about it

src/components/highlighted-search/highlighted-search.story.js
emeryro: fixed, also in a few other stories not up to date

src/components/highlighted-search/highlighted-search.scss
emeryro: new focus mixins come in handy there

PR #4603, opened 2026-07-08
style(focus): add mixin and css classes for focus - FRONT-5391

emeryro:
Add SASS mixin and corresponding CSS classes to handle focus display.
Two options available:
- default focus: to be used in component to avoid declaring it every times
- inverted focus: to be used contextually when the component will be placed on a dark background


src/mixins/color/mixins.scss
```diff
@@ -56,3 +56,21 @@
     --ecl-focus-outline-color: var(--ecl-link-color);
   }
 }
+
+// Mixin to set default focus on an element
+@mixin focus-default() {
+  &:focus-visible {
+    box-shadow: 0 0 0 2px var(--cm-border-inverted, #fff);
+    outline: 2px solid var(--cm-border-primary, var(--c-p));
+    outline-offset: 2px;
```
planctus: i think we have a few cases where we "alter" the offset, couldn't we add a parameter with a default for it so we can do it via the mixin..?

planctus (approved):

PR #4613, opened 2026-07-14
feat(accordion): add background - FRONT-5401

emeryro:
Add white background to accordion items

### DIFF
--- a/src/components/accordion/accordion.scss
+++ b/src/components/accordion/accordion.scss
@@ -74,7 +74,7 @@ $accordion: null !default;
 }
 
 .ecl-accordion__item {
-  background-color: var(--surface-inverted, #fff);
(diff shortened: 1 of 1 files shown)


src/components/accordion/accordion.scss
```diff
@@ -74,6 +74,7 @@ $accordion: null !default;
 }
 
 .ecl-accordion__item {
+  background-color: var(--surface-inverted, #fff);
```
planctus:
```suggestion
  background-color: var(--cm-surface-inverted, #fff);
```

planctus:
weren't we supposed to add also some different color on hover for the active item..?

emeryro:
There is no design provided for the hover color. We can do it later if needed

PR #4615, opened 2026-07-15
feat(tag): update display - FRONT-5390

emeryro:
Update tag display (color and underline)


src/components/tag/tag.scss
```diff
@@ -31,11 +30,9 @@ $tag: null !default;
   }
 
   &:hover {
-    background-color: var(
-      --cm-surface-medium,
-      #{map.get($tag, 'background-color-hover')}
-    );
+    background-color: map.get($tag, 'background-color-hover');
     color: map.get($tag, 'color-hover');
+    cursor: pointer;
```
planctus, line 35: mmh..but why the removable tag should have cursor pointer on the whole element..? Should be applied only to the icon, no..? Or maybe not, the reality is that it depends on where the click listener will be attached, so forget about this..

planctus (approved):

PR #4616, opened 2026-07-15
fix(collapsible-sidebar): Using fixed spacing for the accordion toggle - FRONT-5400

planctus:
…

### DIFF
--- a/src/components/accordion/accordion.scss
+++ b/src/components/accordion/accordion.scss
@@ -175,8 +175,8 @@ $accordion: null !default;
     align-items: center;
     border-block-end: map.get($accordion, 'sidebar', 'border');
     color: map.get($accordion, 'sidebar', 'title-text-color');
-    margin-block-start: calc(1.33em - var(--s-s));
-    padding: var(--s-s) 0;
+    margin-block-start: var(--s-3xl);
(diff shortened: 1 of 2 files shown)


src/components/accordion/accordion.scss
```diff
@@ -175,8 +175,8 @@ $accordion: null !default;
     align-items: center;
     border-block-end: map.get($accordion, 'sidebar', 'border');
     color: map.get($accordion, 'sidebar', 'title-text-color');
-    margin-block-start: calc(1.33em - var(--s-s));
-    padding: var(--s-s) 0;
+    margin-block-start: var(--s-3xl);
+    padding-block-end: var(--s-s);
```
emeryro: with this, there is some unwanted padding on the toggle

Image

emeryro (changes requested):

src/components/accordion/accordion.scss
planctus: ah, i changed it in the last moment and i didn't check it afterwards, need to reset the default padding top, basically, so that the spacing will be 3xl, which is what they are using for the content area

emeryro (approved):

PR #4632, opened 2026-07-27
fix(headings): fix headings font weight on EU components - FRONT-5424

emeryro:
Change font weight of the headings in following component, to be consistent with other headings
- file
- quiz
- story card


planctus:
there was a late discovery, forgot to add it in the ticket, also the slogan ticker should use regular in EU.

planctus (approved):

PR #4636, opened 2026-07-27
chore(storybook): hide/remove compositions - FRONT-5429

emeryro:


planctus (commented):
But why do we remove etrans and only "hide" file upload status..? I mean, why the different approach..? Also, if we remove the packages we should also remove them as dependency in twig-templates.

emeryro:
For now, etrans composition is removed completely, and file upload composition is hidden, as it may still be useful in the future.
Twig-templates updated

PR #4640, opened 2026-07-28
feat(rating-field): new markup structure - FRONT-5430

emeryro:
Use a different markup structure (and corresponding css) for the rating field.
It still uses only css, but now the input are placed in the same order visually and for keyboard.
Data structure is unchanged, but markup has been updated.


planctus (approved):
Brilliant, now it's straightforward, finally..

PR #4641, opened 2026-07-28
feat(slider): Adding slider component to handle the slider pager - FRONT-5427

planctus:


emeryro (commented):
Great, that's cleaner.

There is just some naming inconsistency. The twig file is called slider-pager,  the js file is called slider (but the class is SliderPager), and the root css class is ecl-slider__pager

emeryro (changes requested):
The files are fine, but the package is still names 'slider' (also in the README)

planctus:
That is intentional, frankly, the idea is to create a package "slider" and define for the moment only the pager. That was also the idea when using file names as slider.js and slider.scss, now i changed those but to me it makes sense to keep the package as slider, imagine if tomorrow we add anything else about it, why having separated packages..

PR #4652, opened 2026-07-31
fix(accordion): Resetting background color on sidebar variant - FRONT-5444

planctus:
…-5444


emeryro (commented):
Why are we putting the background override only for the first item (is-first)?
I guess that it is possible to have multiple items in the sidebar

emeryro (approved):

PR #4665, opened 2026-08-10
fix(story-card): fix border radius on mobile - FRONT-5449

emeryro:
Remove unneeded border radius on mobile


planctus:
Why resetting only the values for the bottom corners..? Basically for .ecl-story-card__content we don't need any border radius, no?

emeryro:
true. The request mentioned only the lower corners and I didn't see the others.
Updated

planctus (approved):

PR #4672, opened 2026-08-17
feat(forms): Requred and Optional text visible when the label is hidden -  FRONT-5458 

planctus:
Keep required and optional label visible, even if the form label is hidden


emeryro (approved):
This is fine for me. I am just a bit worried that the required/optional label will suddenly appear on existing form (if filled), but that is the request

PR #4673, opened 2026-08-18
feat(container-queries): Applying to different components - FRONT-5446

planctus:
Applied to:

- form elements
- description list
- fact and figures
- navigation list
- news ticker
- tabs

This revealed an inconsistency both on chromatic side and in relation to the storybook plugin that shows the breakpoint, therefore snapshot viewports have been increased ( + 48px, 32px of inline padding and 16px of scrollbar) and the plugin is now reading a configuration that identifies the container and if present will use its width to determine the breakpoint to be shows while still giving the width of the viewport.


src/components/description-list/description-list.scss
```diff
@@ -13,6 +12,11 @@ $description-list: null !default;
 
 .ecl-description-list,
 %ecl-description-list {
+  container-type: inline-size;
+  container-name: list-container;
```
emeryro: we are missing the ecl namespace here

src/components/description-list/description-list.scss
```diff
@@ -198,8 +202,8 @@ $description-list: null !default;
 
 // Horizontal description lists
 
-@include breakpoints.up('l') {
-  .ecl-description-list--horizontal,
+@container list-container (width >= #{map.get($theme, 'breakpoint', 'l')}) {
+  .ecl-description-list--horizontal .ecl-description-list__content,
```
emeryro, line 206: I see that you added the container definition in the default css.
won't it cause issue, if we have to define other container later (for blockquote for instance)?

src/components/form-group/form-group.scss
```diff
@@ -16,6 +16,8 @@ $form-group: null !default;
   display: block;
   margin: 0;
   padding: 0;
+  container-type: inline-size;
+  container-name: ecl-form-element;
```
emeryro, line 20: that seem to cause issue. The search form in the page example is broken

emeryro (changes requested):

src/components/description-list/description-list.scss
planctus: the container can have multiple names, that is the idea, we will add the other ones in case there will be any.

src/components/form-group/form-group.scss
planctus: oh my god.. :) frankly i have been considering container-type as a way to enable container queries, basically, did not imagine that it could affect the layout, this is as critical as the misalignment with the picture's sources queries in relation to the adoption of container queries, wow..

I don't know, i fixed the search form but this is something out of our control, i mean that we cannot control where a component will be used, what styles will have their containers and there is no bullet proof way of defining a container without any risk.

We have been using container queries in differerent components and apparently never had bad consequences, so we could consider this an "edge" case, maybe, but this discovery has compromised my trust in the container queries again.. :)

src/components/fact-figures/fact-figures.scss
```diff
@@ -16,6 +15,11 @@ $fact-figures: null !default;
   background-color: map.get($fact-figures, 'background');
   border-radius: map.get($fact-figures, 'border-radius');
   margin: 0;
+  container-type: inline-size;
+  container-name: ecl-fact-figures;
+}
+
+.ecl-fact-figures__items {
```
emeryro: this is causing padding issue for elements outside of the items container

Image

emeryro (changes requested):
For the components updated I found this additional issue.
Once fixed, it may be worth running chromatic here

src/components/fact-figures/fact-figures.scss
planctus: auch, yeah, i wanted to keep the markup untouched,i added now a wrapper where to apply the global padding

emeryro (approved):

PR #4674, opened 2026-08-18
feat(checkbox, radio): update disabled colors - FRONT-5457

emeryro:
Change color for disabled state, to improve accessibility


planctus (commented):
The disabled state set in the standalone story is kept in the checkbox group one, guess that is not intended since we have an item that we show disabled by default.

PR #4680, opened 2026-08-24
fix(slider-pager): larger click area for dots - FRONT-5460

emeryro:
Use larger click areas for the slider dots


planctus (commented):
I surely prefer this solution to completely remove the interaction on this element, but style wise it seems to me very impacting, it's not maybe our business but i wonder if it wouldn't be better instead enlarging a bit the dots, something like this: 

Image

PR #4681, opened 2026-08-24
feat(quiz): Color modes, images and illustration  - FRONT-5463 

planctus:


src/components/quiz/quiz-card.html.twig
```diff
@@ -3,9 +3,8 @@
 {#
   Parameters
     - "id" (string) (default: random): Unique id for the quiz item; generated from quiz id or randomized
-    - "category" (string) (default: '') (poll variant) Category of the quiz item, eg: 'Geography'
-    - "success_category" (string) (default: '') Category to show on the back of the card in case of correct answer, eg: 'Correct'
-    - "error_category" (string) (default: '') Category to show on the back of the card in case of incorrect answer, eg: 'Incorrect'
+    - "image" (string) (default: 'none'): With image, on both side or only in the back (None, Reveal, Always)
```
emeryro: parameter names may be a little confusing (picture and image). Maybe use something like `image_display`.

src/components/quiz/quiz-card.html.twig
```diff
@@ -89,146 +99,167 @@
...
         {% endif %}
-        {% if _error_category is not empty %}
-          <span
-            class="ecl-quiz-card__category ecl-quiz-card__category--error"
-            tabindex="0"
-          >
-            {{- _error_category -}}
-          </span>
-        {% endif %}
-        {% if _variant == 'poll' %}
```
emeryro, line 182: we no longer display the status (Correct/Incorrect) for screen readers (previously in ecl-quiz-card__sr-status)
If that's fine for accessibility, ok, but I feel like we are removing important information here.

src/components/quiz/quiz.scss
```diff
@@ -117,6 +117,102 @@ $quiz: null !default;
...
+.ecl-quiz-card__illustration {
+  border-start-start-radius: map.get($theme, 'border-radius', 'm');
+  border-start-end-radius: map.get($theme, 'border-radius', 'm');
+  aspect-ratio: 3.5/1;
+  object-fit: cover;
+  width: 100%;
+}
+
+.ecl-quiz-card__illustration {
+  svg {
```
emeryro: :disappointed_relieved:

src/components/quiz/README.md
```diff
@@ -20,10 +20,9 @@ npm install --save @ecl/quiz
 - **"items"** (array) (default: []) Quiz cards; format:
   - "id" (string) (default: random) Unique id for the card
   - "variant" (string) (default: 'reveal') Card variant; can be 'reveal', 'poll'
-  - "category" (string) (default: '') Category label shown on front of card (poll variant)
```
emeryro, line 23: I'm still a bit worried by the removal of the category... This could be a breaking change for some users.
Hopefully no one is using the quiz yet

emeryro (commented):
Just some small remarks

src/components/quiz/quiz-card.html.twig
planctus: we don't display that "status" at all, not just in the screen readers, i guess that this new design was reviewed also in terms of accessibility, the idea is that if it's not relevant for sighted users and can be removed then the same applies to screen reader users.
I guess that it was removed for other reasons, though... basically when we have an image, there is no room for it anymore, if we will receive requests about it we will handle them.

emeryro (approved):

PR #4690, opened 2026-08-27
feat(megamenu): scroll on mobile - FRONT-5404

emeryro:
Make the whole menu scrollable on mobile, to avoid having the links barely usable if there is a long description

Tablet and desktop display are unchanged


planctus:
It doesn't seem to work when the info panel itself is taller than the available height: 
image
As soon as a portion of the next container (ecl-mega-menu__mega) becomes visible the scrollbar appears.
What i see in the DOM is that if this element is partially visible it gets an height set by the js otherwise it doesn't , but those operations in the js were based on the idea of making the list scrollable, so probably there is still the need to set an height via js, but on a different element.

emeryro:
I removed the height calculus on mobile, as we don't need it anymore.
Should be cleaner that way

src/components/mega-menu/mega-menu.scss
```diff
@@ -1469,6 +1474,10 @@ $mega-menu: null !default;
     }
   }
 
+  .ecl-mega-menu__sublist--scrollable {
```
planctus: it seems to me that this class doesn't exist anymore, it was set by the js to handle the scrolling in mobile which is not needed anymore

src/components/mega-menu/mega-menu.js
```diff
@@ -835,72 +834,7 @@ export class MegaMenu {
           if (this.inner) {
             this.inner.style.top = `${bottomPosition}px`;
           }
```
planctus, line 819: Also the js can be further cleaned, no need to remove the ecl-mega-menu__sublist--no-border and ecl-mega-menu__sublist--scrollable classes, there is also a condition based on the latter in heandleClickGlobal(e)

emeryro:
good point, I did some additional cleanup in the js/css, to remove dead code

src/components/mega-menu/mega-menu.js
emeryro: the condition in handleClickGlobal(e) is on another class (still used). I removed the rest

planctus:
i noticed now that we introduced an issue in the 3 level in mobile, there is an horizontal scrollbar, it is due to the struggle to pretend that there is a grey header and instead it's a collection of elements and pseudo elements, this link here: .ecl-mega-menu__parent-link and its :after companion have width: 100vw, the link has also negative margins to fill the available width. It is probably more surprising that it was working before than it is discovering this issue now.

planctus (approved):

PR #4691, opened 2026-08-27
feat(mega-menu): highlight current item - FRONT-5455

emeryro:
Highlight the current first level item (visually and for accessibility tools)
New label required: `current_aria_label` (some other label were already present but just missing from the documentation)


planctus (commented):
It's only a label, but i find confusing "current section" used as the title of the current menu item, it seems to refer to the menu section while instead it means that the user is in the page linked by the menu item and indeed it's aria-current value is "page" and not "true".  (in a SPA this doesn't make sense, but they will still be able to use the styles and then use the aria-current the way they prefer)

emeryro:
That was the label proposed by Regine
But this title is added only for menu items with children. First level links just get the aria-current="page"

planctus (approved):

PR #4692, opened 2026-08-28
fix(sources): Fixing issues with sources display on multiple lines - FRONT-5469

planctus:


emeryro (changes requested):
Good for me on screen, but to be fixed on print (maybe that was already broken before...)

Image

emeryro (approved):

PR #4693, opened 2026-09-02
feat(table): filter tables - FRONT-5472

emeryro:
Add filtering option to the tables


planctus (commented):
I think we can optimize a bit the js, we have a few repetead DOM queries that we can make only once:

- queryAll('tr', tbody)
- queryOne('tbody', this.element)

The functionality works, although i believe this is increasing the complexity of this component that we are treating more and more as a widget with functionalities more than an UI component.

PR #4694, opened 2026-09-02
feat(gallery): additional thumbnail details for video - FRONT-5471

emeryro:
Display additional information on video thumbnails and overlay
Information are always displayed for video (and can be always displayed on images too if needed)


planctus (approved):
I find just a but weird that these additional info are only available in the thumbnail, you go in the big overlay and they are not there

PR #4697, opened 2026-09-04
feat(carousel):  Teaser navigation, styles update, js refactoring using embla

planctus:


emeryro (approved):
Fine for me technically.
I'm quite sure that we will get remarks regarding the plain banner though (not directly related to the work here)
Some colors are very aggressive, and make the call to action barely visible...

Image
Also on mobile, without image the text is quite lost in the slide

Image

PR #4699, opened 2026-09-07
fix(sources): Fixing the text wrapping in sources when an icon is used - FRONT-5469

planctus:
…


emeryro (changes requested):
Good for screen, just the print to be fixed

emeryro (approved):

PR #4700, opened 2026-09-10
feat(featured-item): update EU link - FRONT-5485

emeryro:
Allow usage of highlighted link on EU, and provide styles for it


planctus (commented):
The scss are simply unreadable, they give me headaches, not your fault, but i really hope that in the future, whatever will be the namespace used for the components, this will be avoided.
I just have one question, not really related to this  work so it can be ignored, but playing with the controls i noticed that the link is displayed in the left when no media is present, i wonder if that is on purpose, i was expecting it on the right as it is when the media is present. 

Image

planctus:
Also not introduced in this pr, we have this issue in EU only whe nthe text is centered: 
image

planctus:
Another small issue with the text centered, not introduced in this pr: 
image

emeryro:
I fixed the center alignment
For the link position without media, this is how it is currently. I suspect accessibility request here, to keep the link closer to the text (when it is left-aligned...)
There is nothing about that in Figma

planctus (approved):

PR #4703, opened 2026-09-14
feat(table): Styles update in EC and EU - FRONT-5483

planctus:


emeryro (commented):
A few remarks concerning the design 

- when there is no header, the first border is missing

Image

- on EU mobile, the color seems inverted for multi header (compared to desktop)

Image

planctus:
eh, i'm not really sure this design in EU makes sense, probably not, but it's like this in figma, ungrouped columns have a lighter background in desktop, in mobile the sub columns have that color and all the others have the darker background..

PR #4712, opened 2026-09-21
feat(layout-wrapper): improve markup and css - FRONT-5494

emeryro:
No longer rely on utilities, and improve display when ECL "default" css is not used


planctus:
well, happy to get rid of the utility but we have to include here all the css for responsive headings on this specific selector...do we really need to support h5 and h6..?

